PR #40, opened 2024-02-23
Changing image_script to use new image definition V1

peytonr18:
Updating image_creation script to use new testgalleryimage (a gen1 image) to avoid issues with TrustedLaunch and converting the image to SIG directly rather than using a managed image.

### DIFF
--- a/demo/image_creation.sh
+++ b/demo/image_creation.sh
@@ -106,7 +106,6 @@ echo "Capturing OsDisk snapshot of vm $vm with image $image"
 echo "*********************************************************************"
 target_disk=$(az disk show --ids $(az vm show -g $rg -n $vm | jq .storageProfile.osDisk.managedDisk.id -r) | jq .name -r)
 az snapshot create -g $rg -n $vm-snapshot --source $target_disk
-az snapshot show -g $rg -n $vm-snapshot
 
 version=$(date '+%Y.%m%d.%H%M%S')
 gallery=testgalleryagent
--- a/demo/image_creation.sh
+++ b/demo/image_creation.sh
@@ -55,7 +55,7 @@ storage=testagent$epoch
 location=eastus
 vm=testvm-$epoch
 ssh_key_path=~/.ssh/id_rsa.pub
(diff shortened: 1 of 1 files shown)


demo/image_creation.sh
```diff
@@ -102,23 +102,22 @@ fi
...
+az snapshot create -g $rg -n $vm-snapshot --source $target_disk
+az snapshot show -g $rg -n $vm-snapshot
 
 version=$(date '+%Y.%m%d.%H%M%S')
 gallery=testgalleryagent
-definition=testgallery
+definition=testgallery-gen1
 gallery_rg=temp-rg-rust-agent-testing
-image_id=$(az image show -g $rg -n $image | jq .id -r)
+subscription_id=$(az vm show -g $rg -n $vm | jq -r '.id')
```
anhvoms, line 115: this is not needed (subscription_id) not used anywhere

demo/image_creation.sh
```diff
@@ -102,23 +102,21 @@ fi
...
-echo "Deallocating and generalizing vm to capture image $image"
+echo "Capturing OsDisk snapshot of vm $vm with image $image"
 echo "*********************************************************************"
-az vm deallocate -g $rg -n $vm
-az vm generalize -g $rg -n $vm
-az image create -g $rg -n $image --source $vm --hyper-v-generation V2
-echo "Done"
+target_disk=$(az disk show --ids $(az vm show -g $rg -n $vm | jq .storageProfile.osDisk.managedDisk.id -r) | jq .name -r)
+az snapshot create -g $rg -n $vm-snapshot --source $target_disk
+az snapshot show -g $rg -n $vm-snapshot
```
anhvoms, line 109: when the az snapshot create command succeeded, does it print out some information already? If it does, I don't think we need the az snapshot show

peytonr18: Az snapshot create does -- think I was originally using it for debugging the snapshot_id but can easily get rid of the az snapshot show as it was just for debugging purposes as you noted.

anhvoms (approved):
Thanks!

PR #44, opened 2024-02-28
README: Don't recommend curl | sh to install Rust

rata:
Using `curl | sh ` is questionable from a security POV and not recommended in several circles already.

Let's not do it ourselves, let's link to the installation page, they can (and do) recommend `curl | sh` anyways.

What do you think?

Below is the commit msg:

---

While we are there, let's remove how to clone the repo. That should be clear on Github already.

Signed-off-by: Rodrigo Campos <rodrigoca@microsoft.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,15 +12,7 @@ Azure-init has very few requirements on its environment, so it may run in a very
 
 ## Installing Rust
 
-Rust can be installed via the command line using the following command: 
-`curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh`, or with other options found here:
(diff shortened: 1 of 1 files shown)


dongsupark:
I am okay with the change, but for now will wait and see if other folks have other opinions.

jeremycline:
I like pointing to the official instructions for dependencies when there's no special requirements to configure/set up so I am also +1, for what it's worth.

t-lo:
I would also prefer pointing to the official Rust instructions. Secondly, this change removes a potential security risk that's so bad it has meme status 😉

t-lo (approved):
Love it, thank you!

dongsupark (approved):

Issue #42, opened 2024-02-28
Update crates h2, regex, thread_local, term


dongsupark:
`cargo audit` gives 3 security vulnerabilities and 1 warning.
As this repo does not have an automated dependabot alert, a manual update of the packages is necessary.

```
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 605 security advisories (from /home/dpark/.cargo/advisory-db)
    Updating crates.io index
    Scanning Cargo.lock for vulnerabilities (158 crate dependencies)
Crate:     h2
Version:   0.3.21
Title:     Resource exhaustion vulnerability in h2 may lead to Denial of Service (DoS)
Date:      2024-01-17
ID:        RUSTSEC-2024-0003
URL:       https://rustsec.org/advisories/RUSTSEC-2024-0003
Solution:  Upgrade to ^0.3.24 OR >=0.4.2
Dependency tree:
h2 0.3.21
├── reqwest 0.11.22
│   └── libazureinit 0.1.1
│       └── azure-init 0.1.1
└── hyper 0.14.27
    └── reqwest 0.11.22

Crate:     regex
Version:   0.2.11
Title:     Regexes with large repetitions on empty sub-expressions take a very long time to parse
Date:      2022-03-08
ID:        RUSTSEC-2022-0013
URL:       https://rustsec.org/advisories/RUSTSEC-2022-0013
Severity:  7.5 (high)
Solution:  Upgrade to >=1.5.5
Dependency tree:
regex 0.2.11
├── rustfmt 0.10.0
│   └── libazureinit 0.1.1
│       └── azure-init 0.1.1
└── env_logger 0.4.3
    └── rustfmt 0.10.0

Crate:     thread_local
Version:   0.3.6
Title:     Data race in `Iter` and `IterMut`
Date:      2022-01-23
ID:        RUSTSEC-2022-0006
URL:       https://rustsec.org/advisories/RUSTSEC-2022-0006
Solution:  Upgrade to >=1.1.4
Dependency tree:
thread_local 0.3.6
└── regex 0.2.11
    ├── rustfmt 0.10.0
    │   └── libazureinit 0.1.1
    │       └── azure-init 0.1.1
    └── env_logger 0.4.3
        └── rustfmt 0.10.0

Crate:     term
Version:   0.4.6
Warning:   unmaintained
Title:     term is looking for a new maintainer
Date:      2018-11-19
ID:        RUSTSEC-2018-0015
[...]

PR #45, opened 2024-02-29
Bump reqwest to the latest version and drop rustfmt

jeremycline:
## How to use

Ensure formatting still works as this drops the rustfmt dependency.

## Testing done

Built the project, ran all the tests, ensured `make fmt` still worked as expected by editing a Rust file:
```
❯ make fmt

**********************************
* Formatting
**********************************

Diff in /home/jcline/devel/github.com/Azure/azure-init/libazureinit/src/distro.rs at line 27:
     ) -> Result<i32, String> {
         match self {
             Distributions::Debian | Distributions::Ubuntu => {
-                let mut home_path = "/home/".to_string(); home_path.push_str(username);
+                let mut home_path = "/home/".to_string();
+                home_path.push_str(username);
 
                 match Command::new("useradd")
                     .arg(username)
make: *** [Makefile:26: fmt] Error 1
```

Together these commits address #42 so (at the time of this PR) cargo-audit should be happy.

### DIFF
--- a/libazureinit/Cargo.toml
+++ b/libazureinit/Cargo.toml
@@ -8,7 +8,7 @@ license = "MIT"
 description = "A common library for provisioning Linux VMs on Azure."
 
 [dependencies]
-reqwest = { version = "0.11.18", default-features = false, features = ["blocking", "json"] }
+reqwest = { version = "0.11.24", default-features = false, features = ["blocking", "json"] }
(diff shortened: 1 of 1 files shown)


dongsupark (approved):
Looks good.
Now `cargo audit` returns ok. :+1:

PR #47, opened 2024-02-29
github: include the instructions as comments in the templates

jeremycline:
Use Markdown comments for the instructions in the PR and issues templates. This way you don't need to remove the instructions as you fill out the form, but they aren't rendered in the resulting issue/PR.

## How to use

Create issues and PRs without needing to delete the instructions as you go.

## Testing done

Examined the output in the VSCode markdown preview window to see the comments not get rendered. This is also the [documented syntax](https://github.github.com/gfm/#html-comment).

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -9,24 +9,32 @@ assignees: ''
 
 ## Description
 
-__[ 1 paragraph concisely describing the bug ]__
+<!-- [ 1 paragraph concisely describing the bug ] -->
 
 ## Impact
 
-__[ 1 sentence detailing the impact this bug is creating for you ]__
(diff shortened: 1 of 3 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -9,24 +9,32 @@ assignees: ''
 
 ## Description
 
-__[ 1 paragraph concisely describing the bug ]__
+<!-- [ 1 paragraph concisely describing the bug ] -->
```
jeremycline, line 12: The [ ] stuff isn't necessary for it to be a comment, but I don't know if people find that visually pleasing or not. I have no personal opinion one way the another.

t-lo (approved):
This is nice, thank you Jeremy!

PR #48, opened 2024-03-01
tests: Only call az login if there's no access token available

jeremycline:
This changes the functional test script to test if the user is already logged in to Azure so that it doesn't unconditionally stop and prompt the user to log in.

## How to use

Log out via `az logout`, run the script, log back in, and then run the script again.

## Testing done
```
az logout
make e2e-test
make e2e-test
```


dongsupark (approved):
Tested, it works well. Thanks!

PR #55, opened 2024-03-07
Set a user-agent and timeout to the HTTP client

jeremycline:
There's two commits in this series, the first adjusts the error handling used in the `azure-init` binary to make clearer to read through. The second adjusts the `libazureinit` APIs that make use of the `reqwest::Client` to accept it as an argument. This allows the caller to control the configuration. Finally, the client is configured in `azure-init` to have a user-agent along with a timeout.

## How to use

Beyond building and running the existing tests, there's not yet tests to assert this actually works. The functional test looks to replicate the `azure-init` binary, so I think it would be interesting to have the test script copy over the actual binary and then make make some assertions about the queries it issues to the metadata service. I assume there's a way to see the requests a guest issues, but need to read up on that.

## Testing done

I compiled it and ran the functional test, but that doesn't make any assertions about the change here.

### DIFF
--- a/tests/functional_tests.rs
+++ b/tests/functional_tests.rs
@@ -3,13 +3,25 @@
 
 use libazureinit::distro::{Distribution, Distributions};
 use libazureinit::imds::PublicKeys;
-use libazureinit::{goalstate, user};
+use libazureinit::{
+    goalstate,
+    reqwest::{header, Client},
+    user,
+};
 
 use std::env;
 
 #[tokio::main]
 async fn main() {
     let cli_args: Vec<String> = env::args().collect();
+    let mut default_headers = header::HeaderMap::new();
+    let user_agent = header::HeaderValue::from_str("azure-init").unwrap();
+    default_headers.insert(header::USER_AGENT, user_agent);
+    let client = Client::builder()
+        .timeout(std::time::Duration::from_secs(30))
+        .default_headers(default_headers)
+        .build()
+        .unwrap();
 
     println!();
     println!("**********************************");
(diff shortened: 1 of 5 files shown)


tests/functional_tests.rs
```diff
@@ -48,9 +40,7 @@ async fn main() {
         username.as_str()
     );
 
-    Distributions::from("ubuntu")
-        .create_user(username.as_str(), "")
-        .expect("Failed to create user");
+    Distributions::from("ubuntu").create_user(username.as_str(), "")?;
```
dongsupark, line 55: Is there a way to keep an error context of error message `Failed to create user`?
Simply dropping the message looks a little backwards.

If `anyhow` is imported in the code base, we could have `.context()`, but that is a long term goal for now.

tests/functional_tests.rs
```diff
@@ -61,20 +51,14 @@ async fn main() {
     );
 
     Distributions::from("ubuntu")
-        .create_user("test_user_2", "azureProvisioningAgentPassword")
-        .expect("Failed to create user");
+        .create_user("test_user_2", "azureProvisioningAgentPassword")?;
```
dongsupark, line 66: the same here

tests/functional_tests.rs
```diff
@@ -99,14 +83,14 @@ async fn main() {
     println!();
     println!("Attempting to set the VM hostname");
 
-    Distributions::from("ubuntu")
-        .set_hostname("test-hostname-set")
-        .expect("Failed to set hostname");
+    Distributions::from("ubuntu").set_hostname("test-hostname-set")?;
```
dongsupark, line 98: the same here

dongsupark (commented):
The direction in general looks good to me.

See below.

tests/functional_tests.rs
dongsupark: Yes, let's split functional tests into a separate PR.

dongsupark (approved):
Looks good. Thanks

PR #56, opened 2024-03-12
Use safe nix interface to retrieve user and group id

jeremycline:
Using libc APIs directly is difficult and should be avoided if possible. In this case, the manual for getpwnam(3) and getgrnam(3) indicate they return NULL if the entry is not found or an error occurs. Since the return value isn't being checked for NULL here, a null pointer dereference can occur here.

Fortunately, the nix crate provides a wrapper for retrieving user details from a username.

I've added some calls to unwrap() which I'll remove once more error handling is in place.

### DIFF
--- a/libazureinit/Cargo.toml
+++ b/libazureinit/Cargo.toml
@@ -19,6 +19,9 @@ serde_json = "1.0.96"
 nix = "0.26.2"
 libc = "0.2.146"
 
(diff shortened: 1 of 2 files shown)


anhvoms (approved):
LGTM!

dongsupark (approved):

Issue #52, opened 2024-03-07
[RFE] Disable provisioning with password


anhvoms:
## Current situation
azure-init allows customers to provision Linux VMs with an admin password. 

## Impact
Password is not as secure as ssh-key. Using password leaves the VM more vulnerable to brute-force attack.

## Ideal future situation
Not supporting password provisioning.

## **Implementation options

A couple options
1) Disable password support completely. Note that Azure does allow customers to provide password to provision VM. In that case azure-init should fail provisioning if password is given.
2) Allow the customer to choose to keep password support as a compile-time configurable option (but disable it by default)

dongsupark:
As discussed, let's bring back password provisioning only in libazureinit, as that could be used in the future for some cases.

We also have to bring back {mount,unmount}_media parts that were dropped by accident.

PR #57, opened 2024-03-13
Disable provisioning with password

dongsupark:
Password authentication by itself not as secure as ssh key. For better security, we should disable password authentication. So if user enabled password authentication in Azure, azure-init now simply fails to provision.

Clean up unnecessary functions in libazureinit, like `mount_media`, `unmount_media`, `allow_password_authentication`.

Fixes https://github.com/Azure/azure-init/issues/52.

## Testing done

Manual test is done.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -10,6 +10,38 @@ use libazureinit::{
 
 const VERSION: &str = env!("CARGO_PKG_VERSION");
 
+fn get_username(
+    imds_body: String,
+) -> Result<String, Box<dyn std::error::Error>> {
+    if imds::get_provision_with_password(&imds_body).map_err(|_| {
+        "Failed to get disable password authentication".to_string()
+    })? {
+        // password authentication is disabled
+        match imds::get_username(imds_body.clone()) {
+            Ok(username) => Ok(username),
+            Err(_err) => Err("Failed to get username".into()),
+        }
+    } else {
+        // password authentication is enabled
+        let ovf_body = media::read_ovf_env_to_string().unwrap();
+        let environment = media::parse_ovf_env(ovf_body.as_str()).unwrap();
+
+        if !environment
+            .provisioning_section
+            .linux_prov_conf_set
+            .password
+            .is_empty()
+        {
+            return Err("password is non-empty".into());
+        }
+
+        Ok(environment
+            .provisioning_section
+            .linux_prov_conf_set
+            .username)
+    }
+}
+
 #[tokio::main]
 async fn main() {
     let mut default_headers = header::HeaderMap::new();
@@ -29,48 +61,17 @@ async fn main() {
         Err(_err) => return,
     };
 
-    let provision_with_password = imds::get_provision_with_password(&imds_body);
-    let disable_authentication = match provision_with_password {
-        Ok(disable_authentication) => disable_authentication,
+    let username = match get_username(imds_body.clone()) {
+        Ok(res) => res,
         Err(_err) => return,
     };
-
-    let username;
-    let mut password = "".to_owned();
-
-    if !disable_authentication {
-        media::make_temp_directory().unwrap();
-
-        media::mount_media();
-
-        let ovf_body = media::read_ovf_env_to_string().unwrap();
-        let environment = media::parse_ovf_env(ovf_body.as_str()).unwrap();
-
-        username = environment
-            .provisioning_section
-            .linux_prov_conf_set
-            .username;
-        password = environment
-            .provisioning_section
-            .linux_prov_conf_set
-            .password;
-
-        let _ = media::allow_password_authentication();
-
-        media::remove_media();
-    } else {
-        let username_result = imds::get_username(imds_body.clone());
-        username = match username_result {
-            Ok(username) => username,
-            Err(_err) => return,
-        };
-    }
 
     let mut file_path = "/home/".to_string();
     file_path.push_str(username.as_str());
 
+    // always pass an empty password
     Distributions::from("ubuntu")
-        .create_user(username.as_str(), password.as_str())
+        .create_user(username.as_str(), "")
         .expect("Failed to create user");
     let _create_directory =
         user::create_ssh_directory(username.as_str(), &file_path).await;
--- a/src/main.rs
+++ b/src/main.rs
@@ -58,12 +58,12 @@ async fn main() {
     let query_result = imds::query_imds(&client).await;
     let imds_body = match query_result {
         Ok(imds_body) => imds_body,
-        Err(_err) => return,
+        Err(_err) => std::process::exit(exitcode::CONFIG),
     };
 
     let username = match get_username(imds_body.clone()) {
         Ok(res) => res,
-        Err(_err) => return,
+        Err(_err) => std::process::exit(exitcode::CONFIG),
     };
 
     let mut file_path = "/home/".to_string();
@@ -79,7 +79,7 @@ async fn main() {
     let get_ssh_key_result = imds::get_ssh_keys(imds_body.clone());
     let keys = match get_ssh_key_result {
         Ok(keys) => keys,
-        Err(_err) => return,
+        Err(_err) => std::process::exit(exitcode::CONFIG),
     };
 
     file_path.push_str("/.ssh");
@@ -89,7 +89,7 @@ async fn main() {
     let get_hostname_result = imds::get_hostname(imds_body.clone());
     let hostname = match get_hostname_result {
         Ok(hostname) => hostname,
-        Err(_err) => return,
+        Err(_err) => std::process::exit(exitcode::CONFIG),
     };
 
     Distributions::from("ubuntu")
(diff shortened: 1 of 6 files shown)


src/main.rs
```diff
@@ -29,42 +29,22 @@ async fn main() {
         Err(_err) => return,
     };
 
-    let provision_with_password = imds::get_provision_with_password(&imds_body);
-    let disable_authentication = match provision_with_password {
+    if match imds::get_provision_with_password(&imds_body) {
```
anhvoms, line 32: The user can set disablePasswordAuthentication to true but provide an ssh-key and not password. We shouldn't fail provisioning in this case. 
Provisioning should only fail if we can detect that there's a password to be provisioned. The only way to know for certain that password is provided is by inspecting the ovf-env.xml

src/main.rs
```diff
@@ -29,42 +29,22 @@ async fn main() {
...
-            .provisioning_section
-            .linux_prov_conf_set
-            .username;
-        password = environment
-            .provisioning_section
-            .linux_prov_conf_set
-            .password;
+    } {
+        // fail if password authentication is enabled.
+        return;
```
anhvoms, line 37: Ideally we should report the error back to the platform so that provisioning can fail. Otherwise, the user will wait for 20 minutes and get a provisioning timeout, without a clear indication that provisioning failed because of password usage. See [#58](https://github.com/Azure/azure-init/issues/58)

Given that the change to report failure can be scoped to itself, I am ok with merging this one before tackling the other one in a separate PR.

anhvoms (changes requested):

src/main.rs
dongsupark: Thanks. This part was done by splitting the username/password part into a separate function `get_username`.

src/main.rs
dongsupark: Thanks, will try to address the error reporting issue later in a separate PR.

src/main.rs
```diff
@@ -10,6 +10,46 @@ use libazureinit::{
...
+) -> Result<String, Box<dyn std::error::Error>> {
+    let username;
+    let password;
+
+    if match imds::get_provision_with_password(&imds_body) {
+        Ok(disable_passwd_auth) => disable_passwd_auth,
+        Err(_err) => {
+            return Err("Failed to get disable password authentication".into())
+        }
+    } {
```
jeremycline, line 24: You could propagate the error here with `?` rather than matching, although the function being called only returns `Ok` at the moment (or panics). If you want to adjust the error type, you can achieve the same thing by using `map_err` like this:

```suggestion
    if imds::get_provision_with_password(&imds_body).map_err(|_err| {
        "Failed to get disable password authentication".to_string()
    })? {
```

This is totally a style nit since it achieves the same thing, but these feel a bit more "Rust-y" to me.

Maybe not something to change in this PR, but the function name is also confusing since `get_provision_with_password() -> true` when `disablePasswordAuthentication == true` and I would expect the opposite based on the name.

src/main.rs
```diff
@@ -10,6 +10,46 @@ use libazureinit::{
...
+            .linux_prov_conf_set
+            .username;
+        password = environment
+            .provisioning_section
+            .linux_prov_conf_set
+            .password;
+
+        if !password.is_empty() {
+            return Err("password is non-empty".into());
+        }
```
jeremycline, line 47: Since you're only using `password` to check if it's empty you could replace this with
```suggestion
        if environment.provisioning_section.linux_prov_conf_set.password.is_empty() {
            return Err("password is non-empty".into());
        }
```

Then you could drop the `let password;` at the start of the function since it's not needed in the outer scope.

If you wanted, you could also drop the `let username;` as well and just return the username from both branches of the `if/else` rather than binding it to the variable.

src/main.rs
```diff
@@ -29,42 +69,12 @@ async fn main() {
...
-
-        media::remove_media();
-    } else {
-        let username_result = imds::get_username(imds_body.clone());
-        username = match username_result {
-            Ok(username) => username,
-            Err(_err) => return,
-        };
-    }
+    let password = "".to_owned();
```
jeremycline, line 77: Since password is only provided to the `Distribution::create_user` function it doesn't need to be a `String`:
```suggestion
    let password = "";
```

Of course, if we're not ever going to have passwords the API should probably not even take a `password` argument.

jeremycline (commented):
Just a few style nits which you should feel free to ignore if you like 😄 

It would be nice to have a functional test that attempts to provision a host with a password and demonstrate it fails, as well, but that might need a bit more thought on how testing should be organized and whatnot.

src/main.rs
dongsupark: `map_err` sounds like a good idea.

And you are right. I have been also confused by the naming all over the place, which is actually the opposite of the actual meaning. But I did not want to touch too many things. ;-)

anhvoms: I would go ahead and change the function name to is_password_authentication_disabled

src/main.rs
dongsupark: Done. Removed username/password variables, and reduced number of lines as much as possible.

src/main.rs
dongsupark: Done. Simply removed the `password` variable, and passed `""`.

src/main.rs
dongsupark: Done, changed to `map_err`.

dongsupark: renamed the function name to `is_password_authentication_disabled` as well.

dongsupark:
Tested this PR manually in Azure, with help of the existing script `demo/image_creation.sh`.
(Though some tweak was needed for running that script.)

A provisioning with a ssh key seems to work well, at least no regression. On that machine, a systemd unit `azure-init.service` runs well, with an empty password given.

I did not yet manage to actually test the case of failing when a non-empty password.

src/main.rs
```diff
@@ -10,6 +10,38 @@ use libazureinit::{
...
+    imds_body: String,
+) -> Result<String, Box<dyn std::error::Error>> {
+    if imds::is_password_authentication_disabled(&imds_body).map_err(|_| {
+        "Failed to get disable password authentication".to_string()
+    })? {
+        // password authentication is disabled
+        match imds::get_username(imds_body.clone()) {
+            Ok(username) => Ok(username),
+            Err(_err) => Err("Failed to get username".into()),
+        }
```
jeremycline, line 23: This was a case I missed that could also be a `map_err`, but it's up to you if you want to adjust it. I'm touching all the errors in #59 anyway and need to rebase it after this is merged so I can adjust it there.

jeremycline (approved):
It looks good to me.

src/main.rs
dongsupark: It looks a little tricky touch that in this PR, because `main` does not return error in this context.
It might be probably better to touch in your PR.

dongsupark:
> I did not yet manage to actually test the case of failing when a non-empty password.

Also tested the scenario, by passing `--admin-password ****` to `az vm create` command.
It works as expected.

anhvoms:
> Tested this PR manually in Azure, with help of the existing script `demo/image_creation.sh`. (Though some tweak was needed for running that script.)
> 
> A provisioning with a ssh key seems to work well, at least no regression. On that machine, a systemd unit `azure-init.service` runs well, with an empty password given.
> 
> I did not yet manage to actually test the case of failing when a non-empty password.

You can try to deploy a VM with boot diagnostic enabled, using password. The deployment will fail in about 20 minutes due to OS Provisioning timeout (because the agent returned an error). Assuming the output from azure-init makes it to serial console (if not we should make any err/warn/into output go to console to aid with debugging) we should see the error about non-empty pasword

dongsupark:
> You can try to deploy a VM with boot diagnostic enabled, using password. The deployment will fail in about 20 minutes due to OS Provisioning timeout (because the agent returned an error). Assuming the output from azure-init makes it to serial console (if not we should make any err/warn/into output go to console to aid with debugging) we should see the error about non-empty pasword

Thanks for the tip.
It turns out azure-init simply returned success even on failure, so azure-init systemd unit looked ok.
With the latest commit I pushed, now azure-init systemd unit fails when it should fail.

However, provisioning itself seems to run even when the systemd unit failed.
That's why I could not see such a scenario of timeout with 20 min upon provisioning failure.
I assume it has something to do with the [issue](https://github.com/Azure/azure-init/issues/58) you created a few days ago, right?
I think that's all I can do for now with this PR.

anhvoms (approved):

anhvoms:
> > You can try to deploy a VM with boot diagnostic enabled, using password. The deployment will fail in about 20 minutes due to OS Provisioning timeout (because the agent returned an error). Assuming the output from azure-init makes it to serial console (if not we should make any err/warn/into output go to console to aid with debugging) we should see the error about non-empty pasword
> 
> Thanks for the tip. It turns out azure-init simply returned success even on failure, so azure-init systemd unit looked ok. With the latest commit I pushed, now azure-init systemd unit fails when it should fail.
> 
> However, provisioning itself seems to run even when the systemd unit failed. That's why I could not see such a scenario of timeout with 20 min upon provisioning failure. I assume it has something to do with the [issue](https://github.com/Azure/azure-init/issues/58) you created a few days ago, right? I think that's all I can do for now with this PR.

I think we should investigate why provisioning didn't timeout. Likely there's another entity/agent in the VM that was reporting health. It would be good to know what it is and if it is expected. We might want to remove such noise in our testing. It shouldn't block this PR, however. See #61

PR #59, opened 2024-03-13
Add errors to the library and remove all unwrap()/expect() calls

jeremycline:
Rather than panicking in library calls, return errors. These errors are definitely not perfect (they may be more or less fine-grained than library users need) but are intended to be a place to start.

Note: I'll rebase this once #57 is merged since it'll conflict with that.

### DIFF
--- a/libazureinit/src/imds.rs
+++ b/libazureinit/src/imds.rs
@@ -10,6 +10,8 @@ use serde::Deserialize;
 use serde_json;
 use serde_json::Value;
 
+use crate::error::Error;
+
 #[derive(Debug, Deserialize, PartialEq, Clone)]
 pub struct PublicKeys {
     #[serde(rename = "keyData")]
@@ -18,9 +20,7 @@ pub struct PublicKeys {
     pub path: String,
 }
 
-pub async fn query_imds(
-    client: &Client,
-) -> Result<String, Box<dyn std::error::Error>> {
+pub async fn query_imds(client: &Client) -> Result<String, Error> {
     let url = "http://169.254.169.254/metadata/instance?api-version=2021-02-01";
     let mut headers = HeaderMap::new();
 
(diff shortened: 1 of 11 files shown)


dongsupark:
rebased needed, thanks. :smiley:

libazureinit/src/media.rs
```diff
@@ -69,40 +71,37 @@ pub fn make_temp_directory() -> Result<(), Box<dyn std::error::Error>> {
...
 
-    if environment
+    if !environment
         .provisioning_section
         .linux_prov_conf_set
         .password
         .is_empty()
     {
-        return Err("Password is empty".into());
+        Err(Error::NonEmptyPassword)
```
jeremycline, line 101: Elsewhere we were returning an error if it _was_ empty and users of this function would immediately test that the password was empty so this is a behavior change.

libazureinit/src/imds.rs
```diff
@@ -32,64 +32,47 @@ pub async fn query_imds(
...
-    imds_body: String,
-) -> Result<Vec<PublicKeys>, Box<dyn std::error::Error>> {
-    let data: Value = serde_json::from_str(&imds_body)
-        .expect("Failed to parse the IMDS JSON.");
+pub fn get_ssh_keys(imds_body: String) -> Result<Vec<PublicKeys>, Error> {
+    let data: Value = serde_json::from_str(&imds_body)?;
     let public_keys =
-        Vec::<PublicKeys>::deserialize(&data["compute"]["publicKeys"])
-            .expect("Failed to deserialize the public ssh keys.");
+        Vec::<PublicKeys>::deserialize(&data["compute"]["publicKeys"])?;
```
dongsupark, line 47: If we drop the error message here, should we then add it instead on the caller side, in `provision()`, when `get_ssh_keys()` returned? Like:
```
let keys = imds::get_ssh_keys(imds_body.clone()).with_context(|| "Failed to get public ssh keys.")?;
```

libazureinit/src/imds.rs
```diff
@@ -32,64 +32,47 @@ pub async fn query_imds(
...
-    imds_body: String,
-) -> Result<String, Box<dyn std::error::Error>> {
-    let data: Value = serde_json::from_str(&imds_body)
-        .expect("Failed to parse the IMDS JSON.");
+pub fn get_username(imds_body: String) -> Result<String, Error> {
+    let data: Value = serde_json::from_str(&imds_body)?;
     let username =
-        String::deserialize(&data["compute"]["osProfile"]["adminUsername"])
-            .expect("Failed to deserialize the admin username.");
+        String::deserialize(&data["compute"]["osProfile"]["adminUsername"])?;
```
dongsupark, line 55: Same. Should we then add it instead on the caller side, in `provision()`, when `get_username()` returned? Like:
```
let username = get_username(imds_body.clone()).with_context(|| "Failed to get the admin username.")?;
```

libazureinit/src/imds.rs
```diff
@@ -32,64 +32,47 @@ pub async fn query_imds(
...
-    imds_body: String,
-) -> Result<String, Box<dyn std::error::Error>> {
-    let data: Value = serde_json::from_str(&imds_body)
-        .expect("Failed to parse the IMDS JSON.");
+pub fn get_hostname(imds_body: String) -> Result<String, Error> {
+    let data: Value = serde_json::from_str(&imds_body)?;
     let hostname =
-        String::deserialize(&data["compute"]["osProfile"]["computerName"])
-            .expect("Failed to deserialize the hostname.");
+        String::deserialize(&data["compute"]["osProfile"]["computerName"])?;
```
dongsupark, line 63: same for `get_hostname`.

libazureinit/src/imds.rs
```diff
@@ -32,64 +32,47 @@ pub async fn query_imds(
...
-    let data: Value = serde_json::from_str(imds_body)
-        .expect("Failed to parse the IMDS JSON.");
+) -> Result<bool, Error> {
+    let data: Value = serde_json::from_str(imds_body)?;
 
     let provision_with_password = String::deserialize(
         &data["compute"]["osProfile"]["disablePasswordAuthentication"],
-    )
-    .expect("Failed to deserialize the provisioning type.");
+    )?;
```
dongsupark, line 75: same for `is_password_authentication_disabled`.

dongsupark (commented):
Thanks. I like this approach, the whole code looks much better now.
I am still reading the PR, but I left a few comments.

libazureinit/src/imds.rs
jeremycline: Indeed. I've slapped a bunch of `with_context` calls in the binary.

src/main.rs
```diff
@@ -43,69 +32,73 @@ fn get_username(
...
-        Err(_err) => std::process::exit(exitcode::CONFIG),
-    };
-
-    let report_health_result =
-        goalstate::report_health(&client, vm_goalstate).await;
-    match report_health_result {
-        Ok(report_health) => report_health,
-        Err(_err) => std::process::exit(exitcode::CONFIG),
-    };
+        .with_context(|| "Failed to set hostname".to_string())?;
```
dongsupark, line 94: nit: `.to_string()` is not needed.

dongsupark (approved):
Looks good to me. Thanks for the PR.
A minor thing below:

src/main.rs
jeremycline: Dropped it, thanks!

dongsupark (approved):

anhvoms (approved):
LGTM!

PR #60, opened 2024-03-14
rename azure-provisioning-agent.service to azure-init.service

dongsupark:
Rename systemd unit to `azure-init.service`.
It is remaining work from the last repo rename work.

## Testing done

Manually tested in Azure.


anhvoms (approved):
LGTM! Thanks

PR #62, opened 2024-03-19
bring back mount_media, remove_media

dongsupark:
Functions `mount_media` and `remove_media` were removed in https://github.com/Azure/azure-init/pull/57.
To parse a ovf_env file, however, both mount_media and remove_media are necessary.
Bring back the functions.

Adjust return types according to the new LibError interface as well.

TODO: unit test for the functions would be also nice.

## Testing done

TBD

### DIFF
--- a/libazureinit/src/media.rs
+++ b/libazureinit/src/media.rs
@@ -6,6 +6,8 @@ use std::fs::create_dir_all;
 use std::fs::File;
 use std::io::Read;
 use std::os::unix::fs::PermissionsExt;
+use std::path::PathBuf;
+use std::process::Command;
 
 use serde::Deserialize;
 use serde_xml_rs::from_str;
@@ -66,27 +68,92 @@ fn default_preprov_type() -> String {
     "None".to_owned()
 }
 
-pub fn make_temp_directory() -> Result<(), Box<dyn std::error::Error>> {
-    let file_path = "/run/azure-init/tmp/";
+pub const PATH_MOUNT_DEVICE: &str = "/dev/sr0";
+pub const PATH_MOUNT_POINT: &str = "/run/azure-init/media/";
 
-    create_dir_all(file_path)?;
+// Some zero-sized structs that just provide states for our state machine
+pub struct Mounted;
+pub struct Unmounted;
 
-    let metadata = fs::metadata(file_path)?;
-    let permissions = metadata.permissions();
-    let mut new_permissions = permissions.clone();
-    new_permissions.set_mode(0o700);
-    fs::set_permissions(file_path, new_permissions)?;
-
-    Ok(())
+pub struct Media<State = Unmounted> {
+    device_path: PathBuf,
+    mount_path: PathBuf,
+    state: std::marker::PhantomData<State>,
 }
 
-pub fn read_ovf_env_to_string() -> Result<String, Error> {
-    let file_path = "/run/azure-init/tmp/ovf-env.xml";
-    let mut file = File::open(file_path)?;
-    let mut contents = String::new();
-    file.read_to_string(&mut contents)?;
+impl Media<Unmounted> {
+    pub fn new(device_path: PathBuf, mount_path: PathBuf) -> Media<Unmounted> {
+        Media {
+            device_path,
+            mount_path,
+            state: std::marker::PhantomData,
+        }
+    }
 
-    Ok(contents)
+    pub fn mount(self) -> Result<Media<Mounted>, Error> {
+        create_dir_all(&self.mount_path)?;
+
+        let metadata = fs::metadata(&self.mount_path)?;
+        let permissions = metadata.permissions();
+        let mut new_permissions = permissions.clone();
+        new_permissions.set_mode(0o700);
+        fs::set_permissions(&self.mount_path, new_permissions)?;
+
+        let mount_status = Command::new("mount")
+            .arg("-o")
+            .arg("ro")
+            .arg(&self.device_path)
+            .arg(&self.mount_path)
+            .status()?;
+
+        if !mount_status.success() {
+            Err(Error::SubprocessFailed {
+                command: "mount".to_string(),
+                status: mount_status,
+            })
+        } else {
+            Ok(Media {
+                device_path: self.device_path,
+                mount_path: self.mount_path,
+                state: std::marker::PhantomData,
+            })
+        }
+    }
+}
+
+impl Media<Mounted> {
+    pub fn unmount(self) -> Result<(), Error> {
+        let umount_status =
+            Command::new("umount").arg(self.mount_path).status()?;
+        if !umount_status.success() {
+            return Err(Error::SubprocessFailed {
+                command: "umount".to_string(),
+                status: umount_status,
+            });
+        }
+
+        let eject_status =
+            Command::new("eject").arg(self.device_path).status()?;
+        if !eject_status.success() {
+            Err(Error::SubprocessFailed {
+                command: "eject".to_string(),
+                status: eject_status,
+            })
+        } else {
+            Ok(())
+        }
+    }
+
+    pub fn read_ovf_env_to_string(&self) -> Result<String, Error> {
+        let mut file_path = self.mount_path.clone();
+        file_path.push("ovf-env.xml");
+        let mut file =
+            File::open(file_path.to_str().unwrap_or(PATH_MOUNT_POINT))?;
+        let mut contents = String::new();
+        file.read_to_string(&mut contents)?;
+
+        Ok(contents)
+    }
 }
 
 pub fn parse_ovf_env(ovf_body: &str) -> Result<Environment, Error> {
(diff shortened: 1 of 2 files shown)


dongsupark:
It would be nice to have unit tests in this PR, but without root privileges, unit tests would fail.
According to my research, Rust does not have a clear and nice way of mocking unit tests like Python.
If we really want to do that in Rust, in theory we could create mount-specific unit tests that can be executed only when `cargo test` runs with privileges.
On the other hand, I think in general running mount/umount commands on system devices should be rather in integration tests, not in unit tests.

cjp256:
I agree that anything running mount/unmount should be an integration test.  Unit tests should not require privilege or necessarily even run on the "target" system.

libazureinit/src/media.rs
```diff
@@ -66,7 +67,47 @@ fn default_preprov_type() -> String {
...
+            status: mount_status,
+        });
+    }
+
+    Ok(())
+}
+
+pub fn remove_media() -> Result<(), Error> {
+    let umount_status = Command::new("umount")
+        .arg("/run/azure-init/tmp/")
```
cjp256, line 90:
```suggestion
        .arg("/run/azure-init/media/")
```

dongsupark: Fixed, and created a new const `PATH_MOUNT_MEDIA`, and replaced hard-coded ones with the new const.

libazureinit/src/media.rs
```diff
@@ -66,8 +67,49 @@ fn default_preprov_type() -> String {
...
-pub fn make_temp_directory() -> Result<(), Box<dyn std::error::Error>> {
-    let file_path = "/run/azure-init/tmp/";
+const PATH_MEDIA_MOUNT: &str = "/run/azure-init/media/";
+
+pub fn mount_media() -> Result<(), Error> {
+    let mount_status = Command::new("mount")
+        .arg("-o")
+        .arg("ro")
+        .arg("/dev/sr0")
+        .arg(PATH_MEDIA_MOUNT)
```
anhvoms, line 77: We have to create the folder PATH_MEDIA_MOUNT before mounting it right? When creating the folder, we should also restrict the read/write access to root only

anhvoms (changes requested):

libazureinit/src/media.rs
jeremycline: If you want to ensure it's impossible to mis-use this API (forgetting to call `make_temp_directory`, for example) you can use a pattern (which I saw in [Rust for Rustaceans](https://rust-for-rustaceans.com/)) which creates a little state machine and hides functions until the object is in the right state. For example:

```
// Some zero-sized structs that just provide states for our state machine
struct Mounted;
struct Unmounted;

struct Media<State = Unmounted> {
    mount_path: PathBuf,
    device_path: PathBuf,
    state: std::marker::PhantomData<State>,
}

impl Media<Unmounted> {
    pub fn new(device_path: PathBuf, mount_path: PathBuf) -> Media<Unmounted> {
        Media {
            mount_path,
            device_path,
            state: std::marker::PhantomData,
        }
    }

    // The only thing you can do with the Media struct is call mount which, if it succeeds,
    // moves it into the Mounted state. From there the only thing you can do with it is
    // call unmount(), but you could make other functionality available by adding to the
    // impl Media<Mounted> block.
    pub fn mount(self) -> Result<Media<Mounted>, Error> {
        create_dir_all(&self.mount_path)?;

        let metadata = fs::metadata(&self.mount_path)?;
        let permissions = metadata.permissions();
        let mut new_permissions = permissions.clone();
        new_permissions.set_mode(0o700);
        fs::set_permissions(&self.mount_path, new_permissions)?;

        let mount_status = Command::new("mount")
            .arg("-o")
            .arg("ro")
            .arg("/dev/sr0")
            .arg(PATH_MEDIA_MOUNT)
            .status()?;

        if !mount_status.success() {
            Err(Error::SubprocessFailed {
                command: "mount".to_string(),
                status: mount_status,
            })
        } else {
            Ok(Media {
                mount_path: self.mount_path,
                device_path: self.device_path,
                state: std::marker::PhantomData,
            })
        }
    }
}

impl Media<Mounted> {
    pub fn unmount(self) -> Result<(), Error> {
        let umount_status =
            Command::new("umount").arg(self.mount_path).status()?;
        if !umount_status.success() {
            return Err(Error::SubprocessFailed {
                command: "umount".to_string(),
                status: umount_status,
            });
        }

        let eject_status =
            Command::new("eject").arg(self.device_path).status()?;
        if !eject_status.success() {
            Err(Error::SubprocessFailed {
                command: "eject".to_string(),
                status: eject_status,
            })
        } else {
            Ok(())
        }
    }
}

```

dongsupark: @jeremycline Thanks. Rewrote the mount part as you suggested.

src/main.rs
```diff
@@ -9,21 +10,34 @@ use libazureinit::distro::{Distribution, Distributions};
...
         // password authentication is enabled
+        let mount_media = Media::new(
+            PathBuf::from(media::PATH_MOUNT_DEVICE),
+            PathBuf::from(media::PATH_MOUNT_POINT),
+        );
+        let mounted = mount_media
+            .mount()
+            .with_context(|| "Failed to mount media.")?;
+
         let ovf_body = media::read_ovf_env_to_string()?;
```
jeremycline, line 34: You could move this function to the `impl Media<Mounted>` so you can't call it without the media being mounted.

dongsupark: Thanks, done.

libazureinit/src/media.rs
```diff
@@ -64,6 +65,28 @@ fn default_preprov_type() -> String {
...
 }
 
+pub fn mount_media() {
+    let _mount_media = Command::new("mount")
+        .arg("-o")
+        .arg("ro")
+        .arg("/dev/sr0")
+        .arg("/run/azure-init/tmp/")
+        .status()
+        .expect("Failed to execute mount command.");
```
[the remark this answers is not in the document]
dongsupark: Agreed, and for now I defined the device as `const PATH_MOUNT_DEVICE`.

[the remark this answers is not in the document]
dongsupark: This is not the case any more with the current code.

libazureinit/src/media.rs
```diff
@@ -66,27 +68,94 @@ fn default_preprov_type() -> String {
...
+            device_path,
+            mount_path,
+            state: std::marker::PhantomData,
+        }
+    }
+
+    // The only thing you can do with the Media struct is call mount which, if it succeeds,
+    // moves it into the Mounted state. From there the only thing you can do with it is
+    // call unmount(), but you could make other functionality available by adding to the
+    // impl Media<Mounted> block.
```
jeremycline, line 96: This comment is outdated now that you added `read_ovf_env_to_string()`. I think you can just delete it and we can have a task to come through and document the library's API in a follow-up PR.

libazureinit/src/media.rs
```diff
@@ -66,27 +68,94 @@ fn default_preprov_type() -> String {
...
+                status: eject_status,
+            })
+        } else {
+            Ok(())
+        }
+    }
 
-    Ok(contents)
+    pub fn read_ovf_env_to_string() -> Result<String, Error> {
+        let file_path = PATH_MOUNT_POINT.to_owned() + "ovf-env.xml";
```
jeremycline, line 152:
```suggestion
    pub fn read_ovf_env_to_string(&self) -> Result<String, Error> {
        let file_path = self.mount_point.clone().push("ovf-env.xml");
```

It would be surprising if you created the `Media` struct with a different mount path and then the "default" one got used. This also requires you to have a mounted `Media` instance.

src/main.rs
```diff
@@ -9,21 +10,34 @@ use libazureinit::distro::{Distribution, Distributions};
...
-        let ovf_body = media::read_ovf_env_to_string()?;
+        let mount_media = Media::new(
+            PathBuf::from(media::PATH_MOUNT_DEVICE),
+            PathBuf::from(media::PATH_MOUNT_POINT),
+        );
+        let mounted = mount_media
+            .mount()
+            .with_context(|| "Failed to mount media.")?;
+
+        let ovf_body = Media::<Mounted>::read_ovf_env_to_string()?;
```
jeremycline, line 34:
```suggestion
        let ovf_body = mounted.read_ovf_env_to_string()?;
```

With the suggested change to make this function take a reference to self, you'll need to use the instance method.

jeremycline (changes requested):
There's one small change required to "force" users to have mounted the media before trying to read from it, but otherwise I think it's good.

libazureinit/src/media.rs
dongsupark: Deleted

libazureinit/src/media.rs
dongsupark: Right, thanks.

jeremycline (approved):
It looks good to me!

anhvoms (approved):

PR #77, opened 2024-04-02
Adding Periodic Run and Issue Creation for Clippy

peytonr18:
Adding weekly scheduled run mostly for the purposes of running clippy periodically.
Also amending the workflow so that if clippy discovers a warning in the codebase, it will create an issue for that run.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -39,9 +39,9 @@ jobs:
       - name: Run Clippy and redirect output to file
         uses: actions-rs/cargo@v1
         with:
-          command: cargo clippy --verbose -- --deny warnings | tee clippy.log
+          command: cargo clippy --verbose -- --deny warnings > clippy.log 2>&1 || true
       - name: run clippy pt 2
-        run: cargo clippy --verbose -- --deny warnings | tee clippy.log
+        run: cargo clippy --verbose -- --deny warnings > clippy.log 2>&1 || true
       - name: Create issue if clippy found warnings
         run: |
           if [ $(grep -c 'warning' clippy.log || true) != '0' ]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -34,8 +36,12 @@ jobs:
...
-          args: --verbose -- --deny warnings
+      - name: Run Clippy and redirect output to file
+        run: cargo clippy --verbose -- --deny warnings > clippy.log 2>&1 || true
+      - name: Create issue if clippy found warnings
+        run: |
+          if [ $(grep -c 'warning' clippy.log || true) != '0' ]; then
+            gh issue create --title "Clippy discovered warnings in the codebase" --body-file clippy.log
+          fi
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
dongsupark, line 47: It means, this GitHub repo must have a personal access token `GITHUB_TOKEN` configured for that. I do not have that privilege.

dongsupark (commented):
Thanks. The idea of periodic running look ok in general.

However I am wondering exactly what you want to achieve.
Do you want to run `cargo clippy` not when creating a PR, and only periodically?
Or do you want to run it both on PR creation and periodically?

In case of former, you would probably like to split the periodic part into a separate workflow.
In case of latter, running `cargo clippy` twice sounds a little superfluous to me.

.github/workflows/ci.yaml
peytonr18: Sorry for not updating this sooner - I was communicating with my team and I think we are going to back off the plan to create an issue for clippy warnings. There is a org-wide movement to move away from using tokens/secrets of any kind and introducing another one right now is probably not the best idea.

peytonr18: In terms of running clippy periodically, we can split it into a separate workflow if you think that would be best. I was planning on just having the entire CI workflow run every Sunday at midnight in addition to on PR creation / pushes to main just to minimize the number of workflows we have running. But, if we think it's best to just split it up, it's not hard to do and I can close this PR and just create a separate clippy workflow.

PR #78, opened 2024-04-05
Create clippy-linting.yml

peytonr18:
Creating separate workflow for running clippy linting tool.

At some point it might be worth generating some kind of email notification in ICM (We have one for a separate internal repo) to alert when something fails and/or uploading the clippy logs as artifacts to be able to review later on, but will hold off on that for now.


anhvoms (approved):
LGTM! Thanks

Issue #66, opened 2024-03-20
azure-init should not assume provisioning media is always at /dev/sr0


anhvoms:
## Description

Currently azure-init assumes that /dev/sr0 location is where the provisioning iso will surface. While this is true many of the times, it's not always the case. For example, in FreeBSD it will be at /dev/cd0. In some special environment, it might show up as /dev/vda1

The right mechanism to find the device is to enumerate block devices and process all devices that have fstype of iso9660 and udf. Because /dev/sr0 will be the correct choice for > 99% of cases, we might default to /dev/sr0 and fall back to enumeration if /dev/sr0 isn't the right choice. See cloud-init handling of this issue for [reference](https://github.com/canonical/cloud-init/blob/cbe5f3a119909d6af51e3f531088c71e54ccc3e6/cloudinit/sources/DataSourceAzure.py#L1920)
## Impact

azure-init won't be able to mount the provisioning iso if the iso isn't showing up at  /dev/sr0

## Expected behavior

azure-init should be able to find the provisioning iso in all Azure environments

anhvoms:
@dongsupark  
https://lib.rs/crates/block-utils

PR #79, opened 2024-04-09
libazureinit: get mounted CDROM device at runtime

dongsupark:
Get mounted device name of CDROM at runtime, instead of hard-coding `/dev/sr0`.

Fixes https://github.com/Azure/azure-init/issues/66

## Testing done

TBD

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,6 +12,10 @@ jobs:
     name: Build and test azure-init
     runs-on: ubuntu-latest
     steps:
+      - name: Install libudev
+        run: |
+          sudo apt update
+          sudo apt install -y libudev-dev
       - uses: actions/checkout@v4
       - uses: actions-rs/toolchain@v1
         with:
--- a/libazureinit/src/error.rs
+++ b/libazureinit/src/error.rs
@@ -29,4 +29,6 @@ pub enum Error {
     UserMissing { user: String },
     #[error("Provisioning a user with a non-empty password is not supported")]
     NonEmptyPassword,
+    #[error("Unable to get list of block devices")]
+    BlockUtils(#[from] block_utils::BlockUtilsError),
 }
--- a/libazureinit/src/media.rs
+++ b/libazureinit/src/media.rs
@@ -14,7 +14,7 @@ use serde_xml_rs::from_str;
 
 use crate::error::Error;
 
-#[derive(Debug, Deserialize, PartialEq, Clone)]
+#[derive(Debug, Default, Deserialize, PartialEq, Clone)]
 pub struct Environment {
     #[serde(rename = "ProvisioningSection")]
     pub provisioning_section: ProvisioningSection,
@@ -22,7 +22,7 @@ pub struct Environment {
     pub platform_settings_section: PlatformSettingsSection,
 }
 
-#[derive(Debug, Deserialize, PartialEq, Clone)]
+#[derive(Debug, Default, Deserialize, PartialEq, Clone)]
 pub struct ProvisioningSection {
     #[serde(rename = "Version")]
     pub version: String,
@@ -30,7 +30,7 @@ pub struct ProvisioningSection {
     pub linux_prov_conf_set: LinuxProvisioningConfigurationSet,
 }
 
-#[derive(Debug, Deserialize, PartialEq, Clone)]
+#[derive(Debug, Default, Deserialize, PartialEq, Clone)]
 pub struct LinuxProvisioningConfigurationSet {
     #[serde(rename = "UserName")]
     pub username: String,
@@ -40,7 +40,7 @@ pub struct LinuxProvisioningConfigurationSet {
     pub hostname: String,
 }
 
-#[derive(Debug, Deserialize, PartialEq, Clone)]
+#[derive(Debug, Default, Deserialize, PartialEq, Clone)]
 pub struct PlatformSettingsSection {
     #[serde(rename = "Version")]
     pub version: String,
@@ -48,7 +48,7 @@ pub struct PlatformSettingsSection {
     pub platform_settings: PlatformSettings,
 }
 
-#[derive(Debug, Deserialize, PartialEq, Clone)]
+#[derive(Debug, Default, Deserialize, PartialEq, Clone)]
 pub struct PlatformSettings {
     #[serde(default = "default_preprov", rename = "PreprovisionedVm")]
     pub preprovisioned_vm: bool,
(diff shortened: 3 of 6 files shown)


dongsupark:
CI fails because the `block-utils` crate depends on `libudev-dev` package, which is not installed by Default on the default Ubuntu runner.
[actions-rs cargo repo](https://github.com/actions-rs/cargo) is meanwhile archived, so not likely to support such a feature.
Will look into the path forwards tomorrow.

dongsupark:
Fixed the CI issue, it was easier than expected.

dongsupark:
Fixed a syntax error in clippy-linting, which was made by https://github.com/Azure/azure-init/pull/78.

.github/workflows/ci.yaml
```diff
@@ -12,6 +12,9 @@ jobs:
     name: Build and test azure-init
     runs-on: ubuntu-latest
     steps:
+      - name: Install libudev
+        run: |
+          sudo apt install -y libudev-dev
```
anhvoms, line 17: should always do an apt update unless we know that it has been done in an earlier step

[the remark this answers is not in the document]
dongsupark: `libudev-dev` is needed only when compiling.
On the other hand `libudev1` is needed at runtime, and that is already installed in the default Ubuntu runner.

dongsupark: Done, thanks

libazureinit/src/media.rs
```diff
@@ -71,6 +71,20 @@ fn default_preprov_type() -> String {
...
 
+const CDROM_VALID_FS: &[&str] = &["iso9660", "udf"];
+
+// Get a mounted device with any filesystem for CDROM
+pub fn get_mount_device() -> Result<String, Error> {
+    if let Some(device) = block_utils::get_mounted_devices()?
+        .iter()
+        .find(|dev| CDROM_VALID_FS.contains(&dev.fs_type.to_str()))
+    {
+        return Ok(device.fs_type.to_string());
```
anhvoms, line 82: This function should return a list of eligible devices and the caller should iterate through them until they find the device with ovf_env.xml file

anhvoms (changes requested):

src/main.rs
```diff
@@ -10,35 +10,52 @@ use libazureinit::distro::{Distribution, Distributions};
...
-            PathBuf::from(media::PATH_MOUNT_POINT),
-        );
-        let mounted = mount_media
-            .mount()
-            .with_context(|| "Failed to mount media.")?;
 
-        let ovf_body = mounted.read_ovf_env_to_string()?;
-        let environment = media::parse_ovf_env(ovf_body.as_str())?;
+        // list of CDROM devices that is available with possible filesystems.
+        let ovf_devices = media::get_mount_device()?;
```
cjp256, line 46: consider decoupling the mount/ovf logic out of get_username()

is the cdrom only going to mounted if we detect password enabled via IMDS?

libazureinit/src/media.rs
```diff
@@ -71,6 +71,22 @@ fn default_preprov_type() -> String {
...
+
+// Get a mounted device with any filesystem for CDROM
+pub fn get_mount_device() -> Result<Vec<String>, Error> {
+    let mut list_devices: Vec<String> = Vec::new();
+
+    while let Some(device) = block_utils::get_mounted_devices()?
+        .iter()
+        .find(|dev| CDROM_VALID_FS.contains(&dev.fs_type.to_str()))
+    {
+        list_devices.push(device.name.clone());
```
jeremycline, line 84:
```suggestion
        .into_iter()
        .find(|dev| CDROM_VALID_FS.contains(&dev.fs_type.to_str()))
    {
        list_devices.push(device.name);
```
You can consume the returned list rather than iterating over references by using `into_iter()`. Then you won't need to clone the `device.name` since you own the struct and never reference it again.

src/main.rs
```diff
@@ -10,35 +10,52 @@ use libazureinit::distro::{Distribution, Distributions};
...
+        // loop until it finds a correct device.
+        for dev in ovf_devices {
+            environment = match mount_parse_ovf_env(dev) {
+                Ok(env) => Some(env),
+                Err(_) => continue,
+            }
+        }
 
         Ok(environment
+            .ok_or(libazureinit::error::Error::BlockUtilsError)?
```
jeremycline, line 58: This error immediately gets turned into `anyhow::Error` (from the return type). Using a library's error outside of the library is somewhat unusual, and particularly in applications you're usually not matching on errors to handle them differently higher up the stack, you're just adding error context and exiting.

You can achieve the same thing by using  `anyhow::anyhow!("Unable to get list of block devices")` here and dropping the `BlockUtilsError` from the library enum:

```suggestion
            .ok_or_else(|| anyhow::anyhow!("Unable to get list of block devices"))?
```

Using `ok_or_else` here will switch from eager to lazy evaluation so you don't create an error if it isn't necessary.

libazureinit/src/error.rs
```diff
@@ -29,4 +29,8 @@ pub enum Error {
     UserMissing { user: String },
     #[error("Provisioning a user with a non-empty password is not supported")]
     NonEmptyPassword,
+    #[error("Unable to get list of block devices")]
+    BlockUtils(#[from] block_utils::BlockUtilsError),
+    #[error("Unable to get list of block devices")]
+    BlockUtilsError,
```
jeremycline, line 35: I'd recommend dropping this and using `anyhow::anyhow!` in the application. The library never returns this error so it would be confusing to a library user to have to consider how to handle it when it never occurs.

jeremycline (changes requested):

libazureinit/src/media.rs
dongsupark: Done, thanks.

libazureinit/src/error.rs
dongsupark: Done

src/main.rs
dongsupark: Sounds like a good idea.
I would say, let's do that in another PR, because I don't want to leave the bug fixes in other commits unmerged for so long time. Ok?

jeremycline (approved):
:sparkles:

src/main.rs
jeremycline: I agree it'd be good to move it out, and also that it's okay for it to be in another PR. I have a WIP branch tidying some of the APIs in this area anyway so I could probably add a commit moving things around here.

PR #82, opened 2024-05-09
imds: deserialize response to a structure

jeremycline:
Rather than pass the JSON string around to functions, deserialize it to a structure that can be used directly by the caller.

### DIFF
--- a/libazureinit/src/imds.rs
+++ b/libazureinit/src/imds.rs
@@ -6,21 +6,79 @@ use reqwest::header::HeaderMap;
 use reqwest::header::HeaderValue;
 use reqwest::Client;
 
-use serde::Deserialize;
+use serde::{Deserialize, Deserializer};
 use serde_json;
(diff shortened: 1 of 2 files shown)


libazureinit/src/imds.rs
```diff
@@ -133,71 +123,78 @@ mod tests {
...
-              }
+              },
+              "publicKeys": [
+                {
+                  "keyData": "ssh-rsa test_key1",
+                  "path": "/path/to/.ssh/authorized_keys"
+                },
+                {
+                    "keyData": "ssh-rsa test_key2",
+                    "path": "/path/to/.ssh/authorized_keys"
```
dongsupark, line 134: Can you please shift the 2 lines, both keyData and path, by 2 bytes, so that the indent can be aligned with `test_key1`?

dongsupark (approved):

Issue #81, opened 2024-04-26
[RFE] decouple mount/ovf logic out of get_username


dongsupark:
## Current situation

@cjp256 pointed out in a [comment](https://github.com/Azure/azure-init/pull/79#discussion_r1562526181) that the mount/ovf logic is in `get_username()`.
That is not ideal, as CDROM mount should be independent of authentication via IMDS.

## Ideal future situation

We should consider decoupling the mount/ovf logic out of `get_username()`.

PR #83, opened 2024-05-14
Add KVP Tracer and OpenTel

peytonr18:
Draft PR to discuss OpenTelemetry Tracer and next steps for making tracer async, refactoring , and instrumenting the rest of the repo

PR #84, opened 2024-05-17
split out mount/OVF logic into get_environment

dongsupark:
As mount/OVF logic should be separate from `get_username()`, split it out into a new function `get_environment()`.

Fixes https://github.com/Azure/azure-init/issues/81

### DIFF
--- a/libazureinit/src/media.rs
+++ b/libazureinit/src/media.rs
@@ -12,6 +12,8 @@ use std::process::Command;
 use serde::Deserialize;
 use serde_xml_rs::from_str;
 
+use tracing;
+
 use crate::error::Error;
 
 #[derive(Debug, Default, Deserialize, PartialEq, Clone)]
@@ -185,6 +187,24 @@ pub fn parse_ovf_env(ovf_body: &str) -> Result<Environment, Error> {
     } else {
         Ok(environment)
     }
+}
+
+// Mount the given device, get OVF environment data, return it.
+pub fn mount_parse_ovf_env(dev: String) -> Result<Environment, Error> {
+    let mount_media =
+        Media::new(PathBuf::from(dev), PathBuf::from(PATH_MOUNT_POINT));
+    let mounted = mount_media.mount().inspect_err(
+        |e| tracing::error!(error = ?e, "Failed to mount media."),
+    )?;
+
+    let ovf_body = mounted.read_ovf_env_to_string()?;
+    let environment = parse_ovf_env(ovf_body.as_str())?;
+
+    mounted.unmount().inspect_err(
+        |e| tracing::error!(error = ?e, "Failed to remove media."),
+    )?;
+
+    Ok(environment)
 }
 
 #[cfg(test)]
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,7 +1,6 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-use std::path::PathBuf;
 use std::process::ExitCode;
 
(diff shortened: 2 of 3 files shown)


src/main.rs
```diff
@@ -36,8 +36,25 @@ fn mount_parse_ovf_env(dev: String) -> Result<Environment, anyhow::Error> {
     Ok(environment)
 }
 
+fn get_environment() -> Result<Environment, anyhow::Error> {
```
jeremycline, line 39: I think this refactor is a fine improvement as-is.

However, it might make sense to move this functionality to the library as a convenience function. `mount_parse_ovf_env` is itself a bit of a convenience function for the media module. I think it would make sense to move `media::parse_ovf_env` to be in `impl Media<Mounted>`, probably combining it with `read_ovf_env_to_string` unless there's a good case someone wants the string rather than the parsed `Environment` object. The combined function could probably just be called `environment` or something like that.

Whether or not you want to take that on in this PR is up to you since I think this is clearer either way.

jeremycline (approved):

src/main.rs
dongsupark: Sounds like a good idea.
However when `parse_ovf_env` is in `impl Media<Mounted>`, then unit tests like `test_get_ovf_env_none_missing` would not be able to find the function.
Let me think.

dongsupark: I was not able to move `parse_ovf_env` into `impl Media<Mounted>`, due to the unit tests as mentioned above. That's not impossible to do in theory, like creating mount before testing in each unit test, etc., but even after that I hit other issues of permission denied, so seems that it is not worth doing so.

Instead, as you suggested, I moved `mount_parse_ovf_env` to libazureinit/media.

libazureinit/src/media.rs
```diff
@@ -187,6 +189,24 @@ pub fn parse_ovf_env(ovf_body: &str) -> Result<Environment, Error> {
     }
 }
 
+// Mount the given device, get OVF environment data, return it.
+pub fn mount_parse_ovf_env(dev: String) -> Result<Environment, anyhow::Error> {
+    let mount_media =
+        Media::new(PathBuf::from(dev), PathBuf::from(PATH_MOUNT_POINT));
+    let mounted = mount_media
+        .mount()
+        .with_context(|| "Failed to mount media.")?;
```
jeremycline, line 198: Ideally the library should exclusively return errors of type `libazureinit::error::Error` and library users can use the types to handle different failure cases. If the client needs to react to mounting vs unmounting failures we could add some additional members to the enum, otherwise we can just propagate the error with `?`.

While the stack trace would make it clear which thing failed, this might also be a good opportunity to introduce a little logging. PR #83 is pulling in `tracing`, but it should be a minor conflict to `cargo add tracing` in this PR and add a `tracing::error!("Failed to mount media.")` call here.

If we're okay bumping the minimum Rust version to 1.76, you can do it like this:
```
let mounted = mount_media
    .mount()
    .inspect_err(|e| tracing::error!(error = ?e, "Failed to mount media."))?;
```

Otherwise, we can use `map_err` and just return the same error:
```
let mounted = mount_media
    .mount()
    .map_err(|e| {
        tracing::error!(error = ?e, "Failed to mount media.");
        e
    })?;
```

dongsupark: Thanks, let's go for `inspect_err` with the minimum Rust 1.76.
It should be ok. Many distros seem to still have <= 1.75, but CI should run with the latest stable 1.78.

jeremycline (approved):
✨

PR #87, opened 2024-06-13
add Azure Linux support

SeanDougherty:
This PR adds support for Azure Linux to the 0.1.1 version of azure-init. I need to rebase to the current head of main, while I do this, I'm marking the PR as draft to facilitate discussion. 

A notable delta between distros is the addition of the "wheel" and "sudo" groups and removal of the other groups in useradd.  

Additionally, this PR adds the os-release crate to determine the OS azure-init is operating on during provisioning time.

Finally, Functional tests for AzureLinux are missing. An update of the functional test to support either Azure Linux OR Ubuntu is a goal. I'm happy to make this change to the functional test in this PR or in a separate PR at the maintainer's request.

### DIFF
--- a/libazureinit/src/distro.rs
+++ b/libazureinit/src/distro.rs
@@ -17,6 +17,7 @@ pub trait Distribution {
 pub enum Distributions {
     Debian,
     Ubuntu,
+    AzureLinux,
 }
 
 impl Distribution for Distributions {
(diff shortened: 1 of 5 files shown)


libazureinit/src/distro.rs
```diff
@@ -26,10 +27,78 @@ impl Distribution for Distributions {
...
+                                return Err("Failed to create user".to_string());
+                            }
+                        }
+                        Err(err) => return Err(err.to_string()),
+                    };
+                    println!("User created without password")
+                } else {
+                    let input = format!("{}:{}", username, password);
+                    
+                    let mut output = Command::new("chpasswd")
```
dongsupark, line 68: Sorry, I am not an expert of Azure Linux, but is there a specific reason why you need to do password-based authentication?

Previously we had the password authentication in the code base, but the PR https://github.com/Azure/azure-init/pull/57 removed that part mainly for security. Though there still seems to be cases where password authentication would be necessary. I am just curious what the situation of Azure Linux would be.

dongsupark (commented):
Thanks for creating the PR. Looks nice.
It is up to you to decide whether to write functional tests in this PR or not.

See below:

t-lo:
Hello Sean 👋 

Thank you for your work, great to see azure-init is useful for you! Just out of curiosity, could you please have a look at https://github.com/Azure/azure-init/pull/86 and check if that PR "magically" adds Azure-Linux support? We're trying to be distro-agnostic going forward, and it might well be Dongsu's PR will implicitly add support for Azure Linux (and a number of other distros, too).

If you need password authentication support for some reason (see Dongsu's question) I'm sure we'll find a way to work this into the generic approach.

libazureinit/src/distro.rs
SeanDougherty: No actually, I believe we do want to eliminate password auth, this was a hold over from me re-implementing the debian logic of an older branch 👍

Thank you for the catch

PR #88, opened 2024-06-20
improve image_creation script

dongsupark:
Improve `image_creation.sh` script.

* Make necessary variables configurable like resource group, storage account, location, vm_name, vm_size, etc.
* Consolidate storage accounts creation into a single account, to avoid defining custom connection string env variables. Instead, generate connection string from the storage account directly to pass it to the subsequent commands.
* Create storage account and container before getting SAS string.
* Create SIG gallery before running az sig image-version create.
* Convert variable names to capital letters.
* Create image-definition before creating image-version in an automated way.
* Explicitly set `SecurityType=TrustedLaunch` for vm create and sig image-definition create.
* Address warnings and infos when running shellcheck, mainly by adding quotes around variables.
* Add command line options -h for help and -v for verbose.

## How to run with custom parameters.
```
RG="my-test-azinit" LOCATION="westeurope" VM_SIZE="Standard_D2ds_v5" BASE_IMAGE="Debian:debian-11:11-backports-gen2:latest" demo/image_creation.sh
```
## Testing done
Done testing with a couple of possible variables.

### DIFF
--- a/demo/image_creation.sh
+++ b/demo/image_creation.sh
@@ -2,124 +2,186 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-set -e -u -x -o pipefail
+set -e -u -o pipefail
 
-epoch=$(date +%s)
-temp_dir=/tmp/staging.$epoch
-target_dir=azure-init
-staging_dir=$temp_dir/$target_dir
+USAGE="Usage: $0 [options]
+It is possible to pass in custom parameters like:
+
+RG=\"myrg\" LOCATION=\"westeurope\" VM_SIZE=\"Standard_D2ds_v5\" \\
+    BASE_IMAGE=\"Debian:debian-11:11-backports-gen2:latest\" \\
+    $0
+
+Options:
+    -v|--verbose    Print out debug message
+    -h|--help       This help message
+"
+
+while [[ $# -gt 0 ]] ; do
+    case "$1" in
+        -h|--help)
+            echo "$USAGE"
+            exit 0
+            ;;
+        -v|--verbose)
+            set -x
+	    shift
+            ;;
+        *)
+            break
+    esac
+done
+
+EPOCH=$(date +%s)
+TEMP_DIR=/tmp/staging.$EPOCH
+TARGET_DIR=azure-init
+STAGING_DIR=$TEMP_DIR/$TARGET_DIR
 echo "*********************************************************************"
 echo "Building the agent"
 echo "*********************************************************************"
 
 cargo build
 
-root_dir=$(git rev-parse --show-toplevel)
+ROOT_DIR=$(git rev-parse --show-toplevel)
 echo "*********************************************************************"
-echo "Staging artifacts to $staging_dir"
+echo "Staging artifacts to $STAGING_DIR"
 echo "*********************************************************************"
-mkdir -p $staging_dir
-cp $root_dir/target/debug/azure-init $staging_dir/
-cp $root_dir/config/azure-init.service $staging_dir/
-cp $root_dir/demo/customdata_template.yml $temp_dir/customdata.yml
+mkdir -p "$STAGING_DIR"
+cp "$ROOT_DIR"/target/debug/azure-init "$STAGING_DIR"/
+cp "$ROOT_DIR"/config/azure-init.service "$STAGING_DIR"/
+cp "$ROOT_DIR"/demo/customdata_template.yml "$TEMP_DIR"/customdata.yml
 echo "Done"
 
 echo "*********************************************************************"
 echo "Creating azure-init.tgz package for upload"
 echo "*********************************************************************"
 rm -f ./azure-init.tgz
-tar cvfz azure-init.tgz -C $temp_dir $target_dir
+tar cvfz azure-init.tgz -C "$TEMP_DIR" "$TARGET_DIR"
+echo "Done"
+
+RG="${RG:-testagent-$EPOCH}"
+STORAGE_ACCOUNT="${STORAGE_ACCOUNT:-azinitsa$EPOCH}"
+STORAGE_CONTAINER="${STORAGE_CONTAINER:-azinitcontainer}"
+LOCATION="${LOCATION:-eastus}"
+
+echo "*********************************************************************"
+echo "Creating resource group $RG"
+echo "*********************************************************************"
+az group create -g "$RG" -l "$LOCATION"
 echo "Done"
 
 echo "*********************************************************************"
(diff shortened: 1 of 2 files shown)


jeremycline (commented):
I tried this out with `RG="jcline-azinit-test2" LOCATION="eastus2" BASE_IMAGE="Debian:debian-11:11-backports-gen2:latest" demo/image_creation.sh` and the VM failed configuration with
```
[   33.134147] cloud-init[918]: /var/log/azure/image/setup.sh: line 14: /etc/netplan/eth0.yaml: No such file or directory
```

I assume /etc/netplan/ doesn't exist on that image.

Since you're not touching the `customdata_template.yml` this is just an existing issue from what I can tell. The changes themselves look good good to me. You might want to point `ShellCheck` at it, but all those were pre-existing so we can do that separately as well.

dongsupark:
> I tried this out with `RG="jcline-azinit-test2" LOCATION="eastus2" BASE_IMAGE="Debian:debian-11:11-backports-gen2:latest" demo/image_creation.sh` and the VM failed configuration with
> 
> ```
> [   33.134147] cloud-init[918]: /var/log/azure/image/setup.sh: line 14: /etc/netplan/eth0.yaml: No such file or directory
> ```
> 
> I assume /etc/netplan/ doesn't exist on that image.

I cannot reproduce it myself, but anyway pushed a fix for that in `customdata_template.yml`.

> 
> Since you're not touching the `customdata_template.yml` this is just an existing issue from what I can tell. The changes themselves look good good to me. You might want to point `ShellCheck` at it, but all those were pre-existing so we can do that separately as well.

anhvoms:
The image creation script was used with Ubuntu 20.04/22.04. It assumed that netplan was used for networking and a lot of code written was based on that assumption. See issue #90

dongsupark:
Added an automated creation of image definition before creating image version.
Addressed warnings and infos when running shellcheck.

demo/image_creation.sh
```diff
@@ -1,125 +1,162 @@
...
+TARGET_DISK=$(az disk show --ids "$(az vm show -g "$RG" -n "$VM_NAME" | jq .storageProfile.osDisk.managedDisk.id -r)" | jq .name -r)
+az snapshot create -g "$RG" -n "$VM_NAME"-snapshot --source "$TARGET_DISK"
+
+IMAGE_VERSION=$(date '+%Y.%m%d.%H%M%S')
+GALLERY_NAME=testgalleryazinit
+GALLERY_DEFINITION=testgalleryazinitdef
+GALLERY_RG=testgalleryazinitrg
+SNAPSHOT_ID=$(az snapshot show -n "$VM_NAME"-snapshot -g "$RG" --query id --output tsv)
+
+# If BASE_IMAGE == canonical:0001-com-ubuntu-server-jammy:22_04-lts-gen2:latest,
```
anhvoms, line 136: remove these commented lines

demo/image_creation.sh
```diff
@@ -1,125 +1,162 @@
...
-az sig image-version create -g $gallery_rg --gallery-name $gallery --gallery-image-definition $definition --gallery-image-version $version --os-snapshot $snapshot_id --target-regions "eastus" --replica-count 1
-if [[ $? -eq 0 ]]
+if az sig image-version create -g $GALLERY_RG --gallery-name $GALLERY_NAME --gallery-image-definition $GALLERY_DEFINITION --gallery-image-version "$IMAGE_VERSION" --os-snapshot "$SNAPSHOT_ID" --target-regions "$LOCATION" --replica-count 1;
 then
     echo "Image publishing finished"
     echo "Deleting staging resource group"
-    az group delete -g $rg --yes --no-wait
+    az group delete -g "$RG" --yes --no-wait
     echo "Done"
 fi
```
anhvoms, line 162: if the image version creation fails, we should print out some error, and remind the user that the resource group is kept for debugging purposes

demo/image_creation.sh
```diff
@@ -1,125 +1,162 @@
...
+
+az sig create --resource-group "$GALLERY_RG" --gallery-name "$GALLERY_NAME"
+
 echo "*********************************************************************"
-echo "Capturing OsDisk snapshot of vm $vm with image $image"
+echo "Creating image definition of $BASE_IMAGE"
 echo "*********************************************************************"
-target_disk=$(az disk show --ids $(az vm show -g $rg -n $vm | jq .storageProfile.osDisk.managedDisk.id -r) | jq .name -r)
-az snapshot create -g $rg -n $vm-snapshot --source $target_disk
+az sig image-definition create --resource-group "$GALLERY_RG" --gallery-name "$GALLERY_NAME" --gallery-image-definition "$GALLERY_DEFINITION" --offer "$GALLERY_OFFER" --publisher "$GALLERY_PUBLISHER" --sku "$GALLERY_SKU" --os-type linux --os-state Specialized --features SecurityType="$SECURITY_TYPE"
```
anhvoms, line 151: os state for the image definition should be Generalized and not specialized
The reason is that specialized deployment in Azure does not wait or need provisioning complete signal. The deployment of specialized image also does not allow passing in user/hostname etc...

demo/image_creation.sh
```diff
@@ -1,125 +1,162 @@
 #!/bin/bash
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
+#
```
anhvoms, line 4: Ideally this should go into a -h message, but I'm ok with improving it later

anhvoms (changes requested):

demo/image_creation.sh
dongsupark: Actually the comments are necessary descriptions including examples, not accidentally left.

Will instead change like below for better descriptions:
```
# If BASE_IMAGE is set to canonical:0001-com-ubuntu-server-jammy:22_04-lts-gen2:latest, then
# GALLERY_PUBLISHER becomes "canonical"
# GALLERY_OFFER becomes "0001-com-ubuntu-server-jammy"
# GALLERY_SKU becomes "22_04-lts-gen2"
```

dongsupark: Done

demo/image_creation.sh
dongsupark: Done

demo/image_creation.sh
dongsupark: Done

demo/image_creation.sh
dongsupark: Done, added command-line options and defined USAGE.

anhvoms (approved):
LGTM! Thank you.

PR #91, opened 2024-06-25
Implement a builder-style provisioning API

jeremycline:
This is based on top of PR https://github.com/Azure/azure-init/pull/86 so I'd recommend just looking at the last commit in the series. It is a first draft of a builder-style interface to allow library users to select the backend for tools when provisioning.

It's not ready to be merged since there's no tests, documentation, or
implementation of default behavior when no backend is explicitly
selected. Additionally, I've not gated backends behind feature flags,
either. I'd love feedback on the general approach and what to do by
default.

My personal opinion is that the default option should iterate through
all available backends until one succeeds.

### DIFF
--- a/libazureinit/src/provision/mod.rs
+++ b/libazureinit/src/provision/mod.rs
@@ -1,0 +1,179 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+pub mod hostname;
+pub mod password;
+pub(crate) mod ssh;
+pub mod user;
+
+use strum::IntoEnumIterator;
+use tracing::instrument;
+
+use crate::error::Error;
+use crate::User;
+
+/// The interface for applying the desired configuration to the host.
+///
+/// By default, all known tools for provisioning a particular resource are tried
+/// until one succeeds. Particular tools can be selected via the
+/// `*_provisioners()` methods ([`Provision::hostname_provisioners`],
+/// [`Provision::user_provisioners`], etc).
+///
+/// To actually apply the configuration, use [`Provision::provision`].
+#[derive(Clone)]
+pub struct Provision {
+    hostname: String,
+    user: User,
+    hostname_backends: Option<Vec<hostname::Provisioner>>,
+    user_backends: Option<Vec<user::Provisioner>>,
+    password_backends: Option<Vec<password::Provisioner>>,
+}
+
+impl Provision {
+    pub fn new(hostname: impl Into<String>, user: User) -> Self {
+        Self {
+            hostname: hostname.into(),
+            user,
+            hostname_backends: None,
+            user_backends: None,
+            password_backends: None,
+        }
+    }
+
+    /// Specify the ways to set the virtual machine's hostname.
+    ///
+    /// By default, all known methods will be attempted. Use this function to
+    /// restrict which methods are attempted. These will be attempted in the
+    /// order provided until one succeeds.
+    pub fn hostname_provisioners(
+        mut self,
+        backends: impl Into<Vec<hostname::Provisioner>>,
+    ) -> Self {
+        self.hostname_backends = Some(backends.into());
+        self
+    }
+
+    /// Specify the ways to create a user in the virtual machine
+    ///
+    /// By default, all known methods will be attempted. Use this function to
+    /// restrict which methods are attempted. These will be attempted in the
+    /// order provided until one succeeds.
+    pub fn user_provisioners(
+        mut self,
+        backends: impl Into<Vec<user::Provisioner>>,
+    ) -> Self {
+        self.user_backends = Some(backends.into());
+        self
+    }
+
+    /// Specify the ways to set a users password.
+    ///
+    /// By default, all known methods will be attempted. Use this function to
+    /// restrict which methods are attempted. These will be attempted in the
+    /// order provided until one succeeds. Only relevant if a password has been
+    /// provided with the [`User`].
+    pub fn password_provisioners(
+        mut self,
+        backend: impl Into<Vec<password::Provisioner>>,
+    ) -> Self {
+        self.password_backends = Some(backend.into());
+        self
+    }
+
+    /// Provision the host.
+    ///
+    /// Provisioning can fail if the host lacks the necessary tools. For example,
+    /// if there is no `useradd` command on the system's `PATH`, or if the command
+    /// returns an error, this will return an error. It does not attempt to undo
+    /// partial provisioning.
+    #[instrument(skip_all)]
+    pub fn provision(self) -> Result<(), Error> {
+        self.user_backends
+            .unwrap_or_else(|| user::Provisioner::iter().collect())
+            .iter()
+            .find_map(|backend| {
+                backend
+                    .create(&self.user)
+                    .map_err(|e| {
+                        tracing::info!(
+                            error=?e,
+                            backend=?backend,
+                            resource="user",
+                            "Provisioning did not succeed"
+                        );
+                        e
+                    })
+                    .ok()
+            })
+            .ok_or(Error::NoUserProvisioner)?;
+
+        self.password_backends
+            .unwrap_or_else(|| password::Provisioner::iter().collect())
+            .iter()
+            .find_map(|backend| {
+                backend
+                    .set(&self.user)
+                    .map_err(|e| {
+                        tracing::info!(
+                            error=?e,
+                            backend=?backend,
+                            resource="password",
+                            "Provisioning did not succeed"
+                        );
+                        e
+                    })
+                    .ok()
+            })
+            .ok_or(Error::NoPasswordProvisioner)?;
+
+        if !self.user.ssh_keys.is_empty() {
+            let user = nix::unistd::User::from_name(&self.user.name)?.ok_or(
+                Error::UserMissing {
+                    user: self.user.name,
+                },
+            )?;
+            ssh::provision_ssh(&user, &self.user.ssh_keys)?;
+        }
+
+        self.hostname_backends
+            .unwrap_or_else(|| hostname::Provisioner::iter().collect())
+            .iter()
+            .find_map(|backend| {
+                backend
+                    .set(&self.hostname)
+                    .map_err(|e| {
+                        tracing::info!(
+                            error=?e,
+                            backend=?backend,
+                            resource="hostname",
+                            "Provisioning did not succeed"
+                        );
+                        e
+                    })
+                    .ok()
+            })
+            .ok_or(Error::NoHostnameProvisioner)?;
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+
+    use crate::User;
+
+    use super::{hostname, password, user, Provision};
+
+    #[test]
+    fn test_successful_provision() {
+        let _p = Provision::new(
+            "my-hostname".to_string(),
+            User::new("azureuser", vec![]),
+        )
+        .hostname_provisioners([hostname::Provisioner::FakeHostnamectl])
+        .user_provisioners([user::Provisioner::FakeUseradd])
+        .password_provisioners([password::Provisioner::FakePasswd])
+        .provision()
+        .unwrap();
+    }
+}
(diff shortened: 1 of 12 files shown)


dongsupark:
Code looks good in general. :+1: 

> My personal opinion is that the default option should iterate through
all available backends until one succeeds.

Looks good to me.

anhvoms:
+1 on having the default iterate through all available backend and letting user pass in their own backend

jeremycline:
Okay, I've made a few changes and this should have all the things implemented (but not documented yet). At this point I probably could break some of these changes out into smaller PRs, like the ssh authorized_key tests/changes.

After a bunch of fiddling I decided I didn't love the "iterate through all the backends" approach and decided that the user needs to specify which ones they want to use. It makes the interface more verbose, and to be honest I don't really like what I've built, but my experience writing library APIs in Rust is not large.

However, what I've got is, I think, service-able and can do the things we want it to. I've added some feature flags to show how a user could restrict which backends get used.

libazureinit/src/provision/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+// State transitions for the Provision structure which are used to force
+// the user to provide backends for things they wish to provision.
+mod private {
+    #[derive(Default)]
+    pub struct NoBackends;
+    #[derive(Default)]
+    pub struct HostnameBackend;
+    #[derive(Default)]
+    pub struct PasswordBackend;
+    #[derive(Default)]
```
dongsupark, line 29: Maybe `UseraddBackend` is missing?

libazureinit/src/provision/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+impl Provision<NoBackends> {
+    pub fn new(hostname: String, username: String) -> Self {
+        Self {
+            hostname,
+            username,
+            ..Default::default()
+        }
+    }
+
+    pub fn hostname_provisioners(
```
dongsupark, line 54: Should `hostname_provisioners` be located under `impl Provision<HostnameBackend>`?

libazureinit/src/provision/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+            user_backends: self.user_backends,
+            password_backends: self.password_backends,
+            state: PhantomData,
+        }
+    }
+}
+
+impl Provision<HostnameBackend> {
+    pub fn user_provisioners(
+        mut self,
```
dongsupark, line 74: Should `user_provisioners` be located under `impl Provisioners<UseraddBackend>`?

dongsupark (commented):
I like the approach in general.
While I am in the middle of reading code, I left comments below:

libazureinit/src/provision/user.rs
```diff
@@ -0,0 +1,52 @@
...
+    let status = Command::new(path_useradd)
+                    .arg(name)
+                    .arg("--comment")
+                    .arg(
+                      "Provisioning agent created this user based on username provided in IMDS",
+                    )
+                    .arg("--groups")
+                    .arg("adm,audio,cdrom,dialout,dip,floppy,lxd,netdev,plugdev,sudo,video")
+                    .arg("-d")
+                    .arg(home_path.clone())
```
dongsupark, line 41: In case of building with Rust 1.70.0, stable 12 months ago, which will be introduced in https://github.com/Azure/azure-init/pull/94, clippy prints out the warning:

```
warning: redundant clone
  --> libazureinit/src/provision/user.rs:41:35
   |
41 |                     .arg(home_path.clone())
   |                                   ^^^^^^^^ help: remove this
   |
note: this value is dropped without further use
  --> libazureinit/src/provision/user.rs:41:26
   |
41 |                     .arg(home_path.clone())
   |                          ^^^^^^^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone
   = note: `#[warn(clippy::redundant_clone)]` on by default
```

libazureinit/src/provision/mod.rs
```diff
@@ -0,0 +1,342 @@
...
+                    })
+                    .ok()
+            })
+            .ok_or(Error::NoPasswordProvisioner)?;
+
+        if !self.keys.is_empty() {
+            let user = nix::unistd::User::from_name(&self.username)?
+                .ok_or_else(|| Error::UserMissing {
+                    user: self.username,
+                })?;
```
dongsupark, line 163: In case of building with Rust 1.70.0, clippy prints out the warning:

```
warning: unnecessary closure used to substitute value for `Option::None`
   --> libazureinit/src/provision/mod.rs:160:24
    |
160 |               let user = nix::unistd::User::from_name(&self.username)?
    |  ________________________^
161 | |                 .ok_or_else(|| Error::UserMissing {
162 | |                     user: self.username,
163 | |                 })?;
    | |__________________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_lazy_evaluations
    = note: `#[warn(clippy::unnecessary_lazy_evaluations)]` on by default
help: use `ok_or(..)` instead
    |
161 ~                 .ok_or(Error::UserMissing {
162 +                     user: self.username,
163 ~                 })?;
    |
```

libazureinit/src/provision/mod.rs
jeremycline: So this is one of the things I didn't really like about my approach - some arguments (user, hostname) are currently required as they're part of `new`, whereas some arguments (currently just password) are not necessarily required. So the state transitions are a little weird as a consequence of that + not having a default for the backends. We could make the hostname and username behave like password so usage would be like:

```
Provision::new()
    .hostname("my-host")
    .hostname_backends(...)
    .user("my-user")
    .username_backends(...)
    .password("my-password")
    .password_backends(...)
    .provision();
```

Also, I think the above interface (and what I've currently implemented) only really makes sense if you're not required to specify a backend, in which case the state transitions are kind of pointless anyway. This would be a simpler interface if we required explicitly selecting backends:

```
Provision::new()
    .hostname("my-host", &[backend1, backend2,])
    .user("my-user", &[backend1, backend2])
    .password("my-password", &[backend1, backend2])
    .provision();
```

I go back and forth on having the user be explicit about how to do the thing. One thing I don't like about what I've currently got is there's no interface for users to say "do it all, and any new backends the library implements in the future". That being the default makes it possible, but we could also add a thing for the user to explicitly opt-in to that. What do you think?

libazureinit/src/provision/user.rs
jeremycline: I'm really surprised that clippy _isn't_ warning about this on newer versions...

dongsupark: > I'm really surprised that clippy _isn't_ warning about this on newer versions...

Starting from Rust clippy [1.71](https://github.com/rust-lang/rust-clippy/blob/master/CHANGELOG.md#rust-171), `redundant clone` has been [moved](https://github.com/rust-lang/rust-clippy/pull/10873) to `nursery`(a.k.a. under development), because of a bunch of false positives in 1.70. That's the reason.

libazureinit/src/provision/mod.rs
dongsupark: Oh, thanks for the detailed explanation.
If I would have to choose between the 2 examples, i.e. complicated interface and a simpler interface, then I would choose the latter, a simpler one.
I have no better idea at the moment.

libazureinit/src/provision/user.rs
```diff
@@ -0,0 +1,52 @@
...
+    let home_path = format!("/home/{name}");
+
+    let status = Command::new(path_useradd)
+                    .arg(name)
+                    .arg("--comment")
+                    .arg(
+                      "Provisioning agent created this user based on username provided in IMDS",
+                    )
+                    .arg("--groups")
+                    .arg("adm,audio,cdrom,dialout,dip,floppy,lxd,netdev,plugdev,sudo,video")
```
SeanDougherty, line 39: For Azure Linux, we will want to specify an alternative list of default groups for the user ([example](https://github.com/Azure/azure-init/pull/87/files#diff-be2f6d40c6bbf8fc0882a218ec29f5579d7b1d83c4449d9efe6814c64b6ba3e3R42)).

There's a few different ways to specify this, but perhaps an environment variable is the most logical? As that is how we are handling distro-specific knowledge in the other situations (i.e., binary paths).



To illustrate current behavior, here is a test trying to call path_useradd with this `--groups` list, and there are a few groups that aren't recognized on AzL. (See photo)
image
_*Note: this test was built off of #100 so there may be some additional differences in the code, but the result of calling `useradd` should be the same as at the time of this writing both PRs use the [same ](https://github.com/Azure/azure-init/blob/5171131d549c29db7d465368dc9d407024b27830/libazureinit/src/distro.rs#L19)--groups list_

libazureinit/src/provision/mod.rs
jeremycline: Okay, I decided to go back to the "if you don't provide a specific list of backends we'll try everything" approach, so a minimal use of the API would be:
```
# minimal usecase
Provision::new("my-host", "my-user", [pubkey1, ...]).provision();

# more specific
Provision::new("my-host", "my-user", [pubkey1, ...])
    .password("my-password")
    .hostname_backends([HostnameBackend::Hostnamectl])
    .provision();
```

I decided to make providing SSH keys mandatory (although you can provide an empty list if you really want).

libazureinit/src/provision/user.rs
jeremycline: Hm, yes. I added a commit to this PR that sets it via environment variable, but I'll mull on it more over the weekend.

I do worry about the discoverability of this approach. I think it's a little weird to have a library accept configuration this way instead of exposing an API and letting the application using the library handle it (either via build-time environment variables, or CLI options, or what have you).

libazureinit/src/provision/mod.rs
```diff
@@ -0,0 +1,186 @@
...
+    user_backends: Option<Vec<user::Provisioner>>,
+    password_backends: Option<Vec<password::Provisioner>>,
+}
+
+impl Provision {
+    pub fn new(
+        hostname: impl Into<String>,
+        username: impl Into<String>,
+        ssh_keys: impl Into<Vec<PublicKeys>>,
+    ) -> Self {
```
dongsupark, line 39: The code looks okay as long as we have only 3 parameters for `new()`.

Just a thought. I am wondering how much scalable it could be, for example when we would have 10 - 20 backends.
Then callers need to pass 10 - 20 parameters to `new()`, which could be tricky to follow and maintain.
However, as far as I understand, Rust does not have a native way of dealing variable args, right?
So we might need to split those into separate structs in the future.

jeremycline: Yeah, I think `new()` is just for required arguments and any change there would be an API breakage. I felt like these three would always be required, but we could also accept no arguments at all to `new()` and if you don't want to create a user for whatever reason, you don't have to. At that point, though, I don't know why you'd use the API.

libazureinit/src/provision/user.rs
jeremycline: Okay, I added a `User` type in commit 3b4b94b583b1d4e92b96a6a5d0d0c7b1345b859a and threaded it through the API. It bundles all the user-related configuration together and allows you to optionally provide a list of supplementary groups. How does that look to you?

libazureinit/src/provision/mod.rs
jeremycline: When adding the group I added a dedicated type for user configuration so it reduces the number of parameters to two.

libazureinit/src/provision/user.rs
```diff
@@ -5,7 +5,57 @@ use std::process::Command;
...
+
+    /// Set a password for the user; this is optional.
+    pub fn with_password(mut self, password: impl Into<String>) -> Self {
+        self.password = Some(password.into());
+        self
+    }
+
+    /// A list of supplemental group names to add the user to.
+    ///
+    /// If any of the groups do not exist on the host, provisioning will fail.
```
jeremycline, line 53: We could, instead, probe for groups that exist and only add the user to those groups. I don't love that because it feels spooky and unexpected ("I asked it to add the user to these groups and it didn't, but also didn't fail"), but on the other hand you wouldn't have to hand-tune the list for each distribution.

jeremycline:
Okay, I cleaned up the Rust docs and ensured all the publicly exported types have documentation, I think this is now ready.

dongsupark (approved):
Thanks

PR #92, opened 2024-06-26
libazureinit: remove requirement of minimum Rust version

dongsupark:
Replace `inspect_err` with `map_err`, to remove requirement of the minimum Rust version from 1.76, which is apparently not wide-spread enough in other development environments.

See also https://github.com/Azure/azure-init/pull/84#pullrequestreview-2086317824.

/cc @SeanDougherty


jeremycline (approved):
I guess we'll need to eventually settle on MSRV and add it to CI. I filed an issue about that.

Issue #93, opened 2024-06-26
[RFE] Select a MSRV


jeremycline:
## Current situation

Prior to #84 we didn't declare a Minimum Supported Rust Version. Based on #92 1.76 is too new. We should select a MSRV deemed acceptably old and add it to CI to ensure we work with it.

## Impact

While it's fairly easy to get the latest versions of Rust via rustup, distributions often lag behind current Rust releases by a lot. It's possible to build with a new toolchain and use the result on older distros, but if we want to build with the distribution toolchains we'll need to be more conservative about using new stdlib features.

## Implementation options

Debian 12 looks to ship 1.63. I believe the latest RHEL 9 minor version includes 1.75.

If the goal is to build with distro-shipped toolchains we'll need to go back to at least 1.63 which shipped in August of 2022.

If we're okay not building with distro toolchains we could pick whatever the project currently builds with and call it a day. That way we are at least explicit about what we support.

dongsupark:
Let's try to run CI with both toolchains, "stable" and "a fairly old version that is still wide-spread". Question is how do we figure out the latter in an automated way, without having to depend on Rust versions of specific distros.

I am thinking about using specific syntax supported by [dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain?tab=readme-ov-file#toolchain-expressions), like:
```
      matrix:
        rust: [stable, stable 18 months ago]
...
  - uses: dtolnay/rust-toolchain@master
    with:
      toolchain: ${{matrix.rust}}
```
As of now "stable 18 months ago" means around [1.67](https://github.com/rust-lang/rust/releases/tag/1.67.0) from Jan 2023, that looks already old enough I suppose.

To do that, we would need to switch from the current `actions-rs/toolchain` to `dtolnay/rust-toolchain` action. That is reasonable because `actions-rs/toolchain` almost unmaintained.

dongsupark:
> As of now "stable 18 months ago" means around [1.67](https://github.com/rust-lang/rust/releases/tag/1.67.0) from Jan 2023, that looks already old enough I suppose.

Actually the current source cannot be built with Rust 1.66.1, 18 months ago, because:
```
error: package `nix v0.29.0` cannot be built because it requires rustc 1.69 or newer, while the currently active rustc version is 1.66.1
Either upgrade to rustc 1.69 or newer, or use
cargo update -p nix@0.29.0 --precise ver
where `ver` is the latest version of `nix` supporting rustc 1.66.1
```

[...]

PR #94, opened 2024-07-02
.github: migrate to dtolnay/rust-toolchain to build with old toolchains

dongsupark:
Migrate to GitHub Actions [dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain), to support multiple toolchains as well as an old toolchain like stable 12 months ago.

Note: it is not possible to go back to e.g. 18 months ago, Rust [1.66.1](https://github.com/rust-lang/rust/releases/tag/1.66.1) as of now, because then nix v0.29.0 of the current code base cannot be built, nix 0.29.0 requires rustc 1.69 or newer.

Now that CI needs to build source code with an old version 12 months ago, Rust 1.70, cargo clippy starts to print out warnings like below.

```rust
error: redundant clone
  --> libazureinit/src/distro.rs:21:35
   |
21 |                     .arg(home_path.clone())
   |                                   ^^^^^^^^ help: remove this
   |
 note: this value is dropped without further use
  --> libazureinit/src/distro.rs:21:26
   |
21 |                     .arg(home_path.clone())
   |                          ^^^^^^^^^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone
   = note: `-D clippy::redundant-clone` implied by `-D warnings`
error: redundant clone
```

Fix redundant clone issue of clippy.

Fixes https://github.com/Azure/azure-init/issues/93.

## Testing done

Tested on my fork.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,35 +11,24 @@ jobs:
   build-test:
     name: Build and test azure-init
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust-toolchain: [stable, stable 12 months ago]
     steps:
       - name: Install libudev
         run: |
           sudo apt update
           sudo apt install -y libudev-dev
       - uses: actions/checkout@v4
-      - uses: actions-rs/toolchain@v1
+      - uses: dtolnay/rust-toolchain@master
         with:
-          profile: minimal
-          toolchain: stable
-          override: true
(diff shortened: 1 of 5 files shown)


jeremycline (approved):
I'm happy with the "one year ago" approach. We may have to re-visit this if some dependency bumps its MSRV aggressively and has security updates, but we can cross that bridge if we come to it.

PR #95, opened 2024-07-02
user: simplify ssh key provisioning and drop unsafe usage

jeremycline:
In order to correctly set the ownership of the authorized_keys file, the function was using libc directly with some unsafe blocks. It's tricky to correctly use the C APIs and in this case, both getpwnam() and getgrnam() return NULL if a matching entry isn't found or if an error occurred. Since we weren't checking the return value this would result in a null pointer dereference.

`nix` provides safe APIs to retrieve the user and group ids so rather than implementing it ourselves, just use those. This also refactors the two separate APIs for creating the directory and writing the keys to a single call that handles it all.

This is extracted from #91 as it's not really related to the API design.

### DIFF
--- a/libazureinit/src/user.rs
+++ b/libazureinit/src/user.rs
@@ -1,75 +1,51 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-use std::fs;
-use std::fs::create_dir;
-use std::fs::File;
-use std::io::Write;
+use std::{
+    fs::Permissions,
+    io::Write,
+    os::unix::fs::{DirBuilderExt, PermissionsExt},
+};
 
-use nix::unistd::{Gid, Uid};
-use std::ffi::CString;
-use std::os::unix::fs::PermissionsExt;
+use tracing::instrument;
 
 use crate::error::Error;
 use crate::imds::PublicKeys;
 
-pub async fn set_ssh_keys(
+pub fn set_ssh_keys(
     keys: Vec<PublicKeys>,
-    username: String,
-    file_path: String,
+    username: impl AsRef<str>,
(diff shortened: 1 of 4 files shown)


anhvoms (approved):
LGTM. Thank you!

libazureinit/src/user.rs
```diff
@@ -1,95 +1,177 @@
...
-    new_permissions.set_mode(0o600);
-    fs::set_permissions(authorized_keys_path.clone(), new_permissions)?;
-
-    let uid_username = CString::new(username.clone())?;
-    let uid_passwd = unsafe { libc::getpwnam(uid_username.as_ptr()) };
-    let uid = unsafe { (*uid_passwd).pw_uid };
-    let new_uid = Uid::from_raw(uid);
-
-    let gid_groupname = CString::new(username)?;
-    let gid_group = unsafe { libc::getgrnam(gid_groupname.as_ptr()) };
```
dongsupark, line 40: Removing the libc-related code like that, it becomes now unnecessary to keep `libc` any more.
Can you please remove [libc](https://github.com/Azure/azure-init/blob/ffe6d547847e56341137d45a01a20456783f2353/libazureinit/Cargo.toml#L22) in `libazureinit/Cargo.toml`?

To catch such issues, we should actually regularly run `cargo machete` or so. And there are other unused crates in this repo, but I will soon create another PR to remove those.

dongsupark (commented):
Looks good in general.

libazureinit/src/user.rs
jeremycline: Good catch, I've dropped it

dongsupark (approved):

Issue #99, opened 2024-07-03
[BUG] media::get_environment attempts to locate a non-existent CDROM device on AzureLinux


SeanDougherty:
## Description

When testing #92 on Azure Linux 3.0, azure-init encounters an issue attempting to retrieve block devices using `media::get_environment()`. This causes azure-init to report an error and exit pre-maturely. 

## Impact

This is part of work to get azure-init supported on Azure Linux 3.0 (as well as many other distros) 

## Environment and steps to reproduce

1. **Set-up**: 

Boot an Azure Linux 3.0 VM, if you need an image, please reach out.

3. **Action(s)**:
    
`tdnf install libudev-devel git -y`
`git clone https://github.com/SeanDougherty/azure-init.git`
`cd azure-init`
`git checkout azl`
`cargo build --all`
`./target/debug/azure-init`

4. **Error**:
`Unable to get list of block devices`
## Expected behavior
Azure-init can get a mountable device for its use. 

## Additional information

I explored more in my branch, and I can see that there are devices available, they just might not be CDROM devices. (See photo)
<img width="450" alt="image" src="https://github.com/Azure/azure-init/assets/22079170/4402cfd7-6fd3-452e-a37a-d90c70758e65">

For reference, [this JSON](https://github.com/microsoft/azurelinux/blob/coal/azure-init-PR-86/toolkit/imageconfigs/coal-marketplace-gen2.json) is the configuration file used by AzL build tools to compose the image. You can see the devices enumerated under `Disks:`.

dongsupark:
Thanks for the bug report.

Azure provides initial provisioning data, `ovf-env.xml`, as either via UDF/iso9660 device or an IMDS endpoint. The issue is that azure-init currently only supports the local device as source, not IMDS.

We should first check for a local mounted device, then fall back to fetching from IMDS.
For example, see [cloud-init](https://github.com/canonical/cloud-init/blob/2b6fe6403db769de14f7c7b7e4aa65f5bea8f3e0/cloudinit/sources/DataSourceAzure.py#L621), or [ignition](https://github.com/coreos/ignition/blob/7548ac6dafbb93fe89baf8853e97b71123cc5aed/internal/providers/azure/azure.go#L84). I think cloud-init code there is too complicated, so would rather recommend the ignition approach.

anhvoms:
> Thanks for the bug report.
> 
[...]

PR #100, opened 2024-07-04
Get username either by IMDS or from a local OVF file

dongsupark:
Username can be obtained either via fetching instance metadata from IMDS or mounting a local device for OVF environment file. It should not fail immediately in a single failure, instead it should fall back to the other mechanism. So it is not a good idea to use `?` for `query()` or `get_environment()`.

Fixes https://github.com/Azure/azure-init/issues/99

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -34,25 +34,28 @@ fn get_environment() -> Result<Environment, anyhow::Error> {
 }
 
 fn get_username(
-    instance_metadata: &InstanceMetadata,
-    environment: &Environment,
+    instance_metadata: Option<&InstanceMetadata>,
+    environment: Option<&Environment>,
 ) -> Result<String, anyhow::Error> {
-    if instance_metadata
-        .compute
-        .os_profile
-        .disable_password_authentication
-    {
-        // password authentication is disabled
-        Ok(instance_metadata.compute.os_profile.admin_username.clone())
-    } else {
-        // password authentication is enabled
-
-        Ok(environment
-            .clone()
-            .provisioning_section
-            .linux_prov_conf_set
-            .username)
+    if let Some(metadata) = instance_metadata {
+        if metadata.compute.os_profile.disable_password_authentication {
+            // If password authentication is disabled,
+            // simply read from IMDS metadata if available.
+            return Ok(metadata.compute.os_profile.admin_username.clone());
+        }
+        // If password authentication is enabled,
+        // fall back to reading from OVF environment file.
     }
+
+    // Read username from OVF environment via mounted local device.
+    environment
+        .map(|env| {
+            env.clone()
+                .provisioning_section
+                .linux_prov_conf_set
+                .username
+        })
+        .ok_or(LibError::UsernameFailure.into())
 }
 
 #[tokio::main]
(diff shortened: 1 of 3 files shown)


dongsupark:
Pushed commits to fix minor issues in the e2e tests script.

@SeanDougherty Could you please retest with this PR, to see if the issue of https://github.com/Azure/azure-init/issues/99 was gone?

I myself tried to run e2e tests with an Azure Linux gallery image, `MicrosoftCBLMariner:azure-linux:azure-linux-gen2:3.20240510.01`, and could not see the booting issue described. Though I am not sure if the current e2e test coverage would be enough for making the booting issue visible.
(After booting I actually hit useradd issues as already known, which would be addressed in separate PRs.)

SeanDougherty:
> Pushed commits to fix minor issues in the e2e tests script.
> 
> @SeanDougherty Could you please retest with this PR, to see if the issue of #99 was gone?
> 
> I myself tried to run e2e tests with an Azure Linux gallery image, `MicrosoftCBLMariner:azure-linux:azure-linux-gen2:3.20240510.01`, and could not see the booting issue described. Though I am not sure if the current e2e test coverage would be enough for making the booting issue visible. (After booting I actually hit useradd issues as already known, which would be addressed in separate PRs.)

After retesting, I can confirm this has resolved the issue. Thank you Dongsu!

And, yes, as you had mentioned, there are other issues with useradd that are being addressed in [this PR](https://github.com/Azure/azure-init/pull/91#discussion_r1667124031)

LGTM

src/main.rs
```diff
@@ -86,8 +89,23 @@ async fn provision() -> Result<(), anyhow::Error> {
...
-    let username = get_username(&instance_metadata, &get_environment()?)?;
+    // Username can be obtained either via fetching instance metadata from IMDS
+    // or mounting a local device for OVF environment file. It should not fail
+    // immediately in a single failure, instead it should fall back to the other
+    // mechanism. So it is not a good idea to use `?` for query() or
+    // get_environment().
+    let instance_metadata = match imds::query(&client).await {
+        Ok(m) => Some(m),
+        Err(_) => None,
+    };
```
jeremycline, line 100: There's a [convenience function](https://doc.rust-lang.org/std/result/enum.Result.html#method.ok) on `Result` to convert to `Option` (and [one to go back the other way](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or)):
```suggestion
    let instance_metadata = match imds::query(&client).await.ok();
```

src/main.rs
```diff
@@ -86,8 +89,23 @@ async fn provision() -> Result<(), anyhow::Error> {
...
+
+    let environment = match get_environment() {
+        Ok(env) => Some(env),
+        Err(_) => None,
+    };
```
jeremycline, line 105:
```suggestion
    let environment = get_environment().ok();
```

jeremycline (approved):
A few superficial notes, but nothing blocking. At some point we should have the library provide a nice interface for this, but we can tackle that later.

src/main.rs
dongsupark: Good catch, fixed.

jeremycline (approved):

Issue #69, opened 2024-03-20
[RFE] passwordless sudo


anhvoms:
I believe currently the admin user is configured to be part of sudoers group, but not passwordless sudo. If we want to maintain consistency with current behaviors from cloud-init and walinuxagent, this should be set. We can also provide this as configurable option (#65)

balakreddy:
Assigning this issue to myself. I will be working on this.
Thanks

PR #101, opened 2024-07-10
Enable passwordless sudo for the user

balakreddy:
Currently admin user is part of the sudoers
group, adding the user for passwordless sudo
from this issue #69

### DIFF
--- a/libazureinit/src/provision/user.rs
+++ b/libazureinit/src/provision/user.rs
@@ -1,7 +1,9 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-use std::process::Command;
+use std::{fs::Permissions, os::unix::fs::PermissionsExt, process::Command};
+
+use std::io::Write;
 
 use tracing::instrument;
 
@@ -88,11 +90,13 @@ pub enum Provisioner {
 
 impl Provisioner {
     pub(crate) fn create(&self, user: &User) -> Result<(), Error> {
-        match self {
+        let _ = match self {
             Self::Useradd => useradd(user),
             #[cfg(test)]
             Self::FakeUseradd => Ok(()),
-        }
+        };
+        let path = "/etc/sudoers.d/azure-init-user";
+        add_user_for_passwordless_sudo(user.name.clone(), path)
     }
 }
 
@@ -123,9 +127,38 @@ fn useradd(user: &User) -> Result<(), Error> {
     Ok(())
 }
 
+fn add_user_for_passwordless_sudo(
+    username: String,
+    path: &str,
+) -> Result<(), Error> {
+    // Create a file under /etc/sudoers.d with azure-init-user
+    let sudoers_path = path;
+    let mut sudoers_file = std::fs::OpenOptions::new()
+        .append(true)
+        .create(true)
+        .open(sudoers_path)?;
+    write!(
+        sudoers_file,
+        "{} ALL=(ALL) NOPASSWD: ALL \n",
+        username.to_string()
+    )?;
+    sudoers_file.flush()?;
+    // Set the permission
+    sudoers_file.set_permissions(Permissions::from_mode(0o600))?;
+    Ok(())
+}
+
 #[cfg(test)]
 mod tests {
+    use std::{
+        fs::{self, Permissions},
+        os::unix::fs::PermissionsExt,
+    };
+
     use crate::User;
+
+    use super::add_user_for_passwordless_sudo;
+    const PATH: &str = "/tmp/test1";
 
     #[test]
     fn password_skipped_in_debug() {
@@ -140,6 +173,19 @@ mod tests {
         assert_eq!(
             "User { name: \"azureuser\", groups: [\"wheel\"], ssh_keys: [], password: false }",
             format!("{:?}", user_without_password)
+        );
+    }
+
+    #[test]
+    fn test_user_insecure() {
+        let user_insecure = User::new("azureuser", []);
+        let a = add_user_for_passwordless_sudo(user_insecure.name, PATH);
+        assert!(a.is_ok());
+        assert!(fs::metadata(PATH).is_ok(), "Specified file not created");
+        assert_eq!(
+            fs::Permissions(PATH).mode(),
+            Permissions.mode(0o600),
+            "Permissions are not set properly"
         );
     }
 }
--- a/libazureinit/src/provision/user.rs
+++ b/libazureinit/src/provision/user.rs
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-use std::{fs::Permissions, os::unix::fs::PermissionsExt, process::Command};
+use std::{os::unix::fs::OpenOptionsExt, process::Command};
 
 use std::io::Write;
 
@@ -90,13 +90,15 @@ pub enum Provisioner {
 
 impl Provisioner {
     pub(crate) fn create(&self, user: &User) -> Result<(), Error> {
-        let _ = match self {
-            Self::Useradd => useradd(user),
+        match self {
+            Self::Useradd => {
+                useradd(user)?;
+                let path = "/etc/sudoers.d/azure-init-user";
+                add_user_for_passwordless_sudo(user.name.as_str(), path)
+            }
             #[cfg(test)]
             Self::FakeUseradd => Ok(()),
-        };
-        let path = "/etc/sudoers.d/azure-init-user";
-        add_user_for_passwordless_sudo(user.name.clone(), path)
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


balakreddy:
@jeremycline can you review this pull request.
Thanks

libazureinit/src/user.rs
```diff
@@ -50,6 +50,18 @@ pub(crate) fn provision_ssh(
     Ok(())
 }
 
+pub fn add_user_for_passwordless_sudo(
+    username: &str,
+) -> Result<(), Error>{
+    let mut sudoers_file = std::fs::OpenOptions::new()
+        .append(true)
+        .create(true)
+        .open("/etc/sudoers")?;
```
jeremycline, line 59: The distributions I'm familiar with (I checked Gentoo, Debian, Ubuntu, Fedora, and RHEL) configure sudo to read additional configuration snippets from `/etc/sudoers.d`.

This will include any file in the `/etc/sudoers.d/` directory as configuration which makes it easy to drop in additional configuration. This makes it easy to drop in snippets of config without parsing the main config to make sure it's not already there, and also when the distribution configuration is updated it doesn't potentially zap your config change (depending on how administrators handle new config files from package updates).

I think we should certainly _prefer_ to write to a `/etc/sudoers.d/<username>` file and given the wide distro support for it maybe only support that approach. Debian documents that it'll ignore files ending in `~` or containing a `.` character so we should at least strip out any `.` characters from the file name.

Also, when we create the file with this approach you'll want to [set the permissions on it](https://doc.rust-lang.org/std/fs/struct.File.html#method.set_permissions) - you can probably just do something like the ssh provisioning: https://github.com/Azure/azure-init/blob/bc83b93e5f9d9638334012479b9876ccfa3fe4e6/libazureinit/src/provision/ssh.rs#L32

jeremycline (changes requested):
I was going to suggest we get this in before #91 but it already got merged so I'm afraid you're going to have to git-rebase and move things a bit. Let me know if you have any questions about that!

jeremycline:
Also, you'll want to run `cargo fmt` to auto-format your code, and also `cargo clippy` which provides tips on writing more idiomatic code, common woopsies, etc.

libazureinit/src/user.rs
balakreddy: Jeremy, I created a azure-init-user file under /etc/sudoers.d and added info for passwordless sudo for the user.

libazureinit/src/user.rs
```diff
@@ -50,6 +50,29 @@ pub(crate) fn provision_ssh(
     Ok(())
 }
 
+pub fn add_user_for_passwordless_sudo(username: &str) -> Result<(), Error> {
+    // Create a file under /etc/sudoers.d with azure-init-user
+    let sudoers_path = "/etc/sudoers.d/azure-init-user";
+    let _ = File::create(sudoers_path)?;
```
jeremycline, line 56: You can delete this line since the next line will also create the file.

libazureinit/src/user.rs
```diff
@@ -50,6 +50,29 @@ pub(crate) fn provision_ssh(
...
+        "{} ALL=(ALL) NOPASSWD: ALL \n",
+        username.to_string()
+    )?;
+    sudoers_file.flush()?;
+    // Set the permission
+    let metadata = fs::metadata(sudoers_path)?;
+    let permissions = metadata.permissions();
+    let mut new_permissions = permissions.clone();
+    new_permissions.set_mode(0o700);
+    fs::set_permissions(sudoers_path, new_permissions)?;
```
jeremycline, line 72: You can create a `Permissions` object from a mode constant, so you can do this instead:
```suggestion
    sudoers_file.set_permissions(Permissions::from_mode(0o600))?;
```
You'll also want to use `0o600` (owner read-write) instead of `0o700` (owner read-write-execute) since this isn't an executable.

libazureinit/src/user.rs
```diff
@@ -50,6 +50,29 @@ pub(crate) fn provision_ssh(
     Ok(())
 }
 
+pub fn add_user_for_passwordless_sudo(username: &str) -> Result<(), Error> {
+    // Create a file under /etc/sudoers.d with azure-init-user
+    let sudoers_path = "/etc/sudoers.d/azure-init-user";
```
jeremycline, line 55: We will want to write a unit test or two for this function, and as you've likely noticed we can't mock things out in the same way one can in, say, Python. Setting the file path here makes writing unit tests difficult since we probably want to write the test file out to some temporary directory.

However, we can use a trick to make this easier: rather than defining the path here, we can accept the path as a function parameter. In our unit tests we'll create a temporary directory and pass a path inside that to this function so we can make sure it creates the file we want with the permissions we expect.

This trick has the somewhat pompous (IMO) name of "dependency injection" which really just means "give the function the thing it needs rather than have it create it", but it's something you'll likely reach for a lot more in Rust than Python since you can't mock things out in the same way.

jeremycline (changes requested):
Let's add some unit tests as well - the tests should be very similar to the tests for the ssh authorized_keys file so I would look at those for inspiration.

This branch still conflicts with the main branch, so we'll also need to rebase it. I can walk you through the process in a call, if you'd like.

libazureinit/src/user.rs
balakreddy: Done

libazureinit/src/user.rs
balakreddy: Done, added the suggestion

libazureinit/src/user.rs
balakreddy: I have created a test case and modified the function as suggested by giving path as parameter, can you take a look at let me know any changes required.

balakreddy:
The changes are not pushed yet, will update here once the changes are pushed.

libazureinit/src/provision/user.rs
```diff
@@ -123,10 +127,39 @@ fn useradd(user: &User) -> Result<(), Error> {
     Ok(())
 }
 
+fn add_user_for_passwordless_sudo(
+    username: String,
```
jeremycline, line 131: Since we're going to use this as part of a format string, we don't need to own the value. Instead, we can take a read-only reference to the string and avoid the need to clone it in the caller. You can change `String` to `&str` here and then you should be able to replace the `user.name.clone()` call with `user.name.as_str()`

libazureinit/src/provision/user.rs
```diff
@@ -123,10 +127,39 @@ fn useradd(user: &User) -> Result<(), Error> {
...
+    let sudoers_path = path;
+    let mut sudoers_file = std::fs::OpenOptions::new()
+        .append(true)
+        .create(true)
+        .open(sudoers_path)?;
+    write!(
+        sudoers_file,
+        "{} ALL=(ALL) NOPASSWD: ALL \n",
+        username.to_string()
+    )?;
```
jeremycline, line 144: There's a `writeln!` that will handle inserting the newline for you. You also shouldn't need the `to_string()` call here.

libazureinit/src/provision/user.rs
```diff
@@ -123,10 +127,39 @@ fn useradd(user: &User) -> Result<(), Error> {
...
+    sudoers_file.flush()?;
+    // Set the permission
+    sudoers_file.set_permissions(Permissions::from_mode(0o600))?;
```
jeremycline, line 147: My concern with this approach is that there's a window between creating the file and setting the permissions that it is potentially world-writable. Additionally, the file has been opened to append, meaning we won't clobber any write someone slips in before the permissions get set appropriately.

Now, _presumably_ this runs early in boot on a trusted image, but we don't know how folks may use the library and violate that expectation. I wasn't aware of this or I would have suggested it earlier, but there's a [Unix extension in the standard library](https://doc.rust-lang.org/std/os/unix/fs/trait.OpenOptionsExt.html), so I recommend doing this instead:
```
use std::os::unix::fs::OpenOptionsExt;

std::fs::OpenOptions::new()
    .mode(0o600)
    .truncate(true)
    .create(true)
    .open(sudoers_path)?;
```

I think we would prefer truncating the file if it already exists rather than accepting any previous file since it's a drop-in file specific to this program. You could imagine someone building an image from one that got provisioned with this tool and failing to clean up this configuration. We wouldn't want to leave that original provisioning user with sudo permissions.

libazureinit/src/provision/user.rs
```diff
@@ -123,10 +127,39 @@ fn useradd(user: &User) -> Result<(), Error> {
...
 mod tests {
+    use std::{
+        fs::{self, Permissions},
+        os::unix::fs::PermissionsExt,
+    };
+
     use crate::User;
 
+    use super::add_user_for_passwordless_sudo;
+    const PATH: &str = "/tmp/test1";
```
jeremycline, line 161: The test should create a temporary directory with [TempDir](https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html) and then have it create a file within that. See the ssh tests for examples

jeremycline (changes requested):

libazureinit/src/provision/user.rs
balakreddy: Done

libazureinit/src/provision/user.rs
balakreddy: Done

libazureinit/src/provision/user.rs
```diff
@@ -88,11 +90,13 @@ pub enum Provisioner {
 
 impl Provisioner {
     pub(crate) fn create(&self, user: &User) -> Result<(), Error> {
-        match self {
+        let _ = match self {
```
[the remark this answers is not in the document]
balakreddy: Done

libazureinit/src/provision/user.rs
balakreddy: Done

libazureinit/src/provision/user.rs
```diff
@@ -92,7 +94,9 @@ impl Provisioner {
             Self::Useradd => useradd(user),
             #[cfg(test)]
             Self::FakeUseradd => Ok(()),
-        }
+        }?;
+        let path = "/etc/sudoers.d/azure-init-user";
+        add_user_for_passwordless_sudo(user.name.as_str(), path)
```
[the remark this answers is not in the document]
balakreddy: Done, made the changes and ran the tests. All passed.

libazureinit/src/provision/user.rs
```diff
@@ -142,4 +171,28 @@ mod tests {
...
+        let _user_insecure = User::new("azureuser", []);
+        let ret =
+            add_user_for_passwordless_sudo(&_user_insecure.name, path_str);
+
+        assert!(ret.is_ok());
+        assert!(
+            fs::metadata(path.clone()).is_ok(),
+            "{path_str} file not created"
+        );
+        let mode = fs::metadata(path_str).unwrap().permissions().mode();
```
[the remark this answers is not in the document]
balakreddy: Made the changes suggested.

libazureinit/src/provision/user.rs
```diff
@@ -142,4 +171,28 @@ mod tests {
             format!("{:?}", user_without_password)
         );
     }
+
+    #[test]
+    fn test_user_insecure() {
```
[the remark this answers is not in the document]
balakreddy: Done

jeremycline (approved):
Thanks for all your work here!

PR #103, opened 2024-07-12
Swapping get_mount_devices from while loop to filter_map

peytonr18:
The while loop for get_mount_devices was stuck in a loop where it was  re-fetching the list of mounted devices every time if anything returned true and then would repeatedly find the same device. 

This should eliminate the issue and allow it to only add the device once.

### DIFF
--- a/libazureinit/Cargo.toml
+++ b/libazureinit/Cargo.toml
@@ -24,7 +24,6 @@ fstab = "0.4.0"
 [dev-dependencies]
 tempfile = "3"
 whoami = "1"
-mockall = "0.11.0"
 
 [lib]
 name = "libazureinit"
--- a/libazureinit/Cargo.toml
+++ b/libazureinit/Cargo.toml
@@ -19,7 +19,7 @@ nix = {version = "0.29.0", features = ["fs", "user"]}
 block-utils = "0.11.1"
 tracing = "0.1.40"
 strum = { version = "0.26.3", features = ["derive"] }
-uuid = { version = "1.0", features = ["v4"] }
+fstab = "0.4.0"
 
 [dev-dependencies]
 tempfile = "3"
--- a/libazureinit/Cargo.toml
+++ b/libazureinit/Cargo.toml
@@ -19,10 +19,12 @@ nix = {version = "0.29.0", features = ["fs", "user"]}
 block-utils = "0.11.1"
 tracing = "0.1.40"
 strum = { version = "0.26.3", features = ["derive"] }
+uuid = { version = "1.0", features = ["v4"] }
 
 [dev-dependencies]
 tempfile = "3"
 whoami = "1"
+mockall = "0.11.0"
 
 [lib]
 name = "libazureinit"
--- a/libazureinit/src/media.rs
+++ b/libazureinit/src/media.rs
@@ -77,14 +77,17 @@ const CDROM_VALID_FS: &[&str] = &["iso9660", "udf"];
 
 // Get a mounted device with any filesystem for CDROM
 pub fn get_mount_device() -> Result<Vec<String>, Error> {
-    let mut list_devices: Vec<String> = Vec::new();
-
-    while let Some(device) = block_utils::get_mounted_devices()?
+    let devices = block_utils::get_mounted_devices()?;
+    let list_devices: Vec<String> = devices
         .into_iter()
-        .find(|dev| CDROM_VALID_FS.contains(&dev.fs_type.to_str()))
-    {
-        list_devices.push(device.name);
-    }
+        .filter_map(|dev| {
+            if CDROM_VALID_FS.contains(&dev.fs_type.to_str()) {
+                Some(dev.name)
+            } else {
+                None
+            }
+        })
+        .collect();
 
     Ok(list_devices)
 }
--- a/libazureinit/src/media.rs
+++ b/libazureinit/src/media.rs
@@ -421,10 +421,10 @@ mod tests {
                 Device {
                     id: Some(Uuid::new_v4()),
                     name: "device1".to_string(),
-                    media_type: MediaType::NVME, // Adjust this to match the MediaType variants
+                    media_type: MediaType::NVME, 
                     device_type: DeviceType::Disk,
                     capacity: 700_000_000,
-                    fs_type: FilesystemType::Ntfs, // Adjust this to match the FilesystemType variants
+                    fs_type: FilesystemType::Ntfs, 
                     serial_number: Some("12345".to_string()),
                     logical_block_size: Some(512),
                     physical_block_size: Some(512),
@@ -432,7 +432,7 @@ mod tests {
                 Device {
                     id: Some(Uuid::new_v4()),
                     name: "device2".to_string(),
-                    media_type: MediaType::Rotational, // Adjust this to match the MediaType variants
+                    media_type: MediaType::Rotational, 
                     device_type: DeviceType::Disk,
                     capacity: 1_000_000_000,
                     fs_type: FilesystemType::Ext4,
@@ -443,10 +443,10 @@ mod tests {
                 Device {
                     id: Some(Uuid::new_v4()),
                     name: "device3".to_string(),
-                    media_type: MediaType::NVME, // Adjust this to match the MediaType variants
+                    media_type: MediaType::NVME, 
                     device_type: DeviceType::Disk,
                     capacity: 700_000_000,
-                    fs_type: FilesystemType::Xfs, // Adjust this to match the FilesystemType variants
+                    fs_type: FilesystemType::Xfs, 
                     serial_number: Some("54321".to_string()),
                     logical_block_size: Some(512),
                     physical_block_size: Some(512),
--- a/libazureinit/src/media.rs
+++ b/libazureinit/src/media.rs
@@ -78,18 +78,13 @@ pub const PATH_MOUNT_POINT: &str = "/run/azure-init/media/";
 const CDROM_VALID_FS: &[&str] = &["iso9660", "udf"];
 const MTAB_PATH: &str = "/etc/mtab";
 
-// Public wrapper function
-pub fn get_mount_device() -> Result<Vec<String>, Error> {
-    wrapped_get_mount_device(Path::new(MTAB_PATH))
-}
-
-// Inner function that takes a file path as an argument
-fn wrapped_get_mount_device(path: &Path) -> Result<Vec<String>, Error> {
-    // Create a new FsTab instance and parse the fstab file
-    let fstab = FsTab::new(path);
+// Get a mounted device with any filesystem for CDROM
+pub fn get_mount_device(path: Option<&Path>) -> Result<Vec<String>, Error> {
+    // Create a new FsTab instance and parse the mtab file
+    let fstab = FsTab::new(path.unwrap_or(Path::new(MTAB_PATH)));
 
     // Get the entries from the FsTab
-    let entries = fstab.get_entries()?;
+    let entries: Vec<FsEntry> = fstab.get_entries()?;
 
     // Filter and collect device names
     let list_devices: Vec<String> = entries
@@ -231,8 +226,9 @@ mod tests {
     use super::*;
     use crate::error::Error;
     use fstab::FsEntry;
-    use std::io;
+    use std::io::{self, Write};
     use std::path::PathBuf;
+    use tempfile::NamedTempFile;
 
     #[test]
     fn test_get_ovf_env_none_missing() {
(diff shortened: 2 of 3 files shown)


anhvoms (changes requested):
Can we add a unit test to ensure the function will return a list of devices with the appropriate fs_type?

libazureinit/src/media.rs
```diff
@@ -77,14 +77,17 @@ const CDROM_VALID_FS: &[&str] = &["iso9660", "udf"];
...
+    let devices = block_utils::get_mounted_devices()?;
+    let list_devices: Vec<String> = devices
         .into_iter()
-        .find(|dev| CDROM_VALID_FS.contains(&dev.fs_type.to_str()))
-    {
-        list_devices.push(device.name);
-    }
+        .filter_map(|dev| {
+            if CDROM_VALID_FS.contains(&dev.fs_type.to_str()) {
+                Some(dev.name)
```
jeremycline, line 85: It turns out that `dev.name` is not the full path to the device, but rather just [the file name](https://docs.rs/block-utils/0.11.1/src/block_utils/lib.rs.html#183). Given how we're using this library, we might be better off using `fstab` crate directly and do something similar to [what block-utils does](https://docs.rs/block-utils/0.11.1/src/block_utils/lib.rs.html#451) except we would use the [FsEntry](https://docs.rs/fstab/latest/fstab/struct.FsEntry.html) directly.

peytonr18: In the end, this makes things so much easier! Will adjust tests accordingly, thank you Jeremy!!

libazureinit/src/media.rs
```diff
@@ -76,19 +78,30 @@ pub const PATH_MOUNT_POINT: &str = "/run/azure-init/media/";
...
+                None
+            }
+        })
+        .collect();
 
     Ok(list_devices)
 }
 
+// Wrapper function
+pub fn get_wrapped_mount_devices() -> Result<Vec<String>, Error> {
```
jeremycline, line 101: Based on some side-channel conversations this is probably going away so this comment is really just a tip for the future.

One of the nice things about `pub` is that you can hide your sins from the outside world. What you could do is define this as `pub fn get_mount_device() ...` and then the function you're wrapping as `fn wrapped_get_mount_device(...)`. That way no one outside the library is even aware there's a wrapper layer.

libazureinit/src/media.rs
```diff
@@ -396,4 +417,66 @@ mod tests {
             _ => panic!("Non-empty passwords aren't allowed"),
         };
     }
+
+    #[test]
+    fn test_wrapped_get_mount_device() {
+        // Mock function to simulate reading /etc/fstab
+        let mock_path = Path::new("/mock/mtab");
+        let mock_get_mounted_devices = || -> Result<Vec<FsEntry>, io::Error> {
```
jeremycline, line 425: Instead of using mock, you could test the function by writing a valid mount table using `tempfile`. You'll want the [NamedTempFile](https://docs.rs/tempfile/latest/tempfile/struct.NamedTempFile.html) version since we pass in a `Path` to the function and so we need a name for it.

Once you make it, you can write the table to the file and pass the file path to the function to test it. Here's a (trimmed) mount table with a `udf` iso as an example, you can replace udf with iso9660 or remove it entirely to test the various function behaviors, of course:
```
sysfs /sys sysfs rw,nosuid,nodev,noexec,relatime 0 0
/dev/sda /var/tmp btrfs rw,relatime,discard=async,space_cache=v2,subvolid=5,subvol=/ 0 0
/dev/sr0 /mnt udf ro,relatime,iocharset=utf8 0 0
```

Once you write the table to the tempfile, you can call `into_temp_path()` which will close the file and give you a `tempfile::TempPath` object which will remove the file when it's dropped. It implements `AsRef<Path>` so you can call `as_ref()` on it to produce a `&Path` to give to your wrapped function.

libazureinit/src/media.rs
```diff
@@ -74,17 +76,32 @@ pub const PATH_MOUNT_DEVICE: &str = "/dev/sr0";
 pub const PATH_MOUNT_POINT: &str = "/run/azure-init/media/";
 
 const CDROM_VALID_FS: &[&str] = &["iso9660", "udf"];
+const MTAB_PATH: &str = "/etc/mtab";
 
-// Get a mounted device with any filesystem for CDROM
+// Public wrapper function
```
jeremycline, line 81: I think we should leave the comment as-is. Although it's not currently a Rust documentation comment (three forward slashes, `///`) it is a public function and so it should have a description of what it does (and ideally how to use it).

In case you've not played with it, you can run `cargo doc` on the project which will produce a set of HTML docs for all public types. See https://doc.rust-lang.org/rustdoc/what-is-rustdoc.html for all the details, but we should aim to have all things marked `pub` to include rustdoc comments.

jeremycline (changes requested):

libazureinit/src/media.rs
dongsupark: Another option would be to simply define only `pub fn get_mount_device(path: Option<&Path>) -> Result<Vec<String>, Error>`, without a wrapper layer. Then caller could either pass a custom path with `Some` or simply `None` to use the default.
IMHO code here fortunately does not need another wrapper layer like that.

```Rust
pub fn get_mount_device(path: Option<&Path>) -> Result<Vec<String>, Error> {
    // Create a new FsTab instance and parse the fstab file
    let fstab = FsTab::new(path.unwrap_or(Path::new(MTAB_PATH)));
```

libazureinit/src/media.rs
peytonr18: I am about to merge a change for this -- let me know if that matches what you were thinking. I tried to add all the scenarios I could think we might need to test, but that may have been overkill. Thank you for the review!

libazureinit/src/media.rs
peytonr18: Should have resolved this, thank you!

libazureinit/src/media.rs
```diff
@@ -75,18 +77,28 @@ pub const PATH_MOUNT_DEVICE: &str = "/dev/sr0";
...
 // Get a mounted device with any filesystem for CDROM
 #[instrument]
-pub fn get_mount_device() -> Result<Vec<String>, Error> {
-    let mut list_devices: Vec<String> = Vec::new();
+pub fn get_mount_device(path: Option<&Path>) -> Result<Vec<String>, Error> {
+    // Create a new FsTab instance and parse the mtab file
+    let fstab = FsTab::new(path.unwrap_or(Path::new(MTAB_PATH)));
 
-    while let Some(device) = block_utils::get_mounted_devices()?
+    // Get the entries from the FsTab
```
anhvoms, line 88: I would drop these comments (create new fstab, get the entries...) as they don't really add any meaningful value. Anyone reading the code can already tell that the method will "get the entries from the FsTab"

libazureinit/src/media.rs
```diff
@@ -75,18 +77,28 @@ pub const PATH_MOUNT_DEVICE: &str = "/dev/sr0";
...
+    let entries: Vec<FsEntry> = fstab.get_entries()?;
+
+    // Filter and collect device names
```
anhvoms, line 91:
```suggestion
    // Retrieve the names of all devices that have cdrom-type filesystem (e.g., udf)
```

libazureinit/src/media.rs
```diff
@@ -75,18 +77,28 @@ pub const PATH_MOUNT_DEVICE: &str = "/dev/sr0";
...
+    let list_devices: Vec<String> = entries
```
anhvoms, line 92:
```suggestion
    let cdrom_devices: Vec<String> = entries
```

libazureinit/src/media.rs
```diff
@@ -75,18 +77,28 @@ pub const PATH_MOUNT_DEVICE: &str = "/dev/sr0";
...
+        .filter_map(|entry| {
+            if CDROM_VALID_FS.contains(&entry.vfs_type.as_str()) {
+                Some(entry.fs_spec.clone())
+            } else {
+                None
+            }
+        })
+        .collect();
 
     Ok(list_devices)
```
anhvoms, line 103:
```suggestion
    Ok(cdrom_devices)
```

anhvoms (changes requested):
A few nitpicking on style, but overall looking good!

libazureinit/Cargo.toml
```diff
@@ -19,10 +19,12 @@ nix = {version = "0.29.0", features = ["fs", "user"]}
 block-utils = "0.11.1"
 tracing = "0.1.40"
 strum = { version = "0.26.3", features = ["derive"] }
+fstab = "0.4.0"
 
 [dev-dependencies]
 tempfile = "3"
 whoami = "1"
+mockall = "0.11.0"
```
dongsupark, line 27: Now that it does not use `mockall`, can you please remove this line?

Interesting, `cargo machete` does not detect the case.

libazureinit/src/media.rs
```diff
@@ -222,6 +231,11 @@ pub fn mount_parse_ovf_env(dev: String) -> Result<Environment, Error> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::error::Error;
+    use fstab::FsEntry;
```
dongsupark, line 235: Not used any more. Can you please remove this line?

```
warning: unused import: `fstab::FsEntry`
   --> libazureinit/src/media.rs:235:9
    |
235 |     use fstab::FsEntry;
    |         ^^^^^^^^^^^^^^
    |
    = note: `#[warn(unused_imports)]` on by default
```

libazureinit/src/media.rs
```diff
@@ -222,6 +231,11 @@ pub fn mount_parse_ovf_env(dev: String) -> Result<Environment, Error> {
...
+    use std::io::{self, Write};
```
dongsupark, line 236: Not used any more. Can you please remove `self`?
```
warning: unused import: `self`
   --> libazureinit/src/media.rs:236:19
    |
236 |     use std::io::{self, Write};
    |                   ^^^^

```

libazureinit/src/media.rs
```diff
@@ -222,6 +231,11 @@ pub fn mount_parse_ovf_env(dev: String) -> Result<Environment, Error> {
...
+    use std::path::PathBuf;
```
dongsupark, line 237: Not used any more. Can you please remove this line?
```
warning: unused import: `std::path::PathBuf`
   --> libazureinit/src/media.rs:237:9
    |
237 |     use std::path::PathBuf;
```

anhvoms (approved):
LGTM. Thanks!

libazureinit/src/media.rs
```diff
@@ -75,20 +77,27 @@ pub const PATH_MOUNT_DEVICE: &str = "/dev/sr0";
...
         .into_iter()
-        .find(|dev| CDROM_VALID_FS.contains(&dev.fs_type.to_str()))
-    {
-        list_devices.push(device.name);
-    }
-
-    Ok(list_devices)
+        .filter_map(|entry| {
+            if CDROM_VALID_FS.contains(&entry.vfs_type.as_str()) {
+                Some(entry.fs_spec.clone())
```
jeremycline, line 93: You don't actually need the `clone()` here. The reason is that the list of entries is being consumed with `into_iter()` and giving us owned values. Since we don't hold onto the entry and reference it elsewhere, we are free to take ownership of the `fs_spec` it contains rather than create a new value from it.

Again, this is a micro-optimization that will never matter in practice, but just an FYI 😄

libazureinit/src/media.rs
```diff
@@ -406,4 +418,84 @@ mod tests {
...
+            /dev/sr0 /mnt/cdrom iso9660 ro,user,noauto 0 0
+            /dev/sr1 /mnt/cdrom2 udf ro,user,noauto 0 0
+        "#;
+        temp_file
+            .write_all(mount_table.as_bytes())
+            .expect("Failed to write to temporary file");
+        let temp_path = temp_file.into_temp_path();
+        let result = get_mount_device(Some(temp_path.as_ref()));
+
+        assert!(result.is_ok());
```
jeremycline, line 436: The test fails if it panics so you could leave this out since the next line unwraps the result.

libazureinit/src/media.rs
```diff
@@ -75,20 +77,27 @@ pub const PATH_MOUNT_DEVICE: &str = "/dev/sr0";
...
 const CDROM_VALID_FS: &[&str] = &["iso9660", "udf"];
+const MTAB_PATH: &str = "/etc/mtab";
 
 // Get a mounted device with any filesystem for CDROM
 #[instrument]
-pub fn get_mount_device() -> Result<Vec<String>, Error> {
-    let mut list_devices: Vec<String> = Vec::new();
+pub fn get_mount_device(path: Option<&Path>) -> Result<Vec<String>, Error> {
+    let fstab = FsTab::new(path.unwrap_or(Path::new(MTAB_PATH)));
+    let entries: Vec<FsEntry> = fstab.get_entries()?;
```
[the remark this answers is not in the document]
peytonr18: I think I originally did it because the compiler was saying we weren't using FsEntry (when we were) so that was my intention but can easily remove!

libazureinit/src/media.rs
peytonr18: Removed this statement from all the tests. Thank you!

dongsupark (approved):

dongsupark:
Adjusted the commit message when squashing and merging.

PR #105, opened 2024-07-16
Add an initial mkosi config

jeremycline:
This is still a work-in-progress.

It still needs:
 - A stubbed out Wireserver and IMDS instance

To try it out as-is, you can build and then boot the image with:

$ mkosi -d fedora
$ sudo mkosi boot

The version of azure-init in target/debug/ will be run as part of the
image startup. You can inspect the result with "systemctl status
azure-init". Run "systemctl poweroff" to leave the environment.

This needs to be done in an Azure VM since it relies on talking to a
real Wireserver and IMDS.

Refer to mkosi docs[0] for installation options

[0] https://github.com/systemd/mkosi?tab=readme-ov-file#installation


mkosi.conf
```diff
@@ -0,0 +1,43 @@
...
+# This only works with a bootable image.
+#QemuArgs=-cdrom target/image.iso
+
+# Boot to a temporary snapshot of the image - changes won't be persisted to the
+# image.
+Ephemeral=true
+# Perform an incremental rebuild of the image
+Incremental=true
+# Make the source available at runtime
+RuntimeBuildSources=yes
```
dongsupark, line 43: How about adding `ToolsTree=default` here?
On my Debian stable system with systemd 252, `mkosi boot` does not run because systemd-nspawn version is too low.

See also [NEWS](https://github.com/systemd/mkosi/blob/main/NEWS.md#v20):
`ToolsTree=default can be used to make sure a recent version of systemd is used to do the image build`.

cjp256:
closing as stale

Issue #63, opened 2024-03-19
[RFE] azure-init should add retries around IMDS and Wireserver operations


anhvoms:
## Current situation

There's no retry when REST API calls to IMDS or wireserver (goal_state, report_health)

## Impact

Without retry, if there's a transient issue from Azure platform, provisioning will fail

## Additional information

When to retry and how many times/how long to retry is a complex topic, especially when IMDS/Wireserver does not provide any guidance. This is the current behavior from cloud-init ([ref](https://github.com/canonical/cloud-init/blob/cacf452bd08a22430b303a6d805b2d57167b31df/cloudinit/sources/azure/imds.py#L20), [ref](https://github.com/canonical/cloud-init/blob/cacf452bd08a22430b303a6d805b2d57167b31df/cloudinit/sources/helpers/azure.py#L225)), which we can use as a reference (or perhaps we can provide this as a config that can be configured within the image? e.g., /etc/azure-init/azure-init.conf)

Total retrying time for IMDS should total no more than 5 minutes, for Wireserver 20 minutes.
Retry around Connection timeout/Read timeout: timeout for rest call should be set at 30s
Retry around non-200 http error codes (410, 404, 503, 400, 500, 429): timeout should be set at 2s, with backoff of 1s

jeremycline:
One thing we might want to consider is replacing the hand-written IMDS client with https://crates.io/crates/azure_svc_imds from https://github.com/Azure/azure-sdk-for-rust/ which, based on the documentation, already implements retries.

anhvoms:
> One thing we might want to consider is replacing the hand-written IMDS client with https://crates.io/crates/azure_svc_imds from https://github.com/Azure/azure-sdk-for-rust/ which, based on the documentation, already implements retries.

Oh, this is nice. I wasn't aware of this crate. We should check it out and see if the retry policy is easy to customize to our need.

anhvoms:
This one should not be closed yet as we have only added retries to IMDS, and not wireserver. Communication to wireserver, similarly, need to be retried upon failure (goalstate.rs)

PR #107, opened 2024-07-18
Use retry logic for HTTP communication with IMDS

dongsupark:
Use retry logic and max duration for HTTP communication with IMDS, creating a generic wrapper retry_loop.
Doing that, it is possible for azure-init to retry to send HTTP requests, when requests failed for some reason.

For now we define a fixed const for timeout, max retries, for simplicity.

Make imds::query() capable of receiving input URL, so that both main binary and unit tests can send requests to different endpoints.

Fixes https://github.com/Azure/azure-init/issues/63

## Testing done

TBD

### DIFF
--- a/libazureinit/Cargo.toml
+++ b/libazureinit/Cargo.toml
@@ -23,6 +23,8 @@ fstab = "0.4.0"
 
 [dev-dependencies]
 tempfile = "3"
+# test-util feature is needed for time::advance in unit tests
+tokio = { version = "1", features = ["full", "test-util"] }
 whoami = "1"
 
 [lib]
--- a/libazureinit/src/error.rs
+++ b/libazureinit/src/error.rs
@@ -47,4 +47,6 @@ pub enum Error {
         "Failed to set the user password; none of the provided backends succeeded"
     )]
     NoPasswordProvisioner,
+    #[error("A timeout error occurred")]
+    Timeout(#[from] tokio::time::error::Elapsed),
 }
--- a/libazureinit/src/imds.rs
+++ b/libazureinit/src/imds.rs
@@ -1,14 +1,17 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-use reqwest;
 use reqwest::header::HeaderMap;
 use reqwest::header::HeaderValue;
-use reqwest::Client;
+use reqwest::{Client, StatusCode};
+
+use std::time::Duration;
 
 use serde::{Deserialize, Deserializer};
 use serde_json;
 use serde_json::Value;
+
+use tokio::time::timeout;
 
 use crate::error::Error;
 
@@ -87,65 +90,110 @@ where
     }
 }
 
-pub async fn query(client: &Client) -> Result<InstanceMetadata, Error> {
-    let url = "http://169.254.169.254/metadata/instance?api-version=2021-02-01";
+// Set of StatusCodes that should be retried,
+// e.g. 400, 404, 410, 429, 500, 503.
+const RETRY_CODES: &[StatusCode] = &[
+    StatusCode::BAD_REQUEST,
+    StatusCode::NOT_FOUND,
+    StatusCode::GONE,
+    StatusCode::TOO_MANY_REQUESTS,
+    StatusCode::INTERNAL_SERVER_ERROR,
+    StatusCode::SERVICE_UNAVAILABLE,
+];
+
+static DEFAULT_IMDS_URL: &str =
+    "http://169.254.169.254/metadata/instance?api-version=2021-02-01";
+
+pub async fn query(
+    client: &Client,
+    retry_interval: Duration,
+    total_timeout: Duration,
+    input_url: Option<&str>,
+) -> Result<InstanceMetadata, Error> {
     let mut headers = HeaderMap::new();
+
+    let url = match input_url {
+        Some(url) => url,
+        None => DEFAULT_IMDS_URL,
+    };
 
     headers.insert("Metadata", HeaderValue::from_static("true"));
 
-    let request = client.get(url).headers(headers);
-    let response = request.send().await?;
+    let response = timeout(total_timeout, async {
+        loop {
+            if let Ok(response) = client
+                .get(url)
+                .timeout(Duration::from_secs(30))
+                .send()
+                .await
+            {
+                let statuscode = response.status();
 
-    if response.status().is_success() {
-        let imds_body = response.text().await?;
-        let metadata: InstanceMetadata = serde_json::from_str(&imds_body)?;
+                if statuscode.is_success() && statuscode == StatusCode::OK {
+                    return response.error_for_status();
+                }
 
-        Ok(metadata)
-    } else {
-        Err(Error::HttpStatus {
-            endpoint: url.to_owned(),
-            status: response.status(),
-        })
-    }
+                if !RETRY_CODES.contains(&statuscode) {
+                    return response.error_for_status();
+                }
+            }
+
+            tokio::time::sleep(retry_interval).await;
+        }
+    })
+    .await?;
+
+    let imds_body = response?.text().await?;
+
+    let metadata: InstanceMetadata = serde_json::from_str(&imds_body)?;
+
+    Ok(metadata)
 }
 
 #[cfg(test)]
 mod tests {
     use serde_json::json;
 
-    use super::{InstanceMetadata, OsProfile};
+    use super::{query, InstanceMetadata, OsProfile, RETRY_CODES};
+
+    use reqwest::{header, Client, StatusCode};
+    use std::time::Duration;
+    use tokio::io::AsyncWriteExt;
+    use tokio::net::TcpListener;
+    use tokio::time;
+
+    static BODY_CONTENTS: &str = r#"
+{
+  "compute": {
+    "azEnvironment": "cloud_env",
+    "customData": "",
+    "evictionPolicy": "",
+    "isHostCompatibilityLayerVm": "false",
+    "licenseType": "",
+    "location": "eastus",
+    "name": "AzTux-MinProvAgent-Test-0001",
+    "offer": "0001-com-ubuntu-server-focal",
+    "osProfile": {
+      "adminUsername": "MinProvAgentUser",
+      "computerName": "AzTux-MinProvAgent-Test-0001",
+      "disablePasswordAuthentication": "true"
+    },
+    "publicKeys": [
+      {
+        "keyData": "ssh-rsa test_key1",
+        "path": "/path/to/.ssh/authorized_keys"
+      },
+      {
+        "keyData": "ssh-rsa test_key2",
+        "path": "/path/to/.ssh/authorized_keys"
+      }
+    ]
+  }
+}"#;
 
     #[test]
     fn instance_metadata_deserialization() {
-        let file_body = r#"
-        {
-            "compute": {
-              "azEnvironment": "cloud_env",
-              "customData": "",
-              "evictionPolicy": "",
-              "isHostCompatibilityLayerVm": "false",
-              "licenseType": "",
-              "location": "eastus",
-              "name": "AzTux-MinProvAgent-Test-0001",
-              "offer": "0001-com-ubuntu-server-focal",
-              "osProfile": {
-                "adminUsername": "MinProvAgentUser",
-                "computerName": "AzTux-MinProvAgent-Test-0001",
-                "disablePasswordAuthentication": "true"
-              },
-              "publicKeys": [
-                {
-                  "keyData": "ssh-rsa test_key1",
-                  "path": "/path/to/.ssh/authorized_keys"
-                },
-                {
-                  "keyData": "ssh-rsa test_key2",
-                  "path": "/path/to/.ssh/authorized_keys"
-                }
-              ]
-            }
-        }"#
-        .to_string();
+        let file_body = BODY_CONTENTS.to_string();
 
         let metadata: InstanceMetadata =
             serde_json::from_str(&file_body).unwrap();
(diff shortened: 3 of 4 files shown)


anhvoms:
Do we know how much this increases the final build size of azure-init?

src/main.rs
```diff
@@ -102,16 +115,28 @@ async fn provision() -> Result<(), anyhow::Error> {
     )?;
     default_headers.insert(header::USER_AGENT, user_agent);
     let client = Client::builder()
-        .timeout(std::time::Duration::from_secs(30))
+        .timeout(std::time::Duration::from_secs(HTTP_TIMEOUT_SEC))
         .default_headers(default_headers)
         .build()?;
 
+    let imds_endpoint = azure_core::Url::parse(IMDS_ENDPOINT)?;
+    let imds_client = ImdsClient::builder(Arc::new(AzureCliCredential::new()))
```
anhvoms, line 123: does this require authentication to work? provisioning shouldn't require obtaining credential

src/main.rs
```diff
@@ -102,16 +115,28 @@ async fn provision() -> Result<(), anyhow::Error> {
...
+        .endpoint(imds_endpoint.clone())
+        .retry(RetryOptions::exponential(ExponentialRetryOptions {
```
anhvoms, line 125: for communicating with IMDS during provisioning it's not recommended to use Exponential retry as it can severely affect provisioning latency.

src/main.rs
```diff
@@ -2,8 +2,15 @@
...
 
 use std::process::ExitCode;
+use std::sync::Arc;
+use std::time::Duration;
 
 use anyhow::Context;
+
+use azure_core::{ExponentialRetryOptions, RetryOptions};
+use azure_identity::AzureCliCredential;
+use azure_svc_imds::Client as ImdsClient;
```
anhvoms, line 12: I looked at the azure sdk source quickly but did not see how to configure which http error codes to retry for. We need to ensure the following errors are retried: 403, 404, 410, 429, 500, 503.  While there are contention on whether the use of these error codes are appropriate on Azure, the fact remains that Azure platform will emit these errors and retrying is required.

anhvoms (changes requested):
I'm not sure how much we gain from using azure sdk vs just implementing our own retry policy here. Can you verify the final code size?

dongsupark:
> I'm not sure how much we gain from using azure sdk vs just implementing our own retry policy here. Can you verify the final code size?

1. debug:

  * without Azure SDK

```
69084 -rwxrwxr-x 2 dpark dpark 70734488 Jul 19 14:17 target/debug/azure-init*
```

  * with Azure SDK

```
114888 -rwxrwxr-x 2 dpark dpark 117639624 Jul 19 13:48 target/debug/azure-init*
```

+66%


2. release:

  * without Azure SDK

```
6904 -rwxrwxr-x 2 dpark dpark 7069624 Jul 19 14:19 target/release/azure-init*
```

  * with Azure SDK

```
10356 -rwxrwxr-x 2 dpark dpark 10603568 Jul 19 14:15 target/release/azure-init*
```

+49%

Tried also turning off default features of Azure SDK crates, but not much difference.

src/main.rs
dongsupark: Done, thanks.

dongsupark:
> I'm on the fence about the SDK - the lack of error codes to retry seems like it might be a deal-breaker. Since we want a fixed retry policy without jitter and so on, the implementation could be a loop with a constant sleep in it, wrapped in https://docs.rs/tokio/latest/tokio/time/fn.timeout.html.

Ok, let me rewrite this PR based on https://docs.rs/tokio/latest/tokio/time/fn.timeout.html or any simpler ones.

libazureinit/src/util.rs
```diff
@@ -0,0 +1,19 @@
...
+    max_duration: Duration,
+) -> T {
+    pin_mut!(delayer);
+
+    if let Ok(result) = timeout(max_duration, &mut delayer).await {
+        result
+    } else {
+        sleep(delay_sec).await;
+        delayer.await
+    }
```
jeremycline, line 18: I think what this does is await the given future once, and if it times out sleeps for `delay_sec` before awaiting the future again without a timeout.

src/main.rs
```diff
@@ -135,7 +138,13 @@ async fn provision() -> Result<(), anyhow::Error> {
     // immediately in a single failure, instead it should fall back to the other
     // mechanism. So it is not a good idea to use `?` for query() or
     // get_environment().
-    let instance_metadata = imds::query(&client).await.ok();
+    let instance_metadata = util::retry_loop(
```
jeremycline, line 141: While having a `util` module isn't unprecedented, it does mean library users have to know about it and that they want to use it in conjunction with `imds::query`. Given the concrete retry policy in #63 I wonder if we even want to expose any tune-ables to users.

The API could either stay `imds::query(&client)` and we handle the retries and timeout internally or, if we want to expose options to the user, change it to be `imds::query(&query, retry_interval, total_timeout)`.

I think what it could look like internal to `imds::query` would be something like this (haven't compile-tested, not formatted for tidiness):
```Rust
// Issue #63 says 5 minutes for IMDS
let response = tokio::time::timeout(Duration::from_secs(5 * 60), async {
    let interval = Duration::from_secs(2);

    loop {
        if let Ok(response) = client.get(url).timeout(Duration::from_secs(30).send().await {
            // Do some response inspection here for what to retry, but basic version is below
            if response.status() == 200 {
                break response;
            }
        }
        // not sure if backoff of 1s described in #63 means we should do `interval += Duration::from_secs(1)`, flat 2s
        // sleep here
        tokio::time::sleep(interval.clone()).await;
    }
    
}).await?;
```

Basically, write a future that loops forever with the retry, and then enforce the overall timeout with tokio. That way we don't need to think about intervals we sleep, time we spent on individual request timeouts, etc.

jeremycline (changes requested):

src/main.rs
dongsupark: Done, thanks.

libazureinit/src/imds.rs
```diff
@@ -87,26 +90,65 @@ where
...
+                if statuscode.is_success() && statuscode == 200 {
+                    break response;
+                }
 
-    if response.status().is_success() {
-        let imds_body = response.text().await?;
-        let metadata: InstanceMetadata = serde_json::from_str(&imds_body)?;
+                if !is_retryable_code(statuscode) {
+                    fastfail = true;
+                    break response;
```
jeremycline, line 133: You could return this by doing `break (response, fastfail);` instead of needing to pre-declare a variable with `mut`. 

You could also return `response.error_for_status()` if it's not on the retry list and then unwrap that result with `?` outside the timeout block to get rid of the `if fastfail` entirely. `reqwest::Error` has a [url](https://docs.rs/reqwest/latest/reqwest/struct.Error.html#method.url) attached to it so we should be able to extract all the information we need from it.

It's all equivalent and what here now isn't difficult to understand, so it's totally up to you.

src/main.rs
```diff
@@ -20,6 +21,8 @@ use tracing_subscriber::layer::SubscriberExt;
 use tracing_subscriber::EnvFilter;
 
 const VERSION: &str = env!("CARGO_PKG_VERSION");
+const HTTP_TIMEOUT_SEC: u64 = 5 * 60;
```
jeremycline, line 24: Maybe `IMDS_HTTP_TIMEOUT_SEC` since we'll add a different one for WireServer?

jeremycline (commented):
A few very minor notes, but it looks good to me. You should also be able to write a unit test for the timeout case, I think, with https://docs.rs/tokio/latest/tokio/time/fn.advance.html although I've not personally done this before so I'm just guessing based on the documentation.

libazureinit/src/error.rs
```diff
@@ -47,4 +47,6 @@ pub enum Error {
         "Failed to set the user password; none of the provided backends succeeded"
     )]
     NoPasswordProvisioner,
+    #[error("An elapsed error occurred")]
+    FromElapsedError(#[from] tokio::time::error::Elapsed),
```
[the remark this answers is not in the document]
dongsupark: Renamed to `Timeout`.

libazureinit/src/imds.rs
```diff
@@ -87,26 +90,65 @@ where
...
+fn is_retryable_code(statuscode: StatusCode) -> bool {
+    // 400, 404, 410, 429, 500, 503
+    [
+        StatusCode::from_u16(400).unwrap(),
+        StatusCode::from_u16(404).unwrap(),
+        StatusCode::from_u16(410).unwrap(),
+        StatusCode::from_u16(429).unwrap(),
+        StatusCode::from_u16(500).unwrap(),
+        StatusCode::from_u16(503).unwrap(),
+    ]
```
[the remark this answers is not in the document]
dongsupark: Removed the separate function, and changed to `RETRY_CODES.contains()`.

libazureinit/src/imds.rs
dongsupark: Changed to `response.error_for_status()`.

src/main.rs
dongsupark: Added prefix `IMDS_`.

dongsupark:
> A few very minor notes, but it looks good to me. You should also be able to write a unit test for the timeout case, I think, with https://docs.rs/tokio/latest/tokio/time/fn.advance.html although I've not personally done this before so I'm just guessing based on the documentation.

Added unit tests using `tokio::time::advance` etc.

libazureinit/src/imds.rs
```diff
@@ -206,4 +254,74 @@ mod tests {
...
+        const IMDS_HTTP_TOTAL_TIMEOUT_SEC: u64 = 5 * 60;
+        const IMDS_HTTP_PERCLIENT_TIMEOUT_SEC: u64 = 30;
+        const IMDS_HTTP_RETRY_INTERVAL_SEC: u64 = 2;
+
+        let mut default_headers = header::HeaderMap::new();
+        let user_agent =
+            header::HeaderValue::from_str("azure-init test").unwrap();
+
+        // Run a local test server that replies with simple test data.
+        let serverlistener = TcpListener::bind("127.0.0.1:3000").await.unwrap();
```
jeremycline, line 269:
```suggestion
        let serverlistener = TcpListener::bind("127.0.0.1:0").await.unwrap();
```
will cause the OS to assign it a free port, and you've already written the rest of the test to use the result of `local_addr()` so it should Just Work.

libazureinit/Cargo.toml
```diff
@@ -12,7 +12,8 @@ description = "A common library for provisioning Linux VMs on Azure."
 reqwest = { version = "0.12.0", default-features = false, features = ["blocking", "json"] }
 serde = {version = "1.0.163", features = ["derive"]}
 thiserror = "1.0.58"
-tokio = { version = "1", features = ["full"] }
+# test-util feature is needed for time::advance in unit tests
+tokio = { version = "1", features = ["full", "test-util"] }
```
jeremycline, line 16: You should be able to put this line in `[dev-dependencies]` and leave it as-is here so it's only enabled for tests.

jeremycline (approved):
Looks great, thanks! I had two little optional notes, but this is good to go from my perspective.

libazureinit/src/imds.rs
```diff
@@ -87,65 +90,110 @@ where
...
+];
+
+static DEFAULT_IMDS_URL: &str =
+    "http://169.254.169.254/metadata/instance?api-version=2021-02-01";
+
+pub async fn query(
+    client: &Client,
+    retry_interval: Duration,
+    total_timeout: Duration,
+    inputurl: Option<&str>,
```
vmarcella, line 111: NIT: we should rename this to `input_url` for consistency

libazureinit/src/imds.rs
```diff
@@ -87,65 +90,110 @@ where
...
+                .send()
+                .await
+            {
+                let statuscode = response.status();
 
-    if response.status().is_success() {
-        let imds_body = response.text().await?;
-        let metadata: InstanceMetadata = serde_json::from_str(&imds_body)?;
+                if statuscode.is_success() && statuscode == StatusCode::OK {
+                    return response.error_for_status();
```
vmarcella, line 133: What would `error_for_status()` return here given that the status code is a success?

libazureinit/src/imds.rs
```diff
@@ -87,65 +90,110 @@ where
...
+    let url = match inputurl {
+        Some(url) => url,
+        None => DEFAULT_IMDS_URL,
+    };
+
     headers.insert("Metadata", HeaderValue::from_static("true"));
 
-    let request = client.get(url).headers(headers);
-    let response = request.send().await?;
+    let response = timeout(total_timeout, async {
```
vmarcella, line 122: @anhvoms Should the retry interval exponentially increase (By factors of 2) as failures to query imds repeatedly happen? It is my understanding that this is what cloud-init does today

src/main.rs
```diff
@@ -20,6 +21,8 @@ use tracing_subscriber::layer::SubscriberExt;
 use tracing_subscriber::EnvFilter;
 
 const VERSION: &str = env!("CARGO_PKG_VERSION");
+const IMDS_HTTP_TIMEOUT_SEC: u64 = 5 * 60;
```
vmarcella, line 24: Because these are only used in a single place, do we need to store these as variables?

vmarcella (changes requested):
For the most part this LGTM, but I do want @anhvoms thoughts on exponentially increasing the retry interval when making repeated calls to IMDS before approving these changes.

libazureinit/src/imds.rs
anhvoms: No, we want fix retry interval to minimize impact to provisioning latency. cloud-init does fixed retry interval, not exponential. We have exponential backup for logging in the case of pre-provisioning VM (because some pre-provisioned VM can stay in pre-provisioning state for very long). Excessive logging isn't a concern for azure-init

anhvoms (approved):

vmarcella (approved):

libazureinit/src/imds.rs
dongsupark: In that case, it would just return the inner response of type `Response` itself, where the outer `response` would be `Ok`.

libazureinit/src/imds.rs
dongsupark: Done

libazureinit/src/imds.rs
dongsupark: Done

src/main.rs
dongsupark: Done

jeremycline (approved):

PR #118, opened 2024-08-09
Allowing existing users when provisioning

peytonr18:
Adding a check for existing users that allows provisioning continue if the user already exists in the system or adds it if the user does not. Max retries of adding user 5 times with 30 second delay between.

### DIFF
--- a/libazureinit/src/provision/user.rs
+++ b/libazureinit/src/provision/user.rs
@@ -186,15 +186,15 @@ fn useradd(user: &User) -> Result<(), Error> {
     tracing::debug!("Running command: {}", useradd_command);
 
     let status = Command::new(path_useradd)
-                    .arg(&user.name)
-                    .arg("--comment")
-                    .arg("azure-init created this user based on username provided in IMDS")
-                    .arg("--groups")
-                    .arg(user.groups.join(","))
-                    .arg("-d")
-                    .arg(home_path)
-                    .arg("-m")
-                    .status()?;
+        .arg(&user.name)
+        .arg("--comment")
+        .arg("azure-init created this user based on username provided in IMDS")
+        .arg("--groups")
+        .arg(user.groups.join(","))
+        .arg("-d")
+        .arg(home_path)
+        .arg("-m")
+        .status()?;
 
     tracing::debug!("useradd command exit status: {}", status);
 
--- a/libazureinit/src/provision/user.rs
+++ b/libazureinit/src/provision/user.rs
@@ -3,6 +3,10 @@
 
 use std::process::Command;
 
+use std::time::Duration;
+
+use std::thread::sleep;
+
 use tracing::instrument;
 
 use crate::{error::Error, imds::PublicKeys};
@@ -97,27 +101,49 @@ impl Provisioner {
 }
 
 #[instrument(skip_all)]
-fn useradd(user: &User) -> Result<(), Error> {
-    let path_useradd = env!("PATH_USERADD");
-    let home_path = format!("/home/{}", user.name);
+fn user_exists(username: &str) -> Result<bool, Error> {
+    let status = Command::new("getent")
+        .arg("passwd")
+        .arg(username)
+        .status()?;
 
-    let status = Command::new(path_useradd)
-                    .arg(&user.name)
-                    .arg("--comment")
-                    .arg(
-                      "Provisioning agent created this user based on username provided in IMDS",
-                    )
-                    .arg("--groups")
-                    .arg(user.groups.join(","))
-                    .arg("-d")
-                    .arg(home_path)
-                    .arg("-m")
-                    .status()?;
-    if !status.success() {
-        return Err(Error::SubprocessFailed {
-            command: path_useradd.to_string(),
-            status,
-        });
+    Ok(status.success())
+}
+
+#[instrument(skip_all)]
+fn useradd(user: &User) -> Result<(), Error> {
+    const MAX_RETRIES: u8 = 5;
+
+    if user_exists(&user.name)? {
+        return Ok(());
+    }
+
+    for attempt in 1..=MAX_RETRIES {
+        let path_useradd = env!("PATH_USERADD");
+        let home_path = format!("/home/{}", user.name);
+
+        let status = Command::new(path_useradd)
+                        .arg(&user.name)
+                        .arg("--comment")
+                        .arg(
+                          "Provisioning agent created this user based on username provided in IMDS",
+                        )
+                        .arg("--groups")
+                        .arg(user.groups.join(","))
+                        .arg("-d")
+                        .arg(home_path)
+                        .arg("-m")
+                        .status()?;
+        if status.success() {
+            return Ok(());
+        } else if attempt < MAX_RETRIES {
+            sleep(Duration::from_secs(30));
+        } else {
+            return Err(Error::SubprocessFailed {
+                command: path_useradd.to_string(),
+                status,
+            });
+        }
     }
 
     Ok(())
--- a/libazureinit/src/provision/user.rs
+++ b/libazureinit/src/provision/user.rs
@@ -9,13 +9,13 @@ use crate::{error::Error, imds::PublicKeys};
 
 /// The user and its related configuration to create on the host.
 ///
-/// A bare minimum user includes a name and a set of SSH public keys to allow the user to
-/// log into the host. Additional configuration includes a set of supplementary groups to
-/// add the user to, and a password to set for the user.
+/// A bare minimum user includes a name and a set of SSH public keys to allow
+/// the user to log into the host. Additional configuration includes a set of
+/// supplementary groups to add the user to, and a password to set for the user.
 ///
 /// By default, the user is included in the `wheel` group which is often used to
-/// grant administrator privileges via the `sudo` command. This can be changed with the
-/// [`User::with_groups`] method.
+/// grant administrator privileges via the `sudo` command. This can be changed
+/// with the [`User::with_groups`] method.
 ///
 /// # Example
 ///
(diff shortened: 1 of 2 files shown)


libazureinit/src/provision/user.rs
```diff
@@ -96,28 +100,50 @@ impl Provisioner {
...
-                    .status()?;
-    if !status.success() {
-        return Err(Error::SubprocessFailed {
-            command: path_useradd.to_string(),
-            status,
-        });
+    const MAX_RETRIES: u8 = 5;
+
+    if user_exists(&user.name)? {
+        return Ok(());
```
anhvoms, line 118: Add a log message indicating that the user exists, so user creation is skipped.
I do see a problem here though. If the user exists, but doesn't have the same groups that were indicated by groups specified by the user, what do we do with it? The safe behavior would be to add the user to the newly specified groups (without removing them from any existing group)

@jeremycline  for opinion on this

libazureinit/src/provision/user.rs
```diff
@@ -96,28 +100,50 @@ impl Provisioner {
...
+    }
+
+    for attempt in 1..=MAX_RETRIES {
```
anhvoms, line 121: I'm not sure if it's worth retrying. I am not able to come up with a scenario where useradd would succeed after trying again while provisioning. If useradd fails, we should log an error and return an error at the end

anhvoms (changes requested):

libazureinit/src/provision/user.rs
```diff
@@ -96,28 +100,50 @@ impl Provisioner {
...
+                        .arg("--groups")
+                        .arg(user.groups.join(","))
+                        .arg("-d")
+                        .arg(home_path)
+                        .arg("-m")
+                        .status()?;
+        if status.success() {
+            return Ok(());
+        } else if attempt < MAX_RETRIES {
+            sleep(Duration::from_secs(30));
```
jeremycline, line 140: Based on the discussion above this is going to go away, but I think it's worth noting that using `std::thread::sleep` could have some unexpected side effects since we might be executing in an asynchronous context here. 

In this case, the `azure-init` binary calls this (indirectly, via `libazure::Provision::provision`) from a thread in the tokio runtime. If we use `std::thread::sleep` we'll put that thread to sleep and tokio will not be able to complete work on that thread while it's asleep. If the user has other asynchronous tasks running, they will see everything freeze for 30 seconds before continuing. `tokio` ships its own [sleep](https://docs.rs/tokio/latest/tokio/time/fn.sleep.html) which will do what you want but allow tokio to keep working on other tasks. To use that, though, we'd need to mark the function as asynchronous.

In general (not just in Rust), we want to be careful about sleeping in library functions. That's not to say one should never sleep in a library: having retries for fallible things is nice for users, but you'll want to note in the API documentation that it will sleep between retries.

libazureinit/src/provision/user.rs
jeremycline: Yeah, this is one reason I opted to just fail in this case. If the user exists, was that intentional or an oversight during the image creation?

We default to adding the user to the superuser group so in the common case leaving the user in its existing groups won't really grant it additional, unexpected permissions. On the other hand, this is in the library, and someone might try to use this library in a way that doesn't grant the user superuser privileges and merging groups could leave the provisioned user with unexpected permissions if the user was supposed to be there and was in some set of groups that granted it expanded permissions.

One option would be to allow the user to exist, but fail if the set of requested groups doesn't exactly match the set of groups its already in. However, even in that narrow case, there are still questions, like are there existing `authorized_keys`? Do we leave those or not?

peytonr18: At least for answering the user exists question, the issue here is that in our testing mechanism for the image, we use azureuser for provisioning, image creation, etc so by the time we get here, the user has technically existed before. 
This lead tuxops (repo we are testing in) to fail. 

image

Moreover, I will just throw out there that when I deleted the user at the end of the image creation, azureinit did not provision a new user in a way that tuxops could login to a VM created from that image. 

With that, the goal was to just make it so we can allow existing users and avoid further trouble. I think allowing the user to exist and fail only if the group doesn't match should fix the tuxops problem but I'd have to test that to confirm. If it's not possible to do this and we are worried about someone abusing the library, I can go back to trying to figure out how to make it work with deleting and creating a new user directly as part of the image creation script.

anhvoms: The deprovisioning process by walinuxagent allows keeping or deleting the last provisioned user in the image. Some customer use this to keep certain user configuration (password, for example) without having to pass them through Azure during provisioning. 
As for the authorized_keys, the existing behavior of walinuxagent and cloud-init it to append the file with new keys. 

We can introduce options such as fail_on_existing_user, fail_on_mismatch_sudoer_groups, and append_authorized_ssh_key_file to allow stricter behaviors for azure-init

libazureinit/src/provision/user.rs
peytonr18: Super good to know -- forgot about the asynchronous context so very very good to know. Thank you for this!! I will file this information for the future.

libazureinit/src/provision/user.rs
peytonr18: With this all being said, the changes I am about to push have the functionality of useradd as follows: 

- User Does Not Exist: user is created with the specified groups.
- User Exists and is in the Same Groups: No changes are made; the useradd completes with a log message (but does not fail like it did before).
- User Exists but is Missing Some Groups: The user is added to the missing groups, and the function completes successfully.

 If this is what we all want and are good with, I can open an issue so we can add flagging to allow stricter behaviors and do that next.

peytonr18:
With the new PR, I was assuming we would want to add some new tests to this program. I think the only possible thing to consider here is how we want to do that -- I wrote something up like this but we may need to mock this out for the sake of not actually touching the groups that exist on the system? May be completely unnecessary as well but thought I would offer the suggestion in case. 

#[cfg(test)]
mod tests {

    use crate::User;
    use super::useradd;
    use std::collections::HashSet;

    #[test]
    fn user_addition() {
        std::process::Command::new("groupadd")
        .arg("testgroup")
        .status()
        .expect("Failed to create test group");

        let user = User::new("azureuser", vec![]).with_groups(vec!["testgroup".to_string()]);
        assert!(useradd(&user).is_ok(), "User should be added successfully or skipped");

        let output = std::process::Command::new("id")
            .arg("-nG")
            .arg("azureuser")
            .output()
            .expect("Failed to run id command");
        let output_string = String::from_utf8_lossy(&output.stdout);
        let groups: HashSet<_> = output_string
            .split_whitespace()
            .collect();
        assert!(groups.contains("testgroup"), "User should be in the testgroup group");

        std::process::Command::new("groupdel")
        .arg("testgroup")
        .status()
        .expect("Failed to delete test group");
    }

    #[test]
    fn user_group_addition() {
        let user_with_new_group = User::new("azureuser", vec![]).with_groups(vec!["adm".to_string()]);
        assert!(useradd(&user_with_new_group).is_ok(), "User should be added successfully or skipped");
        let updated_output: std::process::Output = std::process::Command::new("id")
            .arg("-nG")
            .arg("azureuser")
            .output()
            .expect("Failed to run id command");
        
        let updated_output_string = String::from_utf8_lossy(&updated_output.stdout);
        let updated_groups: HashSet<_> = updated_output_string
            .split_whitespace()
            .collect();
        assert!(updated_groups.contains("wheel"), "User should still be in the wheel group");
        assert!(updated_groups.contains("adm"), "User should now be in the adm group as well");
    }
}

libazureinit/src/provision/user.rs
```diff
@@ -96,26 +96,103 @@ impl Provisioner {
...
+
 #[instrument(skip_all)]
 fn useradd(user: &User) -> Result<(), Error> {
+    if user_exists(&user.name)? {
+        tracing::info!(
+            "User '{}' already exists. Skipping user creation.",
+            user.name
+        );
+
+        let current_groups_output =
```
anhvoms, line 117: This is not necessary. You can just apply usermod -aG. usermod will add the user to the specified groups without affecting existing membership (and won't give an error if the user is already part of any of the groups)

libazureinit/src/provision/user.rs
```diff
@@ -96,26 +96,103 @@ impl Provisioner {
...
+            tracing::info!(
+                "User '{}' is missing some groups. Adding them to the following groups: {}",
+                user.name,
+                group_list
+            );
+
+            let usermod_command =
+                format!("usermod -aG {} {}", group_list, user.name);
+
+            tracing::debug!("Running command: {}", usermod_command);
```
anhvoms, line 146: This is fine for this PR but long term speaking, we should create a wrapper for Command that will log all commands (we need to pay attention to sensitive data and redact them appropriately)

anhvoms (changes requested):

libazureinit/src/provision/user.rs
jeremycline: I'm less concerned if this is matching behavior from other tooling since it's less likely to surprise people. Particularly with walinuxagent allowing one to leave the user created, it would be an unpleasant experience to have this fail if the user exists.

My only request would be to add a note in the API documentation with the behavior. On the `User` struct is probably the most discoverable. In case you've not played with the API documentation before, you can build it locally with `cargo doc -p libazureinit --no-deps` which I find helpful to make sure it formats like I expect.

libazureinit/src/provision/user.rs
peytonr18: I think we already have an issue open for improved logging messages (if not, I'll open one), but I will move that up the priority list in terms of what is next.

libazureinit/src/provision/user.rs
```diff
@@ -96,26 +123,84 @@ impl Provisioner {
...
 }
 
+#[instrument(skip_all)]
+fn user_exists(username: &str) -> Result<bool, Error> {
+    let status = Command::new("getent")
+        .arg("passwd")
+        .arg(username)
+        .status()?;
+
+    Ok(status.success())
```
vmarcella, line 133: So `getent` will fail with a non zero status code if the user doesn't exist?

libazureinit/src/provision/user.rs
```diff
@@ -96,26 +123,84 @@ impl Provisioner {
...
+        }
+
+        return Ok(());
+    }
+
     let path_useradd = env!("PATH_USERADD");
     let home_path = format!("/home/{}", user.name);
 
+    let useradd_command = format!(
+        "{} {} --comment 'Provisioning agent created this user based on username provided in IMDS' --groups {} -d {} -m",
```
vmarcella, line 179: NIT: for the comment, let's make it state `azure-init` created this user based on username provided by IMDS`

libazureinit/src/provision/user.rs
```diff
@@ -96,26 +123,84 @@ impl Provisioner {
...
+
+    tracing::debug!("Running command: {}", useradd_command);
+
     let status = Command::new(path_useradd)
                     .arg(&user.name)
                     .arg("--comment")
-                    .arg(
-                      "Provisioning agent created this user based on username provided in IMDS",
-                    )
+                    .arg("Provisioning agent created this user based on username provided in IMDS")
```
vmarcella, line 191: Same as above comment wrt to the content of the comment attached to the user

vmarcella (approved):
I left a few minor comments but this looks good to me!

PR #119, opened 2024-08-22
Use retry logic for HTTP communication with wireserver

dongsupark:
Use similar retry logic for wireserver communication, like done for IMDS.
Add unit tests for `get_goalstate` and `report_health`.

Split `RETRY_CODES` and `HARDFAIL_CODES` into a separate module `http`, so that those can be used by other parts of libazureinit.

Add back missing headers into `client.get()` in libazureinit IMDS.

Fixes https://github.com/Azure/azure-init/issues/63.

## Testing done

Unit tests pass.

### DIFF
--- a/libazureinit/src/goalstate.rs
+++ b/libazureinit/src/goalstate.rs
@@ -133,9 +133,7 @@ fn build_report_health_file(goalstate: Goalstate) -> String {
 mod tests {
     use super::{build_report_health_file, Goalstate};
 
-    #[test]
-    fn test_parsing_goalstate() {
-        let goalstate_str = "<Goalstate>
+    static GOALSTATE_STR: &str = "<Goalstate>
             <Container>
                 <ContainerId>2</ContainerId>
                 <RoleInstanceList>
@@ -147,7 +145,25 @@ mod tests {
             <Version>example_version</Version>
             <Incarnation>test_goal_incarnation</Incarnation>
         </Goalstate>";
-        let goalstate: Goalstate = serde_xml_rs::from_str(goalstate_str)
+
+    static HEALTH_STR: &str = "<?xml version=\"1.0\" encoding=\"utf-8\"?>\n\
+        <Health xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" xmlns:xsd=\"http://www.w3.org/2001/XMLSchema\">\n\
+            <GoalStateIncarnation>test_goal_incarnation</GoalStateIncarnation>\n\
+            <Container>\n\
+                <ContainerId>2</ContainerId>\n\
+                <RoleInstanceList>\n\
+                    <Role>\n\
+                        <InstanceId>test_user_instance_id</InstanceId>\n\
+                        <Health>\n\
+                            <State>Ready</State>\n\
+                        </Health>\n\
+                    </Role>\n\
+                </RoleInstanceList>\n\
+            </Container>\n\
+        </Health>";
+    #[test]
+    fn test_parsing_goalstate() {
+        let goalstate: Goalstate = serde_xml_rs::from_str(GOALSTATE_STR)
             .expect("Failed to parse the goalstate XML.");
         assert_eq!(goalstate.container.container_id, "2".to_owned());
         assert_eq!(
@@ -164,40 +180,10 @@ mod tests {
 
     #[tokio::test]
     async fn test_build_report_health_file() {
-        let goalstate_str = "
-            <Goalstate>
-                <Container>
-                    <ContainerId>2</ContainerId>
-                    <RoleInstanceList>
-                        <RoleInstance>
-                            <InstanceId>test_user_instance_id</InstanceId>
-                        </RoleInstance>
-                    </RoleInstanceList>
-                </Container>
-                <Version>example_version</Version>
-                <Incarnation>test_goal_incarnation</Incarnation>
-            </Goalstate>";
-        let goalstate: Goalstate = serde_xml_rs::from_str(goalstate_str)
+        let goalstate: Goalstate = serde_xml_rs::from_str(GOALSTATE_STR)
             .expect("Failed to parse the goalstate XML.");
 
-        let expected_output =
-        "<?xml version=\"1.0\" encoding=\"utf-8\"?>\n\
-        <Health xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" xmlns:xsd=\"http://www.w3.org/2001/XMLSchema\">\n\
-            <GoalStateIncarnation>test_goal_incarnation</GoalStateIncarnation>\n\
-            <Container>\n\
-                <ContainerId>2</ContainerId>\n\
-                <RoleInstanceList>\n\
-                    <Role>\n\
-                        <InstanceId>test_user_instance_id</InstanceId>\n\
-                        <Health>\n\
-                            <State>Ready</State>\n\
-                        </Health>\n\
-                    </Role>\n\
-                </RoleInstanceList>\n\
-            </Container>\n\
-        </Health>";
-
         let actual_output = build_report_health_file(goalstate);
-        assert_eq!(actual_output, expected_output);
+        assert_eq!(actual_output, HEALTH_STR);
     }
 }
--- a/libazureinit/src/goalstate.rs
+++ b/libazureinit/src/goalstate.rs
@@ -1,15 +1,19 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
 
-use reqwest;
 use reqwest::header::HeaderMap;
 use reqwest::header::HeaderValue;
-use reqwest::Client;
+use reqwest::{Client, StatusCode};
+
+use std::time::Duration;
 
 use serde::Deserialize;
 use serde_xml_rs::from_str;
 
+use tokio::time::timeout;
+
 use crate::error::Error;
+use crate::http;
 
 #[derive(Debug, Deserialize, PartialEq)]
 pub struct Goalstate {
@@ -41,34 +45,80 @@ pub struct RoleInstance {
     instance_id: String,
 }
 
-pub async fn get_goalstate(client: &Client) -> Result<Goalstate, Error> {
-    let url = "http://168.63.129.16/machine/?comp=goalstate";
+const DEFAULT_GOALSTATE_URL: &str =
+    "http://168.63.129.16/machine/?comp=goalstate";
+
+pub async fn get_goalstate(
+    client: &Client,
+    retry_interval: Duration,
+    total_timeout: Duration,
+    url: Option<&str>,
+) -> Result<Goalstate, Error> {
+    let url = url.unwrap_or(DEFAULT_GOALSTATE_URL);
 
     let mut headers = HeaderMap::new();
     headers.insert("x-ms-agent-name", HeaderValue::from_static("azure-init"));
     headers.insert("x-ms-version", HeaderValue::from_static("2012-11-30"));
 
-    let request = client.get(url).headers(headers);
-    let response = request.send().await?;
+    let response = timeout(total_timeout, async {
+        let now = std::time::Instant::now();
+        loop {
+            if let Ok(response) = client
+                .get(url)
+                .headers(headers.clone())
+                .timeout(Duration::from_secs(http::WIRESERVER_HTTP_TIMEOUT_SEC))
+                .send()
+                .await
+            {
+                let statuscode = response.status();
 
-    if response.status().is_success() {
-        let body = response.text().await?;
+                if statuscode == StatusCode::OK {
+                    tracing::info!(
+                        "HTTP response succeeded with status {}",
+                        statuscode
+                    );
+                    return Ok(response);
+                }
 
-        let goalstate: Goalstate = from_str(&body)?;
-        Ok(goalstate)
-    } else {
-        Err(Error::HttpStatus {
-            endpoint: url.to_owned(),
-            status: response.status(),
-        })
-    }
+                if !http::RETRY_CODES.contains(&statuscode) {
+                    return response.error_for_status().map_err(|error| {
+                        tracing::error!(
+                            ?error,
+                            "{}",
+                            format!(
+                                "HTTP call failed due to status {}",
+                                statuscode
+                            )
+                        );
+                        error
+                    });
+                }
+            }
+
+            tracing::info!("Retrying to get HTTP response in {} sec, remaining timeout {} sec.", retry_interval.as_secs(), total_timeout.saturating_sub(now.elapsed()).as_secs());
+
+            tokio::time::sleep(retry_interval).await;
+        }
+    })
+    .await?;
+
+    let goalstate_body = response?.text().await?;
+
+    let goalstate: Goalstate = from_str(&goalstate_body)?;
+
+    Ok(goalstate)
 }
+
+const DEFAULT_HEALTH_URL: &str = "http://168.63.129.16/machine/?comp=health";
 
 pub async fn report_health(
     client: &Client,
     goalstate: Goalstate,
+    retry_interval: Duration,
+    total_timeout: Duration,
+    url: Option<&str>,
 ) -> Result<(), Error> {
-    let url = "http://168.63.129.16/machine/?comp=health";
+    let url = url.unwrap_or(DEFAULT_HEALTH_URL);
 
     let mut headers = HeaderMap::new();
     headers.insert("x-ms-agent-name", HeaderValue::from_static("azure-init"));
@@ -80,21 +130,40 @@ pub async fn report_health(
 
     let post_request = build_report_health_file(goalstate);
 
-    let response = client
-        .post(url)
-        .headers(headers)
-        .body(post_request)
-        .send()
-        .await?;
+    _ = timeout(total_timeout, async {
+        let now = std::time::Instant::now();
+        loop {
+            if let Ok(response) = client
+                .post(url)
+                .headers(headers.clone())
+                .body(post_request.clone())
+                .timeout(Duration::from_secs(http::WIRESERVER_HTTP_TIMEOUT_SEC))
+                .send()
+                .await
+            {
+                let statuscode = response.status();
 
-    if response.status().is_success() {
-        Ok(())
-    } else {
-        Err(Error::HttpStatus {
-            endpoint: url.to_owned(),
-            status: response.status(),
-        })
-    }
+                if statuscode == StatusCode::OK {
+                    tracing::info!("HTTP response succeeded with status {}", statuscode);
+                    return Ok(response);
+                }
+
+                if !http::RETRY_CODES.contains(&statuscode) {
+                    return response.error_for_status().map_err(|error| {
+                        tracing::error!(?error, "{}", format!("HTTP call failed due to status {}", statuscode));
+                        error
+                    });
+                }
+            }
+
+            tracing::info!("Retrying to get HTTP response in {} sec, remaining timeout {} sec.", retry_interval.as_secs(), total_timeout.saturating_sub(now.elapsed()).as_secs());
+
+            tokio::time::sleep(retry_interval).await;
+        }
+    })
+    .await?;
+
+    Ok(())
 }
 
 fn build_report_health_file(goalstate: Goalstate) -> String {
@@ -131,7 +200,15 @@ fn build_report_health_file(goalstate: Goalstate) -> String {
 
 #[cfg(test)]
 mod tests {
-    use super::{build_report_health_file, Goalstate};
+    use super::{
+        build_report_health_file, get_goalstate, report_health, Goalstate,
+    };
+
+    use reqwest::{header, Client, StatusCode};
+    use std::time::Duration;
+    use tokio::net::TcpListener;
+
+    use crate::{http, unittest};
 
     static GOALSTATE_STR: &str = "<Goalstate>
             <Container>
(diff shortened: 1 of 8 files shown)


libazureinit/src/goalstate.rs
```diff
@@ -41,34 +45,69 @@ pub struct RoleInstance {
...
+    .await?;
+
+    let goalstate_body = response?.text().await?;
+
+    let goalstate: Goalstate = from_str(&goalstate_body)?;
+
+    Ok(goalstate)
 }
 
+static DEFAULT_HEALTH_URL: &str = "http://168.63.129.16/machine/?comp=health";
```
jeremycline, line 98: Same static vs const note as above

libazureinit/src/goalstate.rs
```diff
@@ -41,34 +45,69 @@ pub struct RoleInstance {
...
+
 pub async fn report_health(
     client: &Client,
     goalstate: Goalstate,
+    retry_interval: Duration,
+    total_timeout: Duration,
+    input_url: Option<&str>,
 ) -> Result<(), Error> {
-    let url = "http://168.63.129.16/machine/?comp=health";
+    let url = match input_url {
```
jeremycline, line 107: You can use `unwrap_or` here as above

libazureinit/src/goalstate.rs
```diff
@@ -41,34 +45,69 @@ pub struct RoleInstance {
...
+                .timeout(Duration::from_secs(30))
+                .send()
+                .await
+            {
+                let statuscode = response.status();
 
-    if response.status().is_success() {
-        let body = response.text().await?;
+                if statuscode.is_success() && statuscode == StatusCode::OK {
+                    return response.error_for_status();
```
jeremycline, line 78: Since there's a check on the status code before we get here this is equivalent to `return Ok(response)`, which I think is a little clearer since it doesn't require the reader to know `error_for_status()` returns `Result<Self>` and can only ever return `Ok` in this case.

libazureinit/src/goalstate.rs
```diff
@@ -80,21 +119,33 @@ pub async fn report_health(
...
+                .headers(headers.clone())
+                .body(post_request.clone())
+                .timeout(Duration::from_secs(30))
+                .send()
+                .await
+            {
+                let statuscode = response.status();
+
+                if statuscode.is_success() && statuscode == StatusCode::OK {
+                    return response.error_for_status();
```
jeremycline, line 135: Same note as above

libazureinit/src/http.rs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use reqwest::StatusCode;
+
+// Set of StatusCodes that should be retried,
+// e.g. 400, 404, 410, 429, 500, 503.
```
jeremycline, line 7: Since this is a public constant, using `///` instead of `//` will cause it to be rendered in the HTML docs, and I think we should make an effort to have all public symbols documented. You can see how things look locally with `cargo doc -p libazureinit --no-deps`.

libazureinit/src/http.rs
```diff
@@ -0,0 +1,23 @@
...
+    StatusCode::BAD_REQUEST,
+    StatusCode::NOT_FOUND,
+    StatusCode::GONE,
+    StatusCode::TOO_MANY_REQUESTS,
+    StatusCode::INTERNAL_SERVER_ERROR,
+    StatusCode::SERVICE_UNAVAILABLE,
+];
+
+// Set of StatusCodes that should immediately fail,
+// e.g. 401, 403, 405.
```
jeremycline, line 18: Same note about rustdocs

jeremycline (changes requested):
Just a couple minor adjustments, things look good otherwise!

libazureinit/src/goalstate.rs
```diff
@@ -41,34 +45,69 @@ pub struct RoleInstance {
...
+pub async fn get_goalstate(
+    client: &Client,
+    retry_interval: Duration,
+    total_timeout: Duration,
+    input_url: Option<&str>,
+) -> Result<Goalstate, Error> {
+    let url = match input_url {
+        Some(url) => url,
+        None => DEFAULT_GOALSTATE_URL,
+    };
```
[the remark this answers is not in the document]
dongsupark: Done, thanks.

libazureinit/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 
 pub mod error;
 pub mod goalstate;
+pub mod http;
```
jeremycline, line 6: Sorry, should have caught this last time, but I noticed the HTTP constants aren't used in any public interface, so it would be best to not expose them publicly in the API. You can make them visible to everything within `libazureinit` (and not available outside it) by changing this to 
```suggestion
pub(crate) mod http;
```

That way we minimize the surface of our public API.

libazureinit/src/goalstate.rs
```diff
@@ -41,34 +45,63 @@ pub struct RoleInstance {
...
 
-        let goalstate: Goalstate = from_str(&body)?;
-        Ok(goalstate)
-    } else {
-        Err(Error::HttpStatus {
-            endpoint: url.to_owned(),
-            status: response.status(),
-        })
-    }
+                if http::RETRY_CODES.contains(&statuscode) {
```
jeremycline, line 78: I think you meant to have 
```suggestion
                if !http::RETRY_CODES.contains(&statuscode) {
```

Since this didn't get caught by the tests, I think we'll need to adjust them a bit to make sure we can assert retry codes do attempt several times. Maybe we just turn the retry interval and total wait time down to something acceptable for tests (100ms and 3 seconds respectively? lower even?) and not artificially skip time ahead in some or all tests.

libazureinit/src/goalstate.rs
```diff
@@ -80,21 +113,33 @@ pub async fn report_health(
...
+                .send()
+                .await
+            {
+                let statuscode = response.status();
+
+                if statuscode.is_success() && statuscode == StatusCode::OK {
+                    return Ok(response);
+                }
+
+                if http::RETRY_CODES.contains(&statuscode) {
```
jeremycline, line 132: Same here, I assume we want to return only if the status code _isn't_ in the list of retry codes.

libazureinit/src/http.rs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+use reqwest::StatusCode;
+
+/// Set of StatusCodes that should be retried,
+/// e.g. 400, 404, 410, 429, 500, 503.
+pub const RETRY_CODES: &[StatusCode] = &[
```
jeremycline, line 8: Similar to the `http` module, you can mark these constants as only available to the crate via `pub(crate)` so they're not automatically exposed if someone adds something we _do_ want public to the http module and then changes it from `pub(crate)` to `pub`

jeremycline (changes requested):
I lean towards having the tests not mess with time and just pass in significantly retry time limits, and then in the listener task you're spawning with tokio, counting the number of requests you get and asserting they're >1 for retry codes and ==1 for hard-fail codes.

libazureinit/src/goalstate.rs
```diff
@@ -41,34 +45,63 @@ pub struct RoleInstance {
...
     headers.insert("x-ms-version", HeaderValue::from_static("2012-11-30"));
 
-    let request = client.get(url).headers(headers);
-    let response = request.send().await?;
+    let response = timeout(total_timeout, async {
+        loop {
+            if let Ok(response) = client
+                .get(url)
+                .headers(headers.clone())
+                .timeout(Duration::from_secs(30))
```
anhvoms, line 68: consider moving the timeout value into the http module and call it WIRESERVER_HTTP_TIMEOUT_IN_SECONDS?

libazureinit/src/goalstate.rs
```diff
@@ -80,21 +113,33 @@ pub async fn report_health(
...
-            status: response.status(),
-        })
-    }
+    _ = timeout(total_timeout, async {
+        loop {
+            if let Ok(response) = client
+                .post(url)
+                .headers(headers.clone())
+                .body(post_request.clone())
+                .timeout(Duration::from_secs(30))
```
anhvoms, line 122: same comment here, I would suggest moving the 30s to a constant in the http module

libazureinit/src/goalstate.rs
anhvoms, line 78: what if there's a connection/read timeout error and there's no http error code? we want to retry those as well. Same would be applicable for the imds retry

libazureinit/src/goalstate.rs
```diff
@@ -41,34 +45,63 @@ pub struct RoleInstance {
...
+) -> Result<Goalstate, Error> {
+    let url = url.unwrap_or(DEFAULT_GOALSTATE_URL);
 
     let mut headers = HeaderMap::new();
     headers.insert("x-ms-agent-name", HeaderValue::from_static("azure-init"));
     headers.insert("x-ms-version", HeaderValue::from_static("2012-11-30"));
 
-    let request = client.get(url).headers(headers);
-    let response = request.send().await?;
+    let response = timeout(total_timeout, async {
```
anhvoms, line 63: Consider adding some logging whenever there's a failure, and indicate the retry attempt as well as the remaining timeout

anhvoms (changes requested):

libazureinit/src/goalstate.rs
dongsupark: Fixed the code as suggested, still some to do left.

libazureinit/src/goalstate.rs
dongsupark: Done

libazureinit/src/goalstate.rs
dongsupark: Now that the condition became different, the issue mentioned is automatically resolved.
That means, as we have `if !http::RETRY_CODES.contains`, here we return immediately for `HARDFAIL_CODES`. However, in case of other status code like `RETRY_CODES`, fall back to sleep and retry below.

jeremycline:
> > I lean towards having the tests not mess with time and just pass in significantly retry time limits, and then in the listener task you're spawning with tokio, counting the number of requests you get and asserting they're >1 for retry codes and ==1 for hard-fail codes.
> 
> Sorry, I am not sure I understand quite well. How would it be possible to count the number of requests on the listener task side? What I came up with is like [this](https://github.com/dongsupark/azure-init/commit/ea0f812af7a588d21b9e16947ec11244727fe379), checking for readable/writable from the spawned task, and reduce total timeout to avoid being blocked for long time. Is that what you mean?

Ah, sorry for not being clearer. Here's a not-very-tidy example implementation: https://github.com/Azure/azure-init/commit/2388cca14838f4848c6c68b20eb84fbc1a24d427

The general idea is that you spawn the server loop into a Tokio task (as you already have) along with a way to cancel the task when we want to see how many connections have been handled. Tasks can return values and be joined by awaiting the handle you get from `tokio::spawn` which is how we get the count back.

Do let me know if you've got any other questions, happy to help work through the finer details if this doesn't cover everything!

dongsupark:
> Ah, sorry for not being clearer. Here's a not-very-tidy example implementation: https://github.com/Azure/azure-init/commit/2388cca14838f4848c6c68b20eb84fbc1a24d427
> 
> The general idea is that you spawn the server loop into a Tokio task (as you already have) along with a way to cancel the task when we want to see how many connections have been handled. Tasks can return values and be joined by awaiting the handle you get from `tokio::spawn` which is how we get the count back.

Thanks, nice idea, updated the PR.

dongsupark:
Split `serve_requests` into a separate unittest mod.

dongsupark:
Split `get_http_response_payload` into the unittest mod.

libazureinit/src/goalstate.rs
```diff
@@ -41,34 +45,72 @@ pub struct RoleInstance {
...
+                if !http::RETRY_CODES.contains(&statuscode) {
+                    return response.error_for_status().map_err(|err| {
+                        tracing::error!(error = ?err, "{}", format!("HTTP response failed immediately due to status {}", statuscode));
+                        err
+                    });
+                }
+            }
+
+            rest_timeout = rest_timeout.saturating_sub(Duration::from_secs(http::IMDS_HTTP_TIMEOUT_SEC));
+            tracing::info!("{}", format!("Retrying to get HTTP response in {} sec, remaining timeout {} sec.", retry_interval.as_secs(), rest_timeout.as_secs()));
```
jeremycline, line 90: Since `rest_timeout` is reset to `total_timeout` each loop, I think this always prints the same value for the remaining timeout.

You can use the [Instant](https://doc.rust-lang.org/std/time/struct.Instant.html) type to make the accounting easier: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=cbfb83556466fffc87be4101b7a4d062

libazureinit/src/goalstate.rs
```diff
@@ -80,21 +122,42 @@ pub async fn report_health(
...
```
jeremycline, line 153: Same thing here about the remaining timeout logging.

libazureinit/src/goalstate.rs
```diff
@@ -41,34 +45,72 @@ pub struct RoleInstance {
...
 
-        let goalstate: Goalstate = from_str(&body)?;
-        Ok(goalstate)
-    } else {
-        Err(Error::HttpStatus {
-            endpoint: url.to_owned(),
-            status: response.status(),
-        })
-    }
+                if statuscode.is_success() && statuscode == StatusCode::OK {
```
anhvoms, line 76: this block of code is repeated three times between goalstate, imds, and report health. I would recommend refactoring this into http module so that we can pass in arguments like url, total_timeout, per call timeout, retry codes, etc...
I'm ok with the PR going in the way it is and the refactor to reduce redundant code happening in a follow-up PR

libazureinit/src/imds.rs
```diff
@@ -90,54 +91,55 @@ where
...
+                if !http::RETRY_CODES.contains(&statuscode) {
+                    return response.error_for_status().map_err(|err| {
+                        tracing::error!(error = ?err, "{}", format!("HTTP response failed immediately due to status {}", statuscode));
+                        err
+                    });
                 }
             }
 
+            rest_timeout = rest_timeout.saturating_sub(Duration::from_secs(http::IMDS_HTTP_TIMEOUT_SEC));
+            tracing::info!("{}", format!("Retrying to get HTTP response in {} sec, remaining timeout {} sec.", retry_interval.as_secs(), rest_timeout.as_secs()));
```
jeremycline, line 141: Same note about the remaining timeout log from `goalstate.rs`

libazureinit/src/goalstate.rs
```diff
@@ -41,34 +45,72 @@ pub struct RoleInstance {
...
+                    tracing::info!("{}", format!("HTTP response succeeded with status {}", statuscode));
+                    return Ok(response);
+                }
+
+                if !http::RETRY_CODES.contains(&statuscode) {
+                    return response.error_for_status().map_err(|err| {
+                        tracing::error!(error = ?err, "{}", format!("HTTP response failed immediately due to status {}", statuscode));
```
anhvoms, line 83: In the case of timeout (connection/read), it is not "immediate"

```suggestion
                        tracing::error!(error = ?err, "{}", format!("HTTP response failed due to status {}", statuscode));
```

libazureinit/src/goalstate.rs
```diff
@@ -80,21 +122,42 @@ pub async fn report_health(
...
+                let statuscode = response.status();
+
+                if statuscode.is_success() && statuscode == StatusCode::OK {
+                    tracing::info!("{}", format!("HTTP response succeeded with status {}", statuscode));
+                    return Ok(response);
+                }
+
+                if !http::RETRY_CODES.contains(&statuscode) {
+                    return response.error_for_status().map_err(|err| {
+                        tracing::error!(error = ?err, "{}", format!("HTTP response failed immediately due to status {}", statuscode));
```
anhvoms, line 146:
```suggestion
                        tracing::error!(error = ?err, "{}", format!("HTTP response failed due to status {}", statuscode));
```

libazureinit/src/imds.rs
```diff
@@ -90,54 +91,55 @@ where
...
+                        )
+                    );
+                    return Ok(response);
                 }
 
-                if !RETRY_CODES.contains(&statuscode) {
-                    return response.error_for_status();
+                if !http::RETRY_CODES.contains(&statuscode) {
+                    return response.error_for_status().map_err(|err| {
+                        tracing::error!(error = ?err, "{}", format!("HTTP response failed immediately due to status {}", statuscode));
```
anhvoms, line 134:
```suggestion
                        tracing::error!(error = ?err, "{}", format!("HTTP call failed due to status {}", statuscode));
```

libazureinit/src/lib.rs
```diff
@@ -14,5 +15,7 @@ pub use provision::{
     Provision,
 };
 
+mod unittest;
```
jeremycline, line 18: Maybe 
```suggestion
#[cfg(test)]
mod unittest;
```
so it only shows up to the tests? I'm assuming this works, but I've not tried it myself.

anhvoms (changes requested):
It looks quite good now. I left a few minor comments.

libazureinit/src/unittest.rs
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+#[cfg(test)]
+pub(crate) mod unittests {
```
jeremycline, line 5: If you go with the suggestion in `lib.rs` about guarding the mod declaration with `#[cfg(test)]` you should be able to drop this mod declaration as well.

libazureinit/src/unittest.rs
```diff
@@ -0,0 +1,54 @@
...
+    use reqwest::StatusCode;
+
+    use tokio::io::AsyncWriteExt;
+    use tokio::net::TcpListener;
+    use tokio_util::sync::CancellationToken;
+
+    // Returns expected HTTP response for the given status code and body string.
+    pub(crate) async fn get_http_response_payload(
```
jeremycline, line 13: Since the function doesn't block on any other futures you can declare it as 
```suggestion
    pub(crate) fn get_http_response_payload(
```

jeremycline (changes requested):
I think the logging for the timeout remaining will produce confusing results in the event that a timeout occurs more than once; all my other notes are just nitpicking and optional style things.

Sorry for so many rounds of changes requested, I should have been more careful in initial reviews and caught things earlier - I think this looks solid and is _nearly_ there.

libazureinit/src/goalstate.rs
dongsupark: Agreed in general.
Looking at the code once again, `.is_success()` is actually not needed, because `success` means code from 200 to 300.
So I simply removed the first condition, and kept only the second.

libazureinit/src/goalstate.rs
dongsupark: Done

libazureinit/src/goalstate.rs
```diff
@@ -41,34 +45,72 @@ pub struct RoleInstance {
...
+                        err
+                    });
+                }
+            }
+
+            rest_timeout = rest_timeout.saturating_sub(Duration::from_secs(http::IMDS_HTTP_TIMEOUT_SEC));
+            tracing::info!("{}", format!("Retrying to get HTTP response in {} sec, remaining timeout {} sec.", retry_interval.as_secs(), rest_timeout.as_secs()));
+
+            tokio::time::sleep(retry_interval).await;
```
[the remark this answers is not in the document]
dongsupark: Now that we use `Instant.now`, the issue is automatically gone.

libazureinit/src/goalstate.rs
```diff
@@ -41,34 +45,72 @@ pub struct RoleInstance {
...
-    }
+                if statuscode.is_success() && statuscode == StatusCode::OK {
+                    tracing::info!("{}", format!("HTTP response succeeded with status {}", statuscode));
+                    return Ok(response);
+                }
+
+                if !http::RETRY_CODES.contains(&statuscode) {
+                    return response.error_for_status().map_err(|err| {
+                        tracing::error!(error = ?err, "{}", format!("HTTP response failed immediately due to status {}", statuscode));
+                        err
```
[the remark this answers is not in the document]
dongsupark: Done

libazureinit/src/unittest.rs
dongsupark: Done

libazureinit/src/unittest.rs
dongsupark: Done

libazureinit/src/goalstate.rs
dongsupark: Done

libazureinit/src/goalstate.rs
```diff
@@ -41,34 +45,82 @@ pub struct RoleInstance {
...
+                                statuscode
+                            )
+                        );
+                        error
+                    });
+                }
+            }
+
+            _ = total_timeout.saturating_sub(now.elapsed()).as_secs();
+            tracing::info!("Retrying to get HTTP response in {} sec, remaining timeout {} sec.", retry_interval.as_secs(), total_timeout.as_secs());
```
jeremycline, line 100: Based on my reading of the API, [saturating_sub](https://doc.rust-lang.org/std/time/struct.Duration.html#method.saturating_sub) takes ownership of `total_timeout` and the value returned from `now.elapsed()` and returns a _new_ `Duration` with the result. Since `total_timeout` isn't declared as `mut` that also makes me expect this line to have no side-effects.

I think what you want is
```suggestion
            tracing::info!("Retrying to get HTTP response in {} sec, remaining timeout {} sec.", retry_interval.as_secs(),  total_timeout.saturating_sub(now.elapsed()).as_secs());
```

jeremycline (changes requested):
The notes about the instant inside the loop and the log line apply to all the functions implementing the timeout

libazureinit/src/goalstate.rs
```diff
@@ -41,34 +45,82 @@ pub struct RoleInstance {
...
 
     let mut headers = HeaderMap::new();
     headers.insert("x-ms-agent-name", HeaderValue::from_static("azure-init"));
     headers.insert("x-ms-version", HeaderValue::from_static("2012-11-30"));
 
-    let request = client.get(url).headers(headers);
-    let response = request.send().await?;
+    let response = timeout(total_timeout, async {
+        loop {
+            let now = std::time::Instant::now();
```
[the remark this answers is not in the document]
dongsupark: Done, thanks

jeremycline (approved):
Looks good to me, thanks 💖

dongsupark:
@anhvoms Merging is blocked, could you please have another look?

anhvoms (approved):
LGTM!

PR #123, opened 2024-09-05
Retrieve correct authorizedkeysfile path

### LINKED ISSUE #97: [RFE] retrieve the correct path for authorized key file from sshd config
## Current situation
azure-init assumes that the authorized key file is $HOME/.ssh/authorized_keys which might not be what the sshd is configured to look for (/etc/sshd/config/AuthorizedKeysFile)

## Impact

azure-init might write the provided public key to a place that is not expected by sshd, which will render the VM inaccessible via the provided key

## Ideal future situation

azure-init should process the sshd config to correctly determine where to write the authorized key file.

## **Implementation options

We likely need a module to handle various ssh related functionality. #67 will also benefit from such module

balakreddy:
By default azure-init assumes the authorized keys file is $HOME/.ssh/authorized_keys, this function retrieves informartion from sshd config using sshd -G. If there exists no information it will return None. Fixes #97.

### DIFF
--- a/libazureinit/src/provision/ssh.rs
+++ b/libazureinit/src/provision/ssh.rs
@@ -3,11 +3,13 @@
 
 use std::{
     fs::Permissions,
-    io::Write,
+    io::{self, Write},
     os::unix::fs::{DirBuilderExt, PermissionsExt},
+    path::PathBuf,
+    process::Command,
 };
 
-use tracing::instrument;
+use tracing::{info, instrument};
 
 use crate::error::Error;
 use crate::imds::PublicKeys;
@@ -27,7 +29,11 @@ pub(crate) fn provision_ssh(
     // the permissions are correct.
     std::fs::set_permissions(&ssh_dir, Permissions::from_mode(0o700))?;
 
-    let authorized_keys_path = ssh_dir.join("authorized_keys");
+    let authorized_keys_path = match get_authorized_keys_path_from_sshd()? {
+        Some(path) => user.dir.join(path),
+        None => user.dir.join(".ssh/authorized_keys"),
+    };
+    info!("Using authorized_keys path: {:?}", authorized_keys_path);
     let mut authorized_keys = std::fs::File::create(&authorized_keys_path)?;
     authorized_keys.set_permissions(Permissions::from_mode(0o600))?;
     keys.iter()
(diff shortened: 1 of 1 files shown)


libazureinit/src/provision/ssh.rs
```diff
@@ -37,17 +39,80 @@ pub(crate) fn provision_ssh(
     Ok(())
 }
 
+pub(crate) fn get_authorized_keys_path_from_sshd() -> io::Result<Option<PathBuf>>
```
dongsupark, line 42: `cargo check` gives warning below. Where is the function being called?
```
warning: function `get_authorized_keys_path_from_sshd` is never used
  --> libazureinit/src/provision/ssh.rs:42:15
   |
42 | pub(crate) fn get_authorized_keys_path_from_sshd() -> io::Result<Option<PathBuf>>
   |               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

warning: `libazureinit` (lib) generated 1 warning
```

balakreddy: This function will be used in provision_ssh in ssh.rs to provision the ssh keys on line 30. I will make the changes to use get_authorized_keys_path_from_sshd.

libazureinit/src/provision/ssh.rs
```diff
@@ -37,17 +39,80 @@ pub(crate) fn provision_ssh(
...
+{
+    let sshd_output = Command::new("sshd").arg("-G").output()?;
```
anhvoms, line 44: -G doesn't work for me on Ubuntu. Use -T?

libazureinit/src/provision/ssh.rs
```diff
@@ -37,17 +39,80 @@ pub(crate) fn provision_ssh(
...
+        ));
+    }
+
+    let stdout = sshd_output.stdout;
+    let sshd_output = String::from_utf8_lossy(&stdout);
+    for line in sshd_output.lines() {
+        if line.starts_with("authorizedkeysfile") {
+            let keypath: Vec<&str> = line.split_whitespace().collect();
+            if keypath.len() > 1 {
+                return Ok(Some(PathBuf::from(keypath[1])));
```
anhvoms, line 59: Please add a tracing call to output the path that we use

anhvoms (changes requested):

libazureinit/src/provision/ssh.rs
balakreddy: Done

libazureinit/src/provision/ssh.rs
balakreddy: Used -T for sshd config

dongsupark (approved):
Looks good. I have also run functional tests, which also passed.

@anhvoms should also have a look.

anhvoms (approved):

PR #133, opened 2024-09-20
AzureKVP Telemetry Library for Hyper-V Integration

### LINKED ISSUE #27: Report KVP telemetry to assist with troubleshooting

peytonr18:
The azurekvp library provides a method for collecting and encoding tracing telemetry, designed specifically for integration with Hyper-V. This library captures span data through OpenTelemetry and Rust-based tracing, formats it into Hyper-V Key-Value Pair (KVP) format, and writes it directly to the Hyper-V KVP storage at /var/lib/hyperv/.kvp_pool_1.
 
Key changes in this PR:
- Telemetry Capture: Integrates with the existing tracing infrastructure to capture and transmit telemetry data in the correct format.
- KVP Encoding: Encodes span data into the Hyper-V-specific KVP format for storage and retrieval.
- Unit Tests: Includes unit tests for KVP formatting and verifies the correctness of the encoded telemetry.
- Documentation: Updates the README to include details on the new azurekvp library and the updated tracing process. 

Fixes #27


azurekvp/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+pub mod main;
```
dongsupark, line 4: Why is that needed?
It feels like some unnecessary line that can be cleaned up.

azurekvp/src/tracing.rs
```diff
@@ -0,0 +1,219 @@
...
+use std::path::Path;
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+use uuid::Uuid;
+
+const HV_KVP_EXCHANGE_MAX_KEY_SIZE: usize = 512;
+const HV_KVP_EXCHANGE_MAX_VALUE_SIZE: usize = 2048;
+const HV_KVP_AZURE_MAX_VALUE_SIZE: usize = 1024;
+const EVENT_PREFIX: &str = "AZURE_INIT";
+
+pub fn initialize_tracing() -> sdktrace::Tracer {
```
dongsupark, line 29: Can you please add a comment `///...` for rustdocs?

azurekvp/src/tracing.rs
```diff
@@ -0,0 +1,219 @@
...
+    let provider = sdktrace::TracerProvider::builder().build();
+    global::set_tracer_provider(provider.clone());
+    provider.tracer("azure-kvp")
+}
+
+pub fn handle_span<S>(
```
dongsupark, line 35: Same, rustdocs

azurekvp/src/tracing.rs
```diff
@@ -0,0 +1,219 @@
...
+fn encode_kvp_slice(key: Vec<u8>, value: Vec<u8>) -> Vec<u8> {
+    let mut buffer = Vec::with_capacity(
+        HV_KVP_EXCHANGE_MAX_KEY_SIZE + HV_KVP_EXCHANGE_MAX_VALUE_SIZE,
+    );
+    buffer.extend_from_slice(&key);
+    buffer.extend_from_slice(&value);
+    buffer
+}
+
+pub fn decode_kvp_item(
```
dongsupark, line 149: Same, rustdocs

azurekvp/src/tracing.rs
```diff
@@ -0,0 +1,219 @@
...
+        record_data[HV_KVP_EXCHANGE_MAX_KEY_SIZE..record_data_len].to_vec(),
+    )
+    .unwrap_or_else(|_| String::new())
+    .trim_end_matches('\x00')
+    .to_string();
+
+    Ok((key, value))
+}
+
+pub fn truncate_guest_pool_file(kvp_file: &Path) -> Result<(), Error> {
```
dongsupark, line 177: Why does the function need to be `pub`?
That is not used anywhere outside azurekvp.

spans.log
```diff
@@ -0,0 +1,388 @@
+SpanData {
```
dongsupark, line 1: We do not need to commit the log file, right?

src/main.rs
```diff
@@ -42,6 +50,41 @@ struct Cli {
     )]
     groups: Vec<String>,
 }
+pub struct EmitKVPLayer {
+    file_path: std::path::PathBuf,
+}
+
```
dongsupark, line 56: Adding the KVP code directly in `main.rs`, the file is getting bigger and becoming a mix of different things.
Would it be like a good idea to split the KVP/tracing-related part into another file in the top level, for example `kvp.rs`?

dongsupark (commented):
Thanks for hard work since a long time.

I think it might be a good idea to squash commits into a single one, as it is almost impossible to track the whole history of removing, renaming, or restructuring.

azurekvp/src/lib.rs
peytonr18: I think I had originally set it up this way before I was sure how we were going to split the tracing & main functionality (and forgot to remove it later on) but you are right, we don't need it!! I will remove it!

azurekvp/src/tracing.rs
peytonr18: I'll do a full writeup for all of the functions for rustdocs.

spans.log
peytonr18: Correct -- will remove that from the PR.

README.md
```diff
@@ -48,3 +48,68 @@ Any use of third-party trademarks or logos are subject to those third-party's po
...
+- **Purpose**: This layer integrates with the OpenTelemetry framework to handle context propagation and export span data to an external tracing backend (e.g., Jaeger, Prometheus) or to stdout.
+- **How It Works**:
+  - As spans are created and processed, the `OpenTelemetryLayer` ensures that context is propagated correctly across different parts of the program, which is crucial in distributed systems for tracing requests across service boundaries.
+  - The span data is then exported to a configured backend or stdout, where it can be visualized and analyzed using OpenTelemetry-compatible tools.
+
+#### 3. stderr_layer
+
+- **Purpose**: This layer formats and logs span data to stderr or a specified log file, providing a human-readable output for immediate inspection.
+- **How It Works**:
+  - Each span’s lifecycle events are logged in a structured format, making it easy to see the flow of execution in the console or log files.
```
cjp256, line 84:
```suggestion
  - Each span's lifecycle events are logged in a structured format, making it easy to see the flow of execution in the console or log files.
```

README.md
```diff
@@ -48,3 +48,68 @@ Any use of third-party trademarks or logos are subject to those third-party's po
 ## libazureinit
 
 For common library used by this reference implementation, please refer to [libazureinit](https://github.com/Azure/azure-init/tree/main/libazureinit/).
+
+## Tracing Logic Overview
```
cjp256, line 52: maybe this should be its own doc under /doc/?

README.md
```diff
@@ -48,3 +48,68 @@ Any use of third-party trademarks or logos are subject to those third-party's po
...
+2. **OpenTelemetryLayer**: Context Propagation and Span Export
+3. **stderr_layer**: Formatting and Logging to stderr
+
+These layers work together, yet independently, to process span data as it flows through the program.
+
+### Layer Overview
+
+#### 1. EmitKVPLayer
+
+- **Purpose**: This custom layer is responsible for processing spans by capturing their metadata, generating key-value pairs, encoding them into a specific format, and writing the encoded data to the VM's hyperV file.
```
cjp256, line 68: Maybe drop a link to 

https://learn.microsoft.com/en-us/virtualization/hyper-v-on-windows/reference/integration-services#hyper-v-data-exchange-service-kvp

Indicate this is meant for consumption by `hv_kvp_daemon` service, which typically reads from `/var/lib/hyperv/.kvp_pool_1`.

azurekvp/src/tracing.rs
```diff
@@ -0,0 +1,219 @@
...
+        .map_err(|e| Error::new(std::io::ErrorKind::Other, e))?
+        .as_secs()
+        - get_uptime()?.as_secs();
+
+    match kvp_file.metadata() {
+        Ok(metadata) => {
+            if metadata.mtime() < boot_time as i64 {
+                OpenOptions::new()
+                    .write(true)
+                    .truncate(true)
```
cjp256, line 189: I know this is how cloud-init does it, but I don't think we should be in the business of truncating the KVP pool.  If the pool shouldn't persist across boots, it needs to be moved to /run or similar.

But then the question becomes can we handle a persistent KVP pool rather than assuming clean slate...

Not asking for a change, but something to ponder (how would we support a different path if we fixed this).

azurekvp/src/tracing.rs
```diff
@@ -0,0 +1,219 @@
...
+use std::os::fd::AsRawFd;
+use std::os::unix::fs::MetadataExt;
+use std::path::Path;
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+use uuid::Uuid;
+
+const HV_KVP_EXCHANGE_MAX_KEY_SIZE: usize = 512;
+const HV_KVP_EXCHANGE_MAX_VALUE_SIZE: usize = 2048;
+const HV_KVP_AZURE_MAX_VALUE_SIZE: usize = 1024;
+const EVENT_PREFIX: &str = "AZURE_INIT";
```
cjp256, line 27: A bit cleaner, I don't think the caps is needed.

```suggestion
const EVENT_PREFIX: &str = "azure-init";
```

azurekvp/src/tracing.rs
```diff
@@ -0,0 +1,219 @@
...
+
+    let end_time_dt = DateTime::<Utc>::from(
+        SystemTime::UNIX_EPOCH
+            + std::time::Duration::from_millis(end_time as u64),
+    )
+    .format("%Y-%m-%dT%H:%M:%S%.3fZ")
+    .to_string();
+
+    let event_key = generate_event_key(
+        &span_context.level().to_string(),
```
cjp256, line 69: I don't think level needs to be included for KVP.  Either it's important enough to be emitted (or not).

azurekvp/src/tracing.rs
cjp256: It'd be great if it included the version too.

azurekvp/src/tracing.rs
cjp256: How does the KVP incorporate the result of the span?

azurekvp/src/tracing.rs
```diff
@@ -0,0 +1,219 @@
...
+use std::fs::OpenOptions;
+use std::io::ErrorKind;
+use std::io::{Error, Write};
+use std::os::fd::AsRawFd;
+use std::os::unix::fs::MetadataExt;
+use std::path::Path;
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+use uuid::Uuid;
+
+const HV_KVP_EXCHANGE_MAX_KEY_SIZE: usize = 512;
```
cjp256, line 24: please move all hv hvp interface bits into its own module and decouple it from tracing, etc.   it should export something like a write(key, value) which the kvp span layer can use

README.md
peytonr18: Agreed -- I will make a /doc folder and add it there instead of in the ReadME.

azurekvp/src/tracing.rs
peytonr18: I think, as it is, it doesn't explicitly incorporate the result of the span into the key:value pair generated. But, I think that is something we ought to add. I'll create a specific field in the event_value to include a status for every span.

src/main.rs
peytonr18: Completely agree. I've separated all the kvp/tracing logic so that it is split into its own file (we'll have tracing.rs for all tracing logic and kvp.rs for all HyperV KVP related things).

src/main.rs
```diff
@@ -42,6 +50,41 @@ struct Cli {
...
+        _attrs: &Attributes<'_>,
+        id: &Id,
+        ctx: TracingContext<'_, S>,
+    ) {
+        let start_time = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap()
+            .as_millis();
+        if let Some(span) = ctx.span(id) {
+            span.extensions_mut().insert(start_time);
```
jeremycline, line 78: Okay, so there are a few things here including a gotcha to how extensions work.

## Instant
First, there's a neat type available to calculate durations without needing to touch `SystemTime` at all called [Instant](https://doc.rust-lang.org/std/time/struct.Instant.html). If you create an `Instant`, you can later query how long it's been with:
```Rust
let then = std::time::Instant::now();

// A Duration with the time between `then` and `now`
let now = now.elapsed();
```

You can also compare between instants with `Instant::duration_since` if you have two instants and want to know the time between them, but you shouldn't need that here.

## Extensions
Extensions are span-specific, which means layers from other libraries (OpenTelemetry, the layers that come with tracing-subscriber, some other custom layer we might add later) can see and access the extension. This is mostly not a problem, with one gotcha: the extension entry is looked up by its _type_.

This means that if another layer wants to insert an extension that is the same type as the type we add here, it'll cause the program to panic. `start_time` is `u128` so if any other layer tries to insert a `u128` we'll crash. There's a line at the end of the [insert](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/registry/struct.ExtensionsMut.html#method.insert) API doc talks a bit about this, but if you're not already familiar with the pattern it is probably confusing. It says that extensions should generally be "newtypes".

## Newtype pattern

The newtype pattern is when you make a new type to wrap some other type. This is useful in scenarios where you want to provide some additional functionality on top of an existing type, or in cases like this where we really want to make sure no other layer could provide a conflicting type. It looks like this when applied to the standard library's `Instant` type.

```Rust
// Define a new tuple struct with a single member, the type we're wrapping.
struct MyInstant(std::time::Instant);

fn main() {
    let then = MyInstant(std::time::Instant::now());
    let now = then.0.elapsed();
}
```

The neat thing about this is that because we don't mark `MyInstant` as `pub`, we can be certain no other library could possibly insert the same type into the span's extensions since we're the only crate with access.

### Deref

This is not necessary to do, but it's nice to know. It can be a little annoying to do the `instant.0.elapsed()` thing, so one thing we can do is implement [Deref](https://doc.rust-lang.org/std/ops/trait.Deref.html) for our new type:

```Rust
struct MyInstant(std::time::Instant);

impl std::ops::Deref for MyInstant {
    type Target = std::time::Instant;

    fn deref(&self) -> &Self::Target {
        &self.0
    }
}

fn main() {
    let then = MyInstant(std::time::Instant::now());
    let now = then.elapsed();
}
```
The compiler will silently insert calls to our `deref` implementation so we can treat references to our new type as references to the inner type. This is neat, but as the documentation on the trait mentions, it's something [you want to be careful with](https://doc.rust-lang.org/std/ops/trait.Deref.html#when-to-implement-deref-or-derefmut). In this case it's okay, though, since it's a private type and we're implementing no additional methods.

## The Fix

All this to say, you'll want to use `Instant` instead of converting to a time since the epoch, and you'll want to wrap it in a new type. You'll need to do some re-arranging of the code to make that work, since currently `handle_span` is in a different crate (and since we want to keep our new type private, the `azurekvp` crate won't be able to access it).

My recommendation would be to move the contents of the `azurekvp` crate into `src/kvp.rs` so it's part of the `azure-init` crate. You'll need to add a `mod kvp;` to `src/main.rs`.

If we later want to provide a generic, sharable implementation of a tracing layer for KVP we can break it out into its own crate, but for now I don't think we should split them up.

azurekvp/src/tracing.rs
jeremycline, line 35: Since function is used exclusively for the `on_close` handler, I would recommend just moving the contents into that function and then attaching your helper functions as methods on your `EmitKVPLayer` struct

jeremycline (changes requested):
Looks like a good start!

We'll need to do some thinking about how to handle errors inside the `on_close` span handler since there's a lot of fallible calls and the interface we need to implement doesn't let us return errors, but I decided not to dive into that much just yet since I need to think about it a bit and play with some approaches before suggesting things and it's lunch time.

peytonr18:
Squash merging got really messed up... am working on trying to get all of these into one

azurekvp/src/tracing.rs
```diff
@@ -0,0 +1,154 @@
...
+
+    let event_time_dt = DateTime::<Utc>::from(
+        UNIX_EPOCH
+            + std::time::Duration::from_millis(
+                event_time.duration_since(UNIX_EPOCH).unwrap().as_millis()
+                    as u64,
+            ),
+    )
+    .format("%Y-%m-%dT%H:%M:%S%.3fZ")
+    .to_string();
```
jeremycline, line 142: It looks like this function is used exclusively in the `on_event` handler and that `event_instant` is created immediately before calling this function. Unless I'm missing something, to calculate the event time what we want is `SystemTime::now`. We also want to be careful and avoid panicking unless absolutely necessary. Since we have the span duration, we can fall back to using that if calculating the duration since the epoch fails, or failing that use the default value for Duration:

```rust
    // Calculate the current time, falling back to the elapsed time in the unreasonable event that the
    // system time is before the epoch.
    let event_time = SystemTime::now()
        .duration_since(UNIX_EPOCH)
        .unwrap_or_else(|_| {
            span.extensions()
                .get::<MyInstant>()
                .map(|instant| instant.elapsed())
                .unwrap_or_default()
        });
    // Note that we don't need the `to_string()` since we immediately pass it on to `format!()`
    let event_time_dt = DateTime::<Utc>::from(UNIX_EPOCH + event_time)
        .format("%Y-%m-%dT%H:%M:%S%.3fZ");
```

azurekvp/src/kvp.rs
```diff
@@ -0,0 +1,503 @@
...
+            string: &mut event_message,
+        };
+
+        event.record(&mut visitor);
+
+        if let Some(span) = ctx.lookup_current() {
+            let mut extensions = span.extensions_mut();
+
+            if event.metadata().level() == &tracing::Level::ERROR {
+                extensions.insert("failure".to_string());
```
jeremycline, line 106: Similar to the `Instant` extension, we don't want to use `String` as the type, we'll want a new private type. Additionally, the new type could be an `enum` instead of wrapping a string. Or, if "failure" is the only state (and the absence of the extension signals not-failure) you could define an empty struct and use that: `struct Failure;`

azurekvp/src/kvp.rs
```diff
@@ -0,0 +1,503 @@
...
+            let mut event_level = "INFO".to_string();
+
+            if let Some(recorded_status) = span.extensions().get::<String>() {
+                if recorded_status == "failure" {
+                    status = "failure".to_string();
+                    event_level = "ERROR".to_string();
+                }
+            }
+
+            let end_time = SystemTime::now();
```
jeremycline, line 149: Since this is just grabbing the current time, it seems reasonable to push this call down into the `handle_span` call (unless it's being used to make testing easier, which I didn't spot)

azurekvp/src/kvp.rs
```diff
@@ -0,0 +1,503 @@
...
+
+            handle_span(
```
jeremycline, line 151: I know folks expressed a desire to split this stuff out, but I don't think it really makes sense to have this live in a separate module. I'd really prefer this function just be inlined here, or added as another private method on this struct.

If we want to provide a general-purpose library for folks to use that's fine, but I think we should do that _after_ we come up with an implementation we're satisfied with. Prematurely generalizing makes it difficult to work on and review, in my opinion.

azurekvp/src/kvp.rs
```diff
@@ -0,0 +1,503 @@
...
+        if let Some(span) = ctx.lookup_current() {
+            let mut extensions = span.extensions_mut();
+
+            if event.metadata().level() == &tracing::Level::ERROR {
+                extensions.insert("failure".to_string());
+            }
+
+            let event_instant = MyInstant::now();
+
+            handle_event(&event_message, &span, &self.file_path, event_instant);
```
jeremycline, line 111: Similar to my note below about `handle_span`, I'd really like this function to live here since it's used here exclusively.

azurekvp/src/kvp.rs
```diff
@@ -0,0 +1,503 @@
...
+        match OpenOptions::new().append(true).create(true).open(file_path) {
+            Ok(file) => file,
+            Err(e) => {
+                eprintln!("Failed to open log file: {}", e);
+                return Err(e); // Return the error if the file can't be opened
+            }
+        };
+
+    let fd = file.as_raw_fd();
+    if let Err(e) = flock(fd, FlockArg::LockExclusive) {
```
jeremycline, line 219: Regarding this lock, is it here in order to ensure azure-init doesn't step on itself, or do all users of the kvp files use `flock`?

peytonr18: Yes -- I wanted to make sure azure-init didn't overwrite itself while capturing the telemetry, and locking the file was my way of preventing concurrent writes. But if there is a better way, I'm all ears!

jeremycline: Ah cool, that makes things easier. What we can do is create a task (this would work with a dedicated operating system thread, but we're already using Tokio so we can lean into that) which is the owner of the file. Anyone that wants to write to the file needs to do it by sending it the thing to write over a channel. That way we know there's only one writer to the file and we can skip the locking.

Here's an example of what that could look like:

```rust
pub struct EmitKVPLayer {
    // We can use this to send logs to the writer task from the layer impl on this struct
    // String type used here just as an example, could be anything
    events_tx: tokio::sync::mpsc::UnboundedSender<String>,
}

impl EmitKVPLayer {
    pub fn new(file_path: std::path::PathBuf) -> Result<Self, std::io::Error> {
        // Truncating/opening the file here lets us return an error at layer creation time if there's a problem
        truncate_guest_pool_file(&file_path)?;
        let file = OpenOptions::new().append(true).create(true).open(&file_path)?;

        // The channel has no size bounds, so it could consume an abritrarily large
        // amount of memory. An alternative would be to use a channel that has a limited size
        // and if messages are pushed faster than they can be written out, we could opt to drop
        // messages.
        let (events_tx, events_rx) = tokio::sync::mpsc::unbounded_channel();

        // We should probably hold onto the handle to gracefully cancel the task, and maybe examine the return
        // for errors, but there might not be a nice place for that.
        let task = tokio::spawn(Self::kvp_writer(file, events_rx));

        Ok(Self { events_tx })
    }

    // Forward events from a channel to the KVP file.
    //
    // This function is intented to be run as a Tokio task, and there should be one for the file being
    // written to. Doing this allows us to ensure all file writes are serialized so we don't need to
    // worry about file locking.
    async fn kvp_writer(mut file: File, mut events: tokio::sync::mpsc::UnboundedReceiver<String>) -> Result<(), std::io::Error>{
        while let Some(message) = events.recv().await {
            // We could probably use tracing::error! here if we ensure there's a filter for this layer so it ignores its
            // own logs. That way it makes it to stderr or whatever other layers are configured, but presumably we're
            // logging due to an error and don't want to loop endlessly. With this implementation nothing examines the
            // return.
            file.write_all(message.as_bytes())?;
            file.flush()?;
        }
        Ok(())
    }
}
```

This is a somewhat mangled implementation of [this actor pattern](https://ryhl.io/blog/actors-with-tokio/). It looks a little difference since we don't need to hand out a reference to the channel sender, and we have no reason to respond with a return value (since we can't bubble an error up in the Layer API), but it's a nice pattern for exposing something (in this case, a file) to users while ensuring it has just one owner.

peytonr18: This is awesome, I had no idea Rust had the ability to do this! I'll do some thinking on how to maybe ensure that writer doesn't catch its own errors and loop.. maybe we have it filter to the stderr layer or something instead of to the EmitKVPLayer. I'll get an implementation up and working and see what everyone thinks.

libazureinit/src/kvp.rs
```diff
@@ -0,0 +1,649 @@
+// Copyright (c) Microsoft Corporation.
```
jeremycline, line 1: We should move both this file and the `tracing.rs` file this PR adds to the `azure-init` package instead of `libazureinit` for the moment, for two reasons.

First, I'm not sure we want to expose the layer publicly in the `libazureinit` API. If it's a generally useful tracing layer for KVP it probably makes sense to be its own crate, but I think we should give it a little bit of time to bake before we do something like that. 

Secondly, libraries should not be configuring tracing subscribers. Libraries can use `tracing` and then it's up to the consuming application how to configure the way those logs are formatted/filtered/output.

libazureinit/src/kvp.rs
```diff
@@ -0,0 +1,649 @@
...
+
+//use crate::tracing::{handle_event, handle_span};
+
+use chrono::{DateTime, Utc};
+use std::fmt;
+use uuid::Uuid;
+
+const HV_KVP_EXCHANGE_MAX_KEY_SIZE: usize = 512;
+const HV_KVP_EXCHANGE_MAX_VALUE_SIZE: usize = 2048;
+const HV_KVP_AZURE_MAX_VALUE_SIZE: usize = 1024;
```
anhvoms, line 44: This should b2 1022 and not 1024

libazureinit/src/kvp.rs
```diff
@@ -0,0 +1,649 @@
...
+
+        sleep(Duration::from_secs(1)).await;
+
+        drop(default_guard);
+
+        let contents =
+            std::fs::read(temp_path).expect("Failed to read temp file");
+        println!("Contents of the file (in bytes):\n{:?}", contents);
+
+        let slice_size = 512 + 2048;
```
anhvoms, line 574: The test should also use the constants declared above and not hard-coded number

libazureinit/src/kvp.rs
```diff
@@ -0,0 +1,649 @@
...
+
+        for i in 0..num_slices {
+            let start = i * slice_size;
+            let end = start + slice_size;
+            let slice = &contents[start..end];
+
+            println!("Processing slice {}: start={}, end={}", i, start, end);
+            println!("Slice length: {}", slice.len());
+
+            let key_section = &slice[..512];
```
anhvoms, line 595: same here - please use the declared constants

libazureinit/src/kvp.rs
```diff
@@ -0,0 +1,649 @@
...
+/// * `value` - The value associated with the key.
+fn encode_kvp_item(key: &str, value: &str) -> Vec<Vec<u8>> {
+    let key_bytes = key.as_bytes();
+    let value_bytes = value.as_bytes();
+
+    let key_len = key_bytes.len().min(HV_KVP_EXCHANGE_MAX_KEY_SIZE);
+    let mut key_buf = vec![0u8; HV_KVP_EXCHANGE_MAX_KEY_SIZE];
+    key_buf[..key_len].copy_from_slice(&key_bytes[..key_len]);
+
+    if value_bytes.len() <= HV_KVP_AZURE_MAX_VALUE_SIZE {
```
anhvoms, line 379: Is there a unit test that would validate that the slice's value's length is no more than the azure's limit?

anhvoms (changes requested):

libazureinit/src/kvp.rs
```diff
@@ -0,0 +1,649 @@
...
+    ///
+    /// # Arguments
+    /// * `_field` - A reference to the event's field metadata.
+    /// * `value` - The debug value associated with the field.
+    fn record_debug(
+        &mut self,
+        _field: &tracing::field::Field,
+        value: &dyn std_fmt::Debug,
+    ) {
+        write!(self.string, "{:?}", value).unwrap();
```
jeremycline, line 89: Two things here

### Unwrapping
This is an interesting case; according to the documentation, [writing to a string is considered infallible](https://doc.rust-lang.org/std/fmt/struct.Error.html).

There's one case where it very much can fail: we need to allocate memory to hold the string and there's no guarantee there's memory available. However, lots of Rust APIs don't expose this (Vec::with_capacity(), for example, could fail to allocate memory, but it doesn't return a `Result`).

So, this might be a case where it's okay to panic on an error, but I'd recommend using expect to note why it's okay 
```Rust
write!(self.string, "{:?}", value).expect("Writing to a string should never fail");
```

If we're really out of memory, we're likely to be killed by the OOM killer anyway.

### Formatting

This function will get called repeatedly for each field in an event. For example, say there's an event like:
```Rust
tracing::info!(request_id="some-uuid-thingy", count=3, "Some message here");
```

This will be an event with Level=Info and three fields. We'll write the debug representation of each into our string and it'll end up looking something like `Some message heresome-uuid-thingy3` depending on the order fields are processed. You could use something like the example in the tracing documentation:
```Rust
write!(self.string, "{}={:?}; ", field.name(), value).expect("Writing to a string should never fail");
```
which will give you something like `message=Some message here; request_id="some-uuid-thingy"; count=3;`.

We could also emit something like JSON if we want to make it easy to parse by field on the other side, although that adds overhead and we're space-limited here.

libazureinit/src/kvp.rs
jeremycline, line 379: You may find this if/else easier to write with the help of [chunks](https://doc.rust-lang.org/std/primitive.slice.html#method.chunks) which will iterate over your slice, yielding chunks of the requested size (or, in the case of this flavor of the method, smaller when the slice isn't long enough). That way you don't have to special-case things.

src/main.rs
```diff
@@ -77,25 +77,26 @@ fn get_username(
                 .linux_prov_conf_set
                 .username
         })
-        .ok_or(LibError::UsernameFailure.into())
+        .ok_or_else(|| {
+            event!(Level::ERROR, "Username Failure");
```
jeremycline, line 81: You can also use `tracing::error!("Username Failure")` which is equivalent.

libazureinit/src/kvp.rs
peytonr18: I didn't have a unit test for it, just the general check that it's below HV_KVP_AZURE_MAX_VALUE_SIZE, but there is a unit test now -- the last test at the end of the test block called `test_encode_kvp_item_value_length`. It creates an oversized KVP value and then, in checking that the value is too large, slices it, and then checks that each slice’s value length adheres to the Azure-defined limit.

jeremycline (changes requested):

src/main.rs
peytonr18: That's a great point! I think my intent was to ensure that this specific failure (and similar ones in `main()` and `get_environment())` was treated as a distinct event explicitly tied to the `root` span. My concern was that these failures might get overlooked within the broader context of the `root` span if they weren't explicitly recorded as events.

That said, if you feel `tracing::error` provides sufficient clarity and we don't need that level of distinction, I'm happy to update all occurrences to use `tracing::error` instead!"

libazureinit/src/kvp.rs
peytonr18: I've updated the code to use expect() with an explanation to handle the unlikely failure case when writing to a string. I've also adjusted the formatting to include field names in the output, so it now looks like name=value;, which aligns with the tracing documentation example. Thank you for the great suggestion!

JSON formatting would definitely make parsing easier, but I think it’s best to wait until we see how much space a fully-formed tracer setup takes. It may be worth it, or we may find that the name=field format is the best we can do with the space constraints. I'll make a note of this suggestion to address in the future, but for now, I'd lean towards just implementing the formatting change you recommended.

src/main.rs
jeremycline: Ah, the `tracing::error!` is a [convenience macro](https://docs.rs/tracing/latest/tracing/index.html#shorthand-macros) that creates an event like `tracing::event!` and sets the level to `Level::ERROR` automatically. It should be exactly the same thing, and there's similar convenience macros to create spans at particular levels as well.

Totally up to you since they're the same thing.

src/logging.rs
```diff
@@ -0,0 +1,57 @@
...
+    let provider = SdkTracerProvider::builder().with_config(config).build();
+
+    global::set_tracer_provider(provider.clone());
+    provider.tracer("azure-kvp")
+}
+
+pub fn setup_layers(
+    tracer: sdktrace::Tracer,
+) -> Result<(), Box<dyn std::error::Error>> {
+    let otel_layer = OpenTelemetryLayer::new(tracer);
```
jeremycline, line 29: You might want to consider applying a filter to this layer, or alternatively moving the current filter that's applied to `stderr_layer` to a [global filter](https://docs.rs/tracing-subscriber/0.3.19/tracing_subscriber/layer/index.html#filtering-with-layers) like this:

```Rust
    let subscriber = Registry::default()
        .with(stderr_layer)
        .with(otel_layer)
        .with(emit_kvp_layer)
        .with(EnvFilter::from_env("AZURE_INIT_LOG"))
```

This works because `EnvFilter` implements both the `Filter` trait and the `Layer` trait. It's important to note that a global filter will stop the other layers from even seeing the event so their filters can only filter from the events allowed through the global filter. This can be good (if you have a lot of events and layers filtering can become expensive!) but also can be confusing if you apply a global filter of say, WARN, and then try to get INFO-level events out of a particular layer with a configurable filter.

Global filters are nice for performance-sensitive applications. I once hit an issue where filtering was consuming something like 10% of the CPU cycles because of a trace-level event inside a loop that executed a _lot_ and every time a half-dozen layers had to say "oh I don't care about this event". [There are even features on tracing to allow you to compile out events](https://github.com/tokio-rs/tracing/blob/b02a700ba6850ad813f77e65144114f866074a8f/tracing/Cargo.toml#L58) for particularly performance-sensitive use-cases.

Since we don't have a ton of events or layers we can comfortably afford to have per-layer filtering, so I lean that direction (even if the OpenTelemetry layer and the stderr layer use the same `Filter`) so we can adjust the KVP filter separately.

src/main.rs
```diff
@@ -130,12 +132,29 @@ async fn main() -> ExitCode {
     }
 }
 
-#[instrument]
+#[instrument(name = "root")]
```
jeremycline, line 135: This wasn't introduced in this PR, but by default the macro will use the function parameters as fields which is a bit verbose when it comes to the Config and Cli arguments:
```
2024-12-18T21:18:32.886193Z  INFO root{config=Config { ssh: Ssh { authorized_keys_path: ".ssh/authorized_keys", query_sshd_config: true }, hostname_provisioners: HostnameProvisioners { backends: [Hostnamectl] }, user_provisioners: UserProvisioners { backends: [Useradd] }, password_provisioners: PasswordProvisioners { backends: [Passwd] }, imds: Imds { connection_timeout_secs: 2.0, read_timeout_secs: 60.0, total_retry_timeout_secs: 300.0 }, provisioning_media: ProvisioningMedia { enable: true }, azure_proxy_agent: AzureProxyAgent { enable: true }, wireserver: Wireserver { connection_timeout_secs: 2.0, read_timeout_secs: 60.0, total_retry_timeout_secs: 1200.0 }, telemetry: Telemetry { kvp_diagnostics: true } } opts=Cli { groups: [""], config: None }}: azure_init: new
2024-12-18T21:18:32.886333Z  INFO root{config=Config { ssh: Ssh { authorized_keys_path: ".ssh/authorized_keys", query_sshd_config: true }, hostname_provisioners: HostnameProvisioners { backends: [Hostnamectl] }, user_provisioners: UserProvisioners { backends: [Useradd] }, password_provisioners: PasswordProvisioners { backends: [Passwd] }, imds: Imds { connection_timeout_secs: 2.0, read_timeout_secs: 60.0, total_retry_timeout_secs: 300.0 }, provisioning_media: ProvisioningMedia { enable: true }, azure_proxy_agent: AzureProxyAgent { enable: true }, wireserver: Wireserver { connection_timeout_secs: 2.0, read_timeout_secs: 60.0, total_retry_timeout_secs: 1200.0 }, telemetry: Telemetry { kvp_diagnostics: true } } opts=Cli { groups: [""], config: None }}: azure_init: Kernel Version: 6.12.4-200.fc41.x86_64, OS Version: 41, Azure-Init Version: 0.1.1
```

You can have it skip them both with
```suggestion
#[instrument(name = "root", skip_all)]
```

src/main.rs
```diff
@@ -130,12 +132,29 @@ async fn main() -> ExitCode {
...
+    let os_version = system
+        .os_version()
+        .unwrap_or("Unknown OS Version".to_string());
+    let azure_init_version = env!("CARGO_PKG_VERSION");
+
+    tracing::info!(
+        "Kernel Version: {}, OS Version: {}, Azure-Init Version: {}",
+        kernel_version,
+        os_version,
+        azure_init_version
```
jeremycline, line 150: We capture the version in a constant so you could replace this with that:
```suggestion
        VERSION
```

libazureinit/src/config.rs
```diff
@@ -307,21 +307,27 @@ impl Config {
             if cli_path.is_dir() {
                 figment = Self::merge_toml_directory(figment, cli_path)?;
             } else {
-                tracing::info!(
```
jeremycline, line 310: In my opinion this is nice to have at info-level since configuration being unexpected included/not included is a thing that happens regularly.  This applies to all the log statements about files being included.

libazureinit/src/http.rs
```diff
@@ -111,7 +111,7 @@ async fn request(
         let now = std::time::Instant::now();
         let mut attempt =  0_u32;
         loop {
-            let span = tracing::info_span!("request", attempt, http_status = tracing::field::Empty);
+            let span = tracing::debug_span!("request", attempt, http_status = tracing::field::Empty);
```
jeremycline, line 114: I could go either way here, but I lean towards this being info-level (see my rambling comment on log levels below). I don't feel terribly strongly about it, though.

jeremycline (changes requested):
I got a little side-tracked and haven't reviewed `src/kvp.rs` itself just yet, but figured I'd stop here for today.

Looking back over my comments on the logging setup I realize I was pretty handwavy. The summarized version with clearly stated opinions would be:
- Go with layer-level filtering instead of a global filter
- Go with a (separate) EnvFilter for the KVP layer since runtime configurability is very helpful when it's your only debug option. Default to WARN or above.
- Adjust the stderr filter to default to INFO and apply it to the OTel layer as well.

Sorry I didn't make it all the way through it, I'll finish up tomorrow 😃

src/kvp.rs
```diff
@@ -0,0 +1,684 @@
...
+    .trim_end_matches('\x00')
+    .to_string();
+
+    Ok((key, value))
+}
+
+/// Truncates the guest pool KVP file if it contains stale data (i.e., data
+/// older than the system's boot time). Logs whether the file was truncated
+/// or no action was needed.
+fn truncate_guest_pool_file(kvp_file: &Path) -> Result<(), Error> {
```
jeremycline, line 438: Optional, but since this lives in the `azure-init` crate you can use the `anyhow::Error` type to avoid having to do the `map_err` stuff.

src/kvp.rs
```diff
@@ -0,0 +1,684 @@
...
+        }
+    }
+
+    Ok(())
+}
+
+/// Retrieves the system's uptime using the `sysinfo` crate, returning the duration
+/// since the system booted. This can be useful for time-based calculations or checks,
+/// such as determining whether data is stale or calculating the approximate boot time.
+fn get_uptime() -> Result<Duration, Error> {
```
jeremycline, line 474:
```suggestion
fn get_uptime() -> Duration {
```
I'm guessing this was from when you moved to `sysinfo`, but it looks like there's no calls in the function that are fallible so you can return the Duration directly.

jeremycline (changes requested):
Finished up my review on the KVP bits and I think there are just a handful of little tweaks needed there.

src/kvp.rs
```diff
@@ -0,0 +1,684 @@
...
+/// # Arguments
+/// * `key` - The key as a string slice.
+/// * `value` - The value associated with the key.
+fn encode_kvp_item(key: &str, value: &str) -> Vec<Vec<u8>> {
+    let key_bytes = key.as_bytes();
+    let value_bytes = value.as_bytes();
+
+    let key_len = key_bytes.len().min(HV_KVP_EXCHANGE_MAX_KEY_SIZE);
+    let mut key_buf = vec![0u8; HV_KVP_EXCHANGE_MAX_KEY_SIZE];
+    key_buf[..key_len].copy_from_slice(&key_bytes[..key_len]);
```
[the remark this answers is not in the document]
peytonr18: Thank you for pointing this out—great catch! I wasn’t fully considering the potential for a panic when indexing out of bounds; it’s definitely a subtle but important Rust gotcha. I'll make a note of this to avoid similar issues in the future!! 

Regarding the zero-padding, it is required due to how Hyper-V processes the telemetry. Hyper-V has a minimum length requirement for each entry, so when the entry length is below that threshold (which is often the case), we need to zero-fill to meet the requirement. I’ll make sure this is documented clearly though as part of the change!

src/kvp.rs
```diff
@@ -362,41 +352,51 @@ fn generate_event_key(
 /// exceeds the allowed size, it is split into multiple slices for encoding.
 /// This is used for logging events to a KVP file.
 ///
+/// # Note
```
anhvoms, line 355: why is there only note about the key? the value also needs to be 0-padded

doc/libazurekvp.md
```diff
@@ -0,0 +1,69 @@
+# Tracing Logic Overview
```
anhvoms, line 1: It is not clear on reading the log how to disable writing to kvp? This is a strong requirement per Azure Privacy Review (that we need to document what is collected, and provide the customer with a mean to disable it)

This document probably focuses on the technical aspect of tracing/kvp. I would add another section on the main page about telemetry and describe what we're collecting
See this section
https://learn.microsoft.com/en-us/azure/virtual-machines/linux/using-cloud-init#telemetry

anhvoms (changes requested):

doc/libazurekvp.md
peytonr18: https://github.com/Azure/azure-init/issues/152
Will address this great PR comment in this issue!

src/logging.rs
```diff
@@ -0,0 +1,60 @@
...
+
+pub fn setup_layers(
+    tracer: sdktrace::Tracer,
+) -> Result<(), Box<dyn std::error::Error>> {
+    let otel_layer = OpenTelemetryLayer::new(tracer)
+        .with_filter(EnvFilter::from_env("AZURE_INIT_LOG"));
+
+    let kvp_filter = EnvFilter::builder()
+        .with_default_directive(LevelFilter::INFO.into())
+        .parse_lossy("azure_init=INFO,libazureinit=WARN,kvp=INFO");
```
jeremycline, line 35: I'm hesitant about setting specific logs to have the `kvp` target, for a few reasons.

The target is typically the crate or module path (e.g. `tokio` or `tokio::net` would be targets in the `tokio` crate) so this will be surprising to people, particularly anyone trying to use `libazureinit` since they might believe they can set up a tracing filter like `INFO,libazureinit=WARN` so the default level is `INFO`, but the library only emits warning level or higher. However, since some logs have a target of `kvp` they'll emit at `INFO` level unless they do `INFO,libazureinit=WARN,kvp=WARN`. There's also no guarantee the library user doesn't have a `kvp` crate as a dependency with that target.

It also mixes in some of the choice about what logs get emitted to the event itself rather than the configuration here. It's definitely convenient to tag an event with "kvp", but if there comes a time when we're debugging an issue and we'd like to get more logs out via KVP without rebuilding the binary, this will be difficult.

What I'd prefer would be to mostly rely on log levels and a set of nice default directives. We can let the user provide an alternate configuration to allow for disabling or debugging but have a default that is what we want for most cases.

Here's an example that nearly matches what you've got marked with `kvp`:

```Rust
    // Use `parse()` to fail if the directives are invalid, rather than quietly proceeding with missing filters
    // Also, I did it in a list with `join()` mostly so I could add a comment for each one
    let kvp_filter = EnvFilter::builder().parse(
        [
            // Emit any events that are WARN or higher, no matter what library they originate from.
            "WARN",
            // Emit all INFO+ events from the binary
            "azure_init=INFO",
            // Emit anything INFO+ from anything inside libazure::{provision, config, http}
            "libazureinit::provision=INFO",
            "libazureinit::config=INFO",
            "libazureinit::http=INFO",
        ]
        .join(","),
    )?;
```

This filter picks up a few more events than just the `kvp` ones, but we could reasonably drop those to `DEBUG` level, I think.

Assuming the above is the default and the set of directives is pulled from the configuration file, we can get _really_ specific when we want to do some debugging:
```
    let kvp_filter = EnvFilter::builder().parse(
        [
            // Emit the logs inside the "request" span in the libazureinit::http module, but _only_
            // if the "attempt" field is 0 (e.g. it's the first try).
            "libazureinit::http[request{attempt=0}]=INFO",
            // We want to debug HTTP requests, emit trace level events from the hyper_util dependency we have
            "hyper_util=TRACE", 
        ]
        .join(","),
    )?;
```

jeremycline (commented):
I played around with the filters a bit and got pretty close to the set of logs you're marking with `target: "kvp"` (I think). I would prefer to avoid changing the target if we can help it.

I also merged #151 so the CI should be happy once you merge that in (or rebase)

src/logging.rs
jeremycline: Follow-up: I spent some time trying to create a list of filter directives to exactly match the log messages with the manually set `target: "kvp"` in the PR, but unfortunately tracing does not let you match field names on _events_, it seems to only be for span fields. This is sad, because you can't have a filter that matches on the `message` field, for example.

Thus, if we can't filter solely on the auto-generated target names and levels, we can do something like this:

```Rust
// The default generated target for this log would be `libazureinit::config`
tracing::info!(target: "libazureinit::config::success", "Configuration successfully loaded.");
```

This gives the event a unique target name so we can add `libazureinit::config::success` to the list of filter directives and it will match the expectations users will have about event target names (e.g. `libazureinit=WARN` would apply to it). The downside to this approach is:

1. Some manual work naming an event and adding it to the KVP filter
2. If the code containing the event is refactored, its target won't match its new module name.

Still, I think it's strictly less surprising than a new top-level target namespace like `kvp`.

jeremycline (approved):
Thanks for all your hard work here! ✨

PR #136, opened 2024-10-17
Updating sshd config for PasswordAuthentication

### LINKED ISSUE #67: libazureinit needs to update the sshd config when it provisions a user with password
## Description

Most images by default have sshd's config PasswordAuthentication set to no. If the user indicates that disablePasswordAuthentication should be "False", we should update this field to true by writing an additional config to /etc/ssh/sshd_config.d/

Keep in mind that if the service starts later in the process, updating this config might require a restart of ssh service

balakreddy:
Update /etc/ssh/sshd_config for PasswordAuthentication to yes and restart sshd. Fixes #67

### DIFF
--- a/libazureinit/src/provision/password.rs
+++ b/libazureinit/src/provision/password.rs
@@ -7,6 +7,8 @@ use tracing::instrument;
 
 use crate::{error::Error, User};
 
+use super::ssh::update_sshd_config;
+
 /// Available tools to set the user's password (if a password is provided).
 #[derive(strum::EnumIter, Debug, Clone)]
 #[non_exhaustive]
@@ -29,6 +31,14 @@ impl Provisioner {
 
 #[instrument(skip_all)]
 fn passwd(user: &User) -> Result<(), Error> {
+    // Update the sshd configuration to allow password authentication.
+    let ret = update_sshd_config();
+    if ret.is_err() {
+        return Err(Error::SubprocessFailed {
+            command: "update_sshd_config".to_string(),
+            status: Default::default(),
+        });
+    }
     let path_passwd = env!("PATH_PASSWD");
 
     if user.password.is_none() {
--- a/libazureinit/src/provision/ssh.rs
+++ b/libazureinit/src/provision/ssh.rs
@@ -2,13 +2,14 @@
 // Licensed under the MIT License.
 
 use std::{
-    fs::Permissions,
-    io::{self, Write},
+    fs::{self, OpenOptions, Permissions},
+    io::{self, Read, Write},
     os::unix::fs::{DirBuilderExt, PermissionsExt},
     path::PathBuf,
     process::Command,
 };
 
+use regex::Regex;
 use tracing::{info, instrument};
 
 use crate::error::Error;
@@ -67,20 +68,61 @@ pub(crate) fn get_authorized_keys_path_from_sshd() -> io::Result<Option<PathBuf>
     Ok(None)
 }
 
+pub(crate) fn update_sshd_config() -> io::Result<()> {
+    let sshd_config_path = "/etc/ssh/sshd_config";
+    let temp_sshd_config_path = "/etc/ssh/sshd_config.tmp";
+
+    let mut file_content = String::new();
+    {
+        let mut file = OpenOptions::new().read(true).open(sshd_config_path)?;
+        file.read_to_string(&mut file_content)?;
+    }
+
+    let re =
+        Regex::new(r"(?m)^\s*#?\s*PasswordAuthentication\s+no\s*$").unwrap();
+    let modified_content =
+        re.replace_all(&file_content, "PasswordAuthentication yes");
+
+    if modified_content == file_content {
+        return Ok(());
+    }
+
+    let mut temp_file = OpenOptions::new()
+        .write(true)
+        .create(true)
+        .truncate(true)
+        .open(temp_sshd_config_path)?;
+    temp_file.write_all(modified_content.as_bytes())?;
+    temp_file.set_permissions(fs::Permissions::from_mode(0o644))?;
+
+    fs::rename(temp_sshd_config_path, sshd_config_path)?;
+
+    // Restart the sshd service to apply changes
+    Command::new("systemctl")
+        .arg("restart")
+        .arg("sshd")
+        .output()?;
+
+    Ok(())
+}
+
 #[cfg(test)]
 mod tests {
 
     use nix::unistd::User;
+    use regex::Regex;
     use tracing::info;
 
     use crate::imds::PublicKeys;
+    use crate::provision::ssh::OpenOptions;
     use std::{
-        fs::Permissions,
+        fs::{self, File, Permissions},
         io::{self, Read, Write},
         os::unix::fs::{DirBuilderExt, PermissionsExt},
         path::PathBuf,
         process::Command,
     };
+    use tempfile::TempDir;
 
     fn mock_sshd_output(output: &str) -> io::Result<Option<PathBuf>> {
         let output = Command::new("echo").arg(output).output()?;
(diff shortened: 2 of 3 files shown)


libazureinit/src/provision/password.rs
```diff
@@ -29,6 +31,14 @@ impl Provisioner {
...
 #[instrument(skip_all)]
 fn passwd(user: &User) -> Result<(), Error> {
+    // Update the sshd configuration to allow password authentication.
+    let ret = update_sshd_config();
+    if ret.is_err() {
+        return Err(Error::SubprocessFailed {
+            command: "update_sshd_config".to_string(),
+            status: Default::default(),
+        });
+    }
```
dongsupark, line 41: As far as I understand, `SubprocessFailed` is supposed to be used for dealing with errors when running a separate command-line tool, like `hostname`, `ls` etc.
In this case, since a command-line tool `update_sshd_config` does not exist, I would rather recommend adding a new error type, e.g. `Error::UpdateSshdConfig`.

libazureinit/src/provision/ssh.rs
```diff
@@ -67,20 +68,61 @@ pub(crate) fn get_authorized_keys_path_from_sshd() -> io::Result<Option<PathBuf>
...
+    let temp_sshd_config_path = "/etc/ssh/sshd_config.tmp";
+
+    let mut file_content = String::new();
+    {
+        let mut file = OpenOptions::new().read(true).open(sshd_config_path)?;
+        file.read_to_string(&mut file_content)?;
+    }
+
+    let re =
+        Regex::new(r"(?m)^\s*#?\s*PasswordAuthentication\s+no\s*$").unwrap();
```
dongsupark, line 82: Not sure if a plain `unwrap` is a good idea here in context of the main execution path.
But I see that replacing unwrap with `?` results in another build failures. :thinking:

dongsupark (commented):
Thanks for creating the PR.

libazureinit/src/provision/ssh.rs
jeremycline: So, the regex call can fail for two reasons. The first is if the pattern is invalid, and secondly if the pattern would produce a regular expression above the configured size limit (and the default is said to be "reasonable").

Given that, as long as we have a test that builds this regex, we can feel okay about unwrapping the error, although I think instead of `unwrap()` we should use `expect("The regular expression is invalid or exceeds the default regex size")`.

The second thing is that regular expression compilation is relatively expensive, so the typical pattern when building a regex from a static string like this is to make it static so it's compiled only once. Now, we're only going to ever call this function once, so in practice it makes no difference from a performance perspective, _but_ if we do that pattern it becomes very easy to write a test specifically for the potential errors we're unwrapping.

The pattern looks like this with the standard library's [LazyLock](https://doc.rust-lang.org/std/sync/struct.LazyLock.html):
```rust
use std::sync::LazyLock;

static PASSWORD_REGEX: LazyLock<Regex> = LazyLock::new(|| {
    Regex::new(r"(?m)^\s*#?\s*PasswordAuthentication\s+no\s*$").unwrap()
}); 
```

Unfortunately, LazyLock is new in Rust 1.80 so we will have to use the `lazy_static` crate instead until our Minimum Supported Rust Version is 1.80+. Doing it with lazy_static is [very similar](https://docs.rs/lazy_static/latest/lazy_static/#syntax).

libazureinit/src/provision/ssh.rs
```diff
@@ -67,20 +68,61 @@ pub(crate) fn get_authorized_keys_path_from_sshd() -> io::Result<Option<PathBuf>
     Ok(None)
 }
 
+pub(crate) fn update_sshd_config() -> io::Result<()> {
+    let sshd_config_path = "/etc/ssh/sshd_config";
+    let temp_sshd_config_path = "/etc/ssh/sshd_config.tmp";
```
jeremycline, line 73: Rather than hard-coding this, let's provide the `sshd_config_path` as a parameter. This makes it easy to write a test for it, and it also means we can potentially configure this (via the work in #137)

Also, we could probably also create the temporary file in a directory made by [tempfile](https://docs.rs/tempfile/latest/tempfile/struct.TempDir.html). This way if something goes wrong, we don't leave a file in /etc/ssh/.

libazureinit/src/provision/ssh.rs
```diff
@@ -67,20 +68,61 @@ pub(crate) fn get_authorized_keys_path_from_sshd() -> io::Result<Option<PathBuf>
...
+        .create(true)
+        .truncate(true)
+        .open(temp_sshd_config_path)?;
+    temp_file.write_all(modified_content.as_bytes())?;
+    temp_file.set_permissions(fs::Permissions::from_mode(0o644))?;
+
+    fs::rename(temp_sshd_config_path, sshd_config_path)?;
+
+    // Restart the sshd service to apply changes
+    Command::new("systemctl")
```
jeremycline, line 101: This introduces the assumption that `systemd` is the service manager, which we probably don't want to do. We'll want to add a service manager configuration option and use whatever is configured to reload sshd.

Alternatively, we can cause sshd to reload its configuration by sending it a `SIGHUP` signal (documented in its manual page, see https://www.man7.org/linux/man-pages/man8/sshd.8.html). However, that requires that we know its PID which might be a rather tricky endeavor to do safely across all distributions... cloud-init presumably has decided a way to handle this so maybe learning from them would be best.

libazureinit/src/provision/ssh.rs
```diff
@@ -101,10 +143,38 @@ mod tests {
                 }
             }
         }
-
         Ok(None)
     }
 
+    fn mock_update_sshd_config(sshd_config_path: &PathBuf) -> io::Result<()> {
```
jeremycline, line 149: With the above suggestion about providing the path to the function, you can drop this mock and write your tests against the real function

jeremycline (changes requested):

libazureinit/src/provision/ssh.rs
cjp256: We should limit support for this to a drop-in config:
/etc/ssh/sshd_config.d/50-azure-init.conf

We shouldn't be tampering the main config if we can avoid it.

libazureinit/src/provision/ssh.rs
cjp256: cloud-init's approach:

https://github.com/canonical/cloud-init/blob/main/cloudinit/config/cc_set_passwords.py#L60

I think we'd be best to avoid it altogether with dependency ordering.

libazureinit/src/provision/ssh.rs
balakreddy: Done

libazureinit/src/provision/ssh.rs
balakreddy: Using the suggested /etc/ssh/sshd_config.d/50-azure-init.conf.

libazureinit/src/provision/ssh.rs
balakreddy: Removed this section

libazureinit/src/provision/ssh.rs
balakreddy: Done

libazureinit/src/provision/ssh.rs
```diff
@@ -67,20 +76,65 @@ pub(crate) fn get_authorized_keys_path_from_sshd() -> io::Result<Option<PathBuf>
...
 }
 
+pub(crate) fn update_sshd_config(sshd_config_path: &str) -> Result<(), io::Error> {
+    // Check if the path exists otherwise create it
+    let sshd_config_path = PathBuf::from(sshd_config_path);
+    if !sshd_config_path.exists() {
+        let mut file = std::fs::File::create(&sshd_config_path)?;
+        file.set_permissions(Permissions::from_mode(0o644))?;
+        file.write_all(b"PasswordAuthentication yes\n")?;
+        file.flush()?;
```
dongsupark, line 86: I am curious why we call `flush()` after writing a file, because `File::flush()` for Linux is actually [no-op](https://github.com/rust-lang/rust/blob/27e38f8fc7efc57b75e9a763d7a0ee44822cd5f7/library/std/src/sys/pal/unix/fs.rs#L1310).
If we really want to ensure files to be written to underlying devices, `File::sync_all` or `File::sync_data` could be actually valid, but that would result in performance impact.

libazureinit/src/provision/ssh.rs
```diff
@@ -67,20 +76,65 @@ pub(crate) fn get_authorized_keys_path_from_sshd() -> io::Result<Option<PathBuf>
...
+            .open(temp_sshd_config_path)?;
+        temp_file.write_all(modified_content.as_bytes())?;
+        temp_file.set_permissions(fs::Permissions::from_mode(0o644))?;
+
+        fs::rename(temp_sshd_config_path, &sshd_config_path)?;
+    } else {
+        let mut file =
+            OpenOptions::new().append(true).open(&sshd_config_path)?;
+        file.write_all(b"PasswordAuthentication yes\n")?;
+        file.flush()?;
```
dongsupark, line 117: same here

libazureinit/src/provision/ssh.rs
```diff
@@ -2,17 +2,26 @@
...
 use tracing::{info, instrument};
 
 use crate::error::Error;
 use crate::imds::PublicKeys;
+use lazy_static::lazy_static;
+
+lazy_static! {
+    static ref PASSWORD_REGEX: Regex =
+        Regex::new(r"(?m)^\s*#?\s*PasswordAuthentication\s+(yes|no)\s*$")
+            .unwrap();
```
dongsupark, line 23:
```suggestion
            .expect("The regular expression is invalid or exceeds the default regex size");
```
as @jeremycline suggested.

dongsupark (commented):
Thanks, the error type issue seems now resolved.

libazureinit/src/provision/ssh.rs
balakreddy: I think we can avoid using flush and sync as well.

dongsupark:
Thanks, looks good to me.
Can you please rebase?

dongsupark (approved):

dongsupark:
Still needs an approval from @jeremycline

anhvoms (approved):
LGTM. Thanks

jeremycline (approved):
✨

PR #137, opened 2024-10-21
Implement custom configuration and SSH key path flexibility

peytonr18:
Config Struct PR Changes and Notes

1. New Configuration Hierarchy and --config Flag Support
  - Purpose: Ensure all configuration sources are processed in the correct order: defaults, base file (azure-init.toml), directory files, and CLI-specified paths.
  - Changes:
      - main.rs:
         - Introduced CLI parsing for --config, allowing users to specify a file or directory.
         - If provided, the specified path is passed into the Config struct’s load function.
         - The --config parameter is parsed within provision() and no longer directly set in main(), maintaining separation of concerns.
      - config.rs:
         - `Config::load()`:
             - Starts with defaults (`Config::default()`).
             - Explicitly checks if `azure-init.toml` exists before merging.
             - Introduced `merge_toml_directory` helper function to handle merging `.toml` files from directories, ensuring logic is reused for both `azure-init.toml.d` and CLI directory paths.
             - Processes CLI-specified paths last, merging them into the final configuration.

2. New Enum-Based Provisioners for Hostname, User, and Password
 - Purpose : Streamlines the selection of system tools for setting hostname, user, and password, enabling test mocks and extensibility.
 - Changes:
    - config.rs: Added enums `HostnameProvisioner`, `UserProvisioner`, and `PasswordProvisioner` with default implementations for each.
    - Test Variants: Includes FakeHostnamectl, FakeUseradd, and FakePasswd variants under #[cfg(test)] for unit testing.

3. Updated Provision Struct in mod.rs to Accept Config Object
- Purpose: Integrates the Config struct into provisioning, enabling flexible overrides and default handling.
- Changes:
   - Provision now accepts a `Config` object on initialization, replacing previous hard-coded values.
   - Provisioning Flow: Methods `hostname_provisioners`, `user_provisioners`, and `password_provisioners` were removed from main.rs and are now part of Config. Provisioning logic directly references these config-driven settings.
  
4. /doc
- The configuration.md file inside /doc that acts like a high level design doc that can also serve as user documentation.


libazureinit/src/provision/config.rs
```diff
@@ -0,0 +1,167 @@
...
+    pub strict_validation: bool,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct TelemetryConfig {
+    pub kvp_diagnostics: bool,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Config {
```
jeremycline, line 57: If you put `#[serde(default)]` (see https://serde.rs/container-attrs.html) on this struct, any missing fields will be filled in by the implementation of `Default` so users don't need to fill in a complete config file if they want to just change one section setting.

jeremycline (changes requested):
I haven't quite finished reviewing things, but it's getting late-ish for me so I'll stop here and come back tomorrow. Mostly just organizational things and tricks with defaults and serde, looks like a really solid start!

libazureinit/src/provision/config.rs
```diff
@@ -0,0 +1,167 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+use crate::error::Error;
+use serde::{Deserialize, Serialize};
+use std::fs;
+use toml;
+use tracing;
+
+#[derive(Serialize, Deserialize, Debug)]
```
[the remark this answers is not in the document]
peytonr18: Oh -- this is super cool. With this, we could eliminate the general azure-init.conf file and just merge in whatever we wanted to specify via 10-azure-init-ssh.conf. Is that what you were thinking?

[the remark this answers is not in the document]
jeremycline: Basically, although I would not hard code a path. Instead, we can add a parameter to the command line so a user can provide `--config` or set the `AZURE_INIT_CONFIG` environment variable to a path where the configuration lives. Having a default of `/etc/azure-init.conf` or `/etc/azure-init/cli.conf` or something would be okay, but I don't think it's really necessary since we're not dictating where the config lives anywhere else, and we can provide reasonable defaults.

The distribution can install the config wherever it wants and configure the systemd unit (or whatever init system they use) to include an environment variable pointing to the config.

libazureinit/src/provision/config.rs
```diff
@@ -0,0 +1,167 @@
+// Copyright (c) Microsoft Corporation.
```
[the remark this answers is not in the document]
peytonr18: I am totally good w this but just to clarify, do you mean putting it next to main.rs in the primary azure-init/src folder? I could be totally wrong but I think if we do this, we have to have a lib.rs file that has both listed as modules or add mod config.rs inside of main.rs. 

I also think I may need to do a little research on how to import the config.rs mod into the libazureinit library. It's tricky because I think the library crate can't automatically use things from the main crate since libraries are supposed to be reusable without assuming a specific main binary.

CUSTOMCONFIG.md
anhvoms, line 1: I would rephrase this document to a high level design that can serve as a user documentation. The doc should avoid some implementation detail such as Serde validation will fail. You want to describe the behavior of azure-init when given a bad config, you don't want to describe how that behavior happens (that conversation can happen at the PR conversation level). Some sections such as pros/cons don't have to be listed in this doc but rather at the PR summary/description.

This doc should also get moved to a folder called doc, and named configuration.md

CUSTOMCONFIG.md
```diff
@@ -0,0 +1,402 @@
...
+#### 1. CLI Override (`--config`)
+   - The `--config` flag specifies the path to load the configuration from, which can be either a single file or a directory.
+     - **File:** If the path points to a file, only that file is loaded.
+     - **Directory:** If the path points to a directory, a series of `.toml` files in the specified directory will be loaded and merged based on specific rules.
+   - **Example:** `azure-init --config /path/to/custom-config.toml`
+
+#### 2. Directory Loading Logic
+   - If the `--config` parameter points to a directory, `azure-init` follows this hierarchy:
+     - First, it looks for a base file named `azure-init.toml` in the directory.
+     - Then, it merges any additional `.toml` files in a `.d` subdirectory within the specified directory, in alphabetical order.
```
anhvoms, line 44:
```suggestion
     - Then, it merges any additional `.toml` files in a `.d` subdirectory within the specified directory, in lexicographical order.
```

CUSTOMCONFIG.md
```diff
@@ -0,0 +1,402 @@
...
+
+**Order of Loading:**
+
+1. Loads configuration directly from defaults specified in `config.rs`.
+
+## Validation Behaviors
+
+Custom enum types for `NetworkManager` and `SshAuthorizedKeysPathQueryMode` validate that config settings set by the user match the supported options. If a user enters an unsupported config value, deserialization will fail because Serde will not be able to map that value to one of the enum variants.
+
+```rust
```
anhvoms, line 105: This shouldn't be as part of the documentation, unless this is what we're expecting the users to put into the config (which I hope not). It should just list an example of what the user should put in the config if they want to use `NetworkManager`, for example

CUSTOMCONFIG.md
```diff
@@ -0,0 +1,402 @@
...
+}
+```
+
+#### Example of the expected error caused by entering an unsupported config value:
+
+```javascript
+Error: Invalid value 'unsupported_value' for NetworkManager. Expected 'NetworkManager' or 'systemd-networkd'.
+```
+
+The enum also relies on the `#[serde(rename = "disabled")]` attribute, which maps the Rust enum variant `Disabled` to the string "disabled."  This allows clear internal code naming while matching formats set by users in their configuration.
```
anhvoms, line 130: This doesn't need to be part of the doc

CUSTOMCONFIG.md
```diff
@@ -0,0 +1,402 @@
...
+
+- **More Objects to Manage**: Users will need to manage both the `Config` and `Provision` objects. This might add some cognitive overhead, as they need to understand how to construct both.
+
+- **API Complexity**: This design might be more complex for users who are accustomed to a simple builder-style API. Instead of calling individual methods for each configuration option, they must work with the `Config` object, which is an API-breaking change.
+
+### Summary
+
+This approach centralizes configuration in a `Config` struct and passes it into `Provision`. It eliminates the need for a method-per-configuration approach and reduces duplication, providing a single source of truth for the configuration. Given the flexibility and consistency this provides, I believe this is the best approach for the long-term maintainability of the `azure-init` project.
+
+## Config Struct PR Changes and Notes - as of 10/25/2024
```
anhvoms, line 310: We should have an example config with all the possible settings which represent azure-init out-of-box default behavior, or document that the azure-init.toml represents all the settings at their default value

config/azure-init.conf
anhvoms, line 1: Is it .conf or .toml? In the doc above you're proposing the usage of toml but the file is .conf

CUSTOMCONFIG.md
```diff
@@ -0,0 +1,402 @@
+# Custom Configuration Design for azure-init
+
+## Objective
+The azure-init custom configuration architecture enables dynamic and flexible management of various settings for virtual machines provisioned with the light-weight agent. Customizable settings include SSH, networking, IMDS, provisioning media, azure proxy agent, wire server, error handling, and telemetry. This flexible design ensures that users can adapt configurations to their specific needs.
```
anhvoms, line 4: it should be wireserver and not "wire server"

CUSTOMCONFIG.md
```diff
@@ -0,0 +1,402 @@
...
+```javascript
+Error: Invalid value 'unsupported_value' for NetworkManager. Expected 'NetworkManager' or 'systemd-networkd'.
+```
+
+The enum also relies on the `#[serde(rename = "disabled")]` attribute, which maps the Rust enum variant `Disabled` to the string "disabled."  This allows clear internal code naming while matching formats set by users in their configuration.
+
+### Configuration Fields
+
+#### Network Struct:
+- **manage_configuration**: `bool`
```
anhvoms, line 135: are we managing network config? I don't think we currently have any network management code. I would not put it in and amend the documentation when we start influence networking behavior

config/azure-init.conf
```diff
@@ -0,0 +1,33 @@
...
+[azure_proxy_agent]
+enable = true
+
+[wireserver]
+connection_timeout_secs = 2.0
+read_timeout_secs = 60
+retry_timeout_secs = 1200
+
+[errors]
+strict_validation = true
```
anhvoms, line 30: this is not documented, what does it do?

libazureinit/src/config.rs
```diff
@@ -0,0 +1,282 @@
...
+    fn default() -> Self {
+        Self {
+            backends: vec![PasswordProvisioner::default()],
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct Imds {
+    pub connection_timeout_secs: f64,
```
anhvoms, line 128: why is connection timeout a float but read timeout unsigned int? They probably should all be float so that we can specify fraction of seconds if we want to

libazureinit/src/config.rs
```diff
@@ -0,0 +1,282 @@
...
+
+impl Default for AzureProxyAgent {
+    fn default() -> Self {
+        Self { enable: true }
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct Wireserver {
+    pub connection_timeout_secs: f64,
```
anhvoms, line 167: same comment with imds timeout above

config/azure-init.conf
```diff
@@ -0,0 +1,33 @@
...
+
+[ssh]
+authorized_keys_path = "~/.ssh/authorized_keys"
+configure_password_authentication = false
+authorized_keys_path_query_mode = "disabled"
+
+[imds]
+connection_timeout_secs = 2.0
+read_timeout_secs = 60
+retry_timeout_secs = 300
```
anhvoms, line 16:
```suggestion
total_retry_timeout_secs = 300
```

config/azure-init.conf
```diff
@@ -0,0 +1,33 @@
...
+[provisioning_media]
+enable = true
+
+[azure_proxy_agent]
+enable = true
+
+[wireserver]
+connection_timeout_secs = 2.0
+read_timeout_secs = 60
+retry_timeout_secs = 1200
```
anhvoms, line 27:
```suggestion
total_retry_timeout_secs = 1200
```

libazureinit/src/provision/mod.rs
```diff
@@ -5,9 +5,14 @@ pub mod password;
 pub(crate) mod ssh;
 pub mod user;
 
-use strum::IntoEnumIterator;
+//use nix::libc::USER_PROCESS;
```
anhvoms, line 8: remove commented code

libazureinit/src/provision/ssh.rs
```diff
@@ -1,70 +1,84 @@
...
     keys: &[PublicKeys],
+    authorized_keys_path: Option<PathBuf>,
+    query_mode: SshAuthorizedKeysPathQueryMode,
 ) -> Result<(), Error> {
+    let authorized_keys_path = match query_mode {
+        SshAuthorizedKeysPathQueryMode::SshdG => {
+            tracing::info!("Attempting to get authorized keys path via sshd -G as configured.");
+            let sshd_output = Command::new("sshd").arg("-G").output()?;
+            if !sshd_output.status.success() {
+                return Err(Error::Io(std::io::Error::new(
```
anhvoms, line 28: does this get logged to tracing/console so that we can troubleshoot?

libazureinit/src/provision/ssh.rs
```diff
@@ -1,70 +1,84 @@
...
+                        let keypath: Vec<&str> =
+                            line.split_whitespace().collect();
+                        keypath.get(1).map(|path| user.dir.join(path))
+                    } else {
+                        None
+                    }
+                })
+                .unwrap_or_else(|| {
+                    user.dir.join(authorized_keys_path.as_deref().unwrap_or(
+                        PathBuf::from("~/.ssh/authorized_keys").as_path(),
```
anhvoms, line 52: If I'm reading the code correctly, if there's no authorized_keys_path found in sshd -G, we will use the default `~/.ssh/authorized_keys`
1) The default should be coming from our config file, not hardcoded here
2) sshd -G is supposed to list all options, it's not possible to not authorized_keys_path listed. If we get to the `or` part of the `unwrap`, something is horribly wrong and even if we decide to pick a default, it should be logged (here we're silently falling to a default without logging). 
3) The code in this section is also inconsistent in its behavior. A few lines above you're failing without falling back to a default if `sshd -G` fails. Here there's a default. I would suggest maintaining a strict behavior and fail here without a default (and log it clearly so that it's easy to troubleshoot)

anhvoms (changes requested):

config/azure-init.conf
peytonr18: It should be toml -- those are going to be deleted, I meant to do that before the PR. Now that we set defaults in congi.rs, we don't need either of those files outside of testing functionality.

CUSTOMCONFIG.md
peytonr18: It was in the list of potential flags Chris and I had discussed but you're right -- we don't have any network management code at the moment so I will take it out until we do!

libazureinit/src/config.rs
```diff
@@ -0,0 +1,257 @@
...
+    pub password_provisioners: PasswordProvisioners,
+    pub imds: Imds,
+    pub provisioning_media: ProvisioningMedia,
+    pub azure_proxy_agent: AzureProxyAgent,
+    pub wireserver: Wireserver,
+    pub telemetry: Telemetry,
+}
+
+impl Config {
+    pub fn load(cli_overrides: Option<PathBuf>) -> Result<Config, Error> {
```
vmarcella, line 187: NIT: I would just name `cli_overrides` -> `path`

doc/configuration.md
```diff
@@ -0,0 +1,282 @@
...
+## Configuration Fields
+
+#### Ssh Struct:
+- **authorized_keys_path**: `PathBuf`
+  - **Default**: `~/.ssh/authorized_keys`
+  - **Description**: Specifies the file path to the SSH authorized keys.
+- **configure_password_authentication**: `bool`
+  - **Default**: `false`
+  - **Description**: Controls whether password authentication is configured for SSH.
+- **authorized_keys_path_query_mode**: `SshAuthorizedKeysPathQueryMode`
```
anhvoms, line 115: I think we only have two behaviors: querying using `sshd -G`, or not querying it. In that case this one should be a boolean?

cjp256: Agreed, I would suggest keeping the proposed authorized_keys_path and dropping the fallback_sshd_default option, e.g.

[sshd]
query_sshd_config: bool = true
authorized_keys_path: string = .ssh/authorized_keys

the scope for querying the sshd_config can be useful beyond authorizedkeys, e.g. PasswordAuthentication handling so I'd keep the name generic.

- if query_sshd_config = true and sshd -G succeeds, authorizedkeys is determined dynamically from sshd.

- if query_sshd_config = true and sshd -G fails, failure is reported but VM will still be configured according to authorized_keys_path, which likely is the default .ssh/authorized_keys (we still want to configure the VM as much as possible even if reporting failure)

- if query_sshd_config = false, authorized_keys_path is used as-is.

libazureinit/src/provision/mod.rs
```diff
@@ -23,156 +24,116 @@ use crate::User;
...
-                            backend=?backend,
-                            resource="password",
-                            "Provisioning did not succeed"
-                        );
-                        e
-                    })
-                    .ok()
+            .find_map(|backend| match backend {
+                PasswordProvisioner::Passwd => {
+                    password::Provisioner::Passwd.set(&self.user).ok()
```
vmarcella, line 77: Do we want to panic here if the Password cannot be set?

libazureinit/src/provision/mod.rs
```diff
@@ -23,156 +24,116 @@ use crate::User;
...
+            })
+            .ok_or(Error::NoHostnameProvisioner)?;
+
+        self.config
+            .user_provisioners
+            .backends
+            .iter()
+            .find_map(|backend| match backend {
+                UserProvisioner::Useradd => {
+                    user::Provisioner::Useradd.create(&self.user).ok()
```
vmarcella, line 64: Do we want to panic here if the user cannot be created?

tests/functional_tests.rs
```diff
@@ -98,12 +100,13 @@ async fn main() {
         },
     ];
 
-    Provision::new("my-hostname".to_string(), User::new(username, keys))
-        .hostname_provisioners([HostnameProvisioner::Hostnamectl])
-        .user_provisioners([UserProvisioner::Useradd])
-        .password_provisioners([PasswordProvisioner::Passwd])
-        .provision()
-        .expect("Failed to provision host");
+    Provision::new(
```
vmarcella, line 103: So the builder API is completely deprecated now that we have a configuration file?

libazureinit/src/provision/ssh.rs
```diff
@@ -277,7 +287,9 @@ mod tests {
             },
         ];
 
-        provision_ssh(&user, &keys, Some(".ssh/xauthorized_keys".to_string()))
+        let authorized_keys_path = user.dir.join(".ssh/xauthorized_keys");
```
vmarcella, line 290: Is the file name supposed to be `xauthorized_keys` instead of `authorized_keys`?

vmarcella: Ah, I see that this is in a test, never mind.

libazureinit/src/provision/ssh.rs
```diff
@@ -340,18 +354,13 @@ mod tests {
...
-        provision_ssh(
-            &user,
-            &keys[1..],
-            Some(".ssh/xauthorized_keys".to_string()),
-        )
-        .unwrap();
+        let authorized_keys_path = user.dir.join(".ssh/xauthorized_keys");
+
+        provision_ssh(&user, &keys[1..], authorized_keys_path.clone(), false)
+            .unwrap();
```
vmarcella, line 360: How come we unwrap here? Do we not want to assert anything about the state of what's returned from `provision_ssh`?

vmarcella (commented):
Left a few comments with some questions that I had, but overall it looks good to me. I'll review again once the comments are addressed

libazureinit/src/provision/mod.rs
peytonr18: I would say the same here as I did above -- was trying to mirror the original behavior and maintain predictability in behavior.  But, if people think panicking might make the error more immediately obvious, then can panic instead of returning ok_or(Error::NoUserProvisioner)!

tests/functional_tests.rs
peytonr18: I wouldn't say it's entirely deprecated, but you're right in that the builder API is not essential now for things like hostname, username, and password settings. We could still find a way to use it for runtime-specific overrides or maybe testing scenarios if we wanted. 

I will also note that losing some of the builder API’s flexibility was the biggest drawback of the design change to add a Config struct into Provision, but I chose this approach because it keeps things consistent and allows us to easily override values in the config when adjustments are needed. It also centralizes everything, so we don’t end up with two separate Config structs.

libazureinit/src/provision/ssh.rs
peytonr18: I'm pretty sure we unwrap here to ensure provision_ssh succeeds without adding extra assertions, since the main focus of the test is verifying the content of .ssh/xauthorized_keys. Because provision_ssh should always succeed in this setup, unwrapping simplifies the test. 

Across the board, we do unwrap in all the tests because they're testing specific scenarios with the file content and unwrapping helps us quickly spot any issues with that.

With that all being said though, I don't necessarily know that we can gleam a ton of information about what happened during provision_ssh by looking only at what it returns. Given that provision_ssh only returns Ok(()) or an error, there’s limited value in adding assertions solely for its return state... I think we care more about the directories being set up correctly / whether it correctly sets permissions, writes keys, or falls back to a default path. That's what those tests are inherently checking.

doc/configuration.md
```diff
@@ -0,0 +1,288 @@
...
+   - **Example:** `azure-init --config /path/to/custom-config.toml`
+
+#### 2. Directory Loading Logic
+   - If the `--config` parameter points to a directory, `azure-init` follows this hierarchy:
+     - First, it looks for a base file named `azure-init.toml` in the directory.
+     - Then, it merges any additional `.toml` files in a `.d` subdirectory within the specified directory, in lexicographical order.
+     ```text
+     /etc/azure-init/
+     ├── azure-init.toml            # Base configuration
+     └── .d/
```
cjp256, line 44: this directory would normally include the config file name prefixing .d

```suggestion
     └── azure-init.toml.d/
```

doc/configuration.md
```diff
@@ -0,0 +1,288 @@
...
+
+### Example 1: Single Configuration File
+**Command:**
+```sh
+azure-init --config /path/to/custom-config.toml
+```
+**Order of Merging:**
+1. Loads `/path/to/custom-config.toml`.
+2. Applies any CLI overrides if present.
+3. Fills in missing values with defaults from `config.rs`.
```
cjp256, line 61: based on the description, I expect the order of precedence (lowest to highest) to be:

(1) compile-time defaults
(2a) custom configuration (e.g. azure-init.toml)
(2b) custom configurations found in azure-init.toml.d, in lexicographical order of file names
(3) cli options

I think maybe it would help to clarify it as such.  "fills in missing values" feels a bit mismatched in terms of "merging". If it were an order of operations for merging, the defaults should like at the top and future merges overwrite it.

doc/configuration.md
```diff
@@ -0,0 +1,288 @@
...
+## Validation Behaviors
+
+Custom enum types for fields validate that config settings set by the user match the supported options. If a user enters an unsupported config value, 
+deserialization will fail because Serde will not be able to map that value to one of the enum variants.
+
+### Example: 
+```toml
+# Configure the query_sshd_config field.
+[ssh]
+query_sshd_config = false
```
cjp256, line 104: is this an unsupported value?  it's not clear from the example

cjp256: regardless, elaborate on what will happen on validation failure

e.g. service will fail with non-zero exit code, and failure will be reported to Azure platform

doc/configuration.md
```diff
@@ -0,0 +1,312 @@
...
+
+```toml
+# Invalid value for query_sshd_config (not a boolean)
+[ssh]
+query_sshd_config = "not_a_boolean" # This will cause a validation failure
+```
+
+
+## Configuration Fields
+
```
dongsupark, line 132: Thanks for your hard work.

Just curious, why do we have here in markdown documents such a huge list of configuration fields with every detail, including default value, etc?
Those are helpful, but in the long term not so maintainable. Think about, for example, a case of changing a default value from `true` to `false`. Then you would need to touch both code and the separate doc. 
I would prefer having such details in individual rustdocs of each struct, enum, etc. in actual `.rs` files.

peytonr18: Originally, it was included as part of an ask by another dev to outline how Confg works and how  I chose the architecture I did for the config. There should be rustdocs for each of the structs that detail the default and the use for each struct in config.rs, secondary to this. But if you think this is confusing to have / maintain as well, I can always remove it!

src/main.rs
```diff
@@ -93,7 +95,12 @@ async fn main() -> ExitCode {
         a library is mis-using the tracing API.",
     );
 
-    match provision().await {
+    let opts = Cli::parse();
+
+    let config = Config::load(opts.config.clone())
+        .expect("Failed to load configuration");
```
jeremycline, line 101: Instead of panicking here, we could print out the error and an example correct config:

```Rust
    let config = Config::load(opts.config.clone())
        .map_err(|error| {
            eprintln!("Failed to load configuration ({error:?})")
        });
```
And then return `ExitCode::FAILURE` if load return an `Err`. This way the user gets a non-zero exit code (which they'd also get with a panic) but don't get a potentially confusing backtrace. Incorrect configuration is a "normal" thing, and we don't want to crash ungracefully unless we're in a state we didn't expect.

Since we don't know what log level the user might have configured, we print directly to standard error. You may also need to implement `Display` for `Config` to make the above work, something like (totally untested):
```
impl Display for Config {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        write!(
            f,
            "{}",
            toml::ser::to_string_pretty(&self).unwrap_or_default()
        )
    }
}
```

jeremycline (commented):
The only thing that worries me here is the merging bit, everything else looks solid.

libazureinit/src/config.rs
```diff
@@ -0,0 +1,681 @@
...
+
+        Ok(config)
+    }
+
+    /// Merges two configurations, giving priority to values from `other`.
+    ///
+    /// This method combines two configurations, with each field in `other` overwriting the
+    /// corresponding field in `self`. This allows for merging multiple configurations in
+    /// order of precedence, for example, applying CLI-specified configurations over defaults.
+    fn merge(mut self, other: Config) -> Config {
```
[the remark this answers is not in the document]
peytonr18: Oh, I see what you're saying here. I could see there being value in making merge happen at the individual key level! I think the change would just be merging individual keys within each section rather than one large merge the way it currently is. 


Something like this would then be in the struct for each field: 
```rust
pub struct Ssh {
    pub key_a: Option<String>,
    pub key_b: Option<String>,
    pub key_c: Option<String>,
}

impl Ssh {
    pub fn merge(mut self, other: Ssh) -> Ssh {
        if let Some(key_a) = other.key_a {
            self.key_a = Some(key_a);
        }
        if let Some(key_b) = other.key_b {
            self.key_b = Some(key_b);
        }
        if let Some(key_c) = other.key_c {
            self.key_c = Some(key_c);
        }
        self
    }
}
```
I don't think that would be a tough change, if it's one that would make more sense for the user. 
@cjp256 and @anhvoms, any thoughts here on section level merge vs. individual key merge?

doc/configuration.md
```diff
@@ -0,0 +1,312 @@
...
+  - **Variants**:
+    - **FakeUseradd**: Testing provisioner that simulates `useradd`.
+
+#### PasswordProvisioners Struct
+- **backends**: `PasswordProvisioner`
+  - **Default**: `Passwd`
+  - **Description**: Specifies the tool used to set user passwords.
+  - **Variants**:
+    - **FakePasswd**: Testing provisioner that simulates `passwd`.
+#### Imds Struct:
```
cjp256, line 164:
```suggestion

#### Imds Struct:
```

doc/configuration.md
dongsupark: Yes, I understand that listing configs in the markdown docs could be meaningful during design phase or as draft pull request. However, keeping such details in separate documents does not look like an idiomatic Rust style. We should either move them to `.rs` files or simply remove.

libazureinit/Cargo.toml
```diff
@@ -20,16 +20,27 @@ block-utils = "0.11.1"
...
+glob = "0.3"
+toml = "0.5"
 regex = "1"
 lazy_static = "1.4"
 tempfile = "3.3.0"
+tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
+opentelemetry = "0.26"
+opentelemetry_sdk = "0.26"
+tracing-opentelemetry = "0.27"
+uuid = { version = "1.2", features = ["v4"] }
```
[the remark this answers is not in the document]
peytonr18: I'll go ahead and remove them for now, and they can be merged back in with the KVP PR. Thank you!

doc/configuration.md
peytonr18: I removed those details from the document, along with many others that may be difficult to maintain.

jeremycline (changes requested):
One last thing I noticed regarding duplicating some enums, but I think we're neeeeearly there. I should have noticed this first time around, sorry about that.

libazureinit/src/provision/mod.rs
```diff
@@ -23,156 +24,120 @@ use crate::User;
...
+            .backends
+            .iter()
+            .find_map(|backend| match backend {
+                HostnameProvisioner::Hostnamectl => {
+                    hostname::Provisioner::Hostnamectl.set(&self.hostname).ok()
+                }
+                #[cfg(test)]
+                HostnameProvisioner::FakeHostnamectl => Some(()),
+            })
+            .ok_or_else(|| Error::NoHostnameProvisioner)?;
```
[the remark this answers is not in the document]
peytonr18: This is super helpful, thank you! I'll go ahead and make the change just because I support any changes to making it better, no matter how small, but thank you again for pointing that out!

libazureinit/src/provision/ssh.rs
```diff
@@ -333,7 +345,9 @@ mod tests {
             },
         ];
 
-        provision_ssh(&user, &keys, Some(".ssh/xauthorized_keys".to_string()))
+        let authorized_keys_path = user.dir.join(".ssh/xauthorized_keys");
+
+        provision_ssh(&user, &keys, authorized_keys_path.clone(), false)
```
dongsupark, line 350: Why is `authorized_keys_path` being cloned? Without `.clone()` it seems to work fine.

libazureinit/src/provision/mod.rs
```diff
@@ -23,156 +24,120 @@ use crate::User;
...
+                    password::Provisioner::Passwd.set(&self.user).ok()
+                }
+                #[cfg(test)]
+                PasswordProvisioner::FakePasswd => Some(()),
             })
             .ok_or(Error::NoPasswordProvisioner)?;
 
         if !self.user.ssh_keys.is_empty() {
+            let authorized_keys_path =
+                self.config.ssh.authorized_keys_path.clone();
```
dongsupark, line 90: Same, it works fine without `.clone()`.

libazureinit/src/provision/mod.rs
```diff
@@ -23,156 +24,120 @@ use crate::User;
...
+            let query_sshd_config = self.config.ssh.query_sshd_config;
+
             let user = nix::unistd::User::from_name(&self.user.name)?.ok_or(
                 Error::UserMissing {
-                    user: self.user.name,
+                    user: self.user.name.clone(),
```
dongsupark, line 95: Same, it works fine without `.clone()`.

libazureinit/src/provision/ssh.rs
```diff
@@ -371,7 +385,9 @@ mod tests {
...
```
dongsupark, line 390: Same, it works fine without `.clone()`.

libazureinit/src/provision/ssh.rs
```diff
@@ -396,18 +412,13 @@ mod tests {
...
-        .unwrap();
-        provision_ssh(
-            &user,
-            &keys[1..],
-            Some(".ssh/xauthorized_keys".to_string()),
-        )
-        .unwrap();
+        let authorized_keys_path = user.dir.join(".ssh/xauthorized_keys");
+
+        provision_ssh(&user, &keys[1..], authorized_keys_path.clone(), false)
```
dongsupark, line 417: On the other hand, this `clone` is indeed required.

libazureinit/src/provision/ssh.rs
```diff
@@ -396,18 +412,13 @@ mod tests {
...
+            .unwrap();
+
+        provision_ssh(&user, &keys[1..], authorized_keys_path.clone(), false)
```
dongsupark, line 420: On the other hand, this `clone` is indeed required.

libazureinit/Cargo.toml
```diff
@@ -21,16 +21,20 @@ block-utils = "0.11.1"
 tracing = "0.1.40"
 strum = { version = "0.26.3", features = ["derive"] }
 fstab = "0.4.0"
+toml = "0.5"
 regex = "1"
 lazy_static = "1.4"
 tempfile = "3.3.0"
+uuid = { version = "1.2", features = ["v4"] }
```
dongsupark, line 28: The `uuid` crate is not needed any more.

Tip: [cargo-machete](https://github.com/bnjbvr/cargo-machete) is your friend.

doc/configuration.md
```diff
@@ -0,0 +1,156 @@
...
+4. Fills in missing values with defaults from `config.rs`.
+
+## Validation and Deserialization Process
+
+Azure Init uses strict validation on configuration fields to ensure they match expected types and values. If a configuration includes an unsupported value or incorrect type, deserialization will fail.
+
+### Error Handling During Deserialization
+- When a configuration file is loaded, its contents are parsed and converted from `.toml` into structured data. If a field in the file contains an invalid value (e.g., `query_sshd_config` is set to `"not_a_boolean"` instead of `true``or `false`), the parsing process will fail with a deserialization error due to the mismatched type.
+
+### Propagation of Deserialization Errors
```
dongsupark, line 81: unnecessary whitespace at the end of line

doc/configuration.md
```diff
@@ -0,0 +1,156 @@
...
+  - `azure-init` attempts to dynamically query the authorized keys path using the `sshd -G` command.
+  - If `sshd -G` succeeds: The dynamically queried path is used for the authorized keys.
+  - If `sshd -G fail`s: The failure is logged, but azure-init continues using the fallback path specified in authorized_keys_path (default: `.ssh/authorized_keys`).
+- `query_sshd_config = false`:
+  - `azure-init` skips querying `sshd -G` entirely
+  - The value in `authorized_keys_path` is used directly, without any dynamic path detection.
+
+### 3. Handling of Provisioners in `azure-init`
+
+The `azure-init` configuration allows for custom settings of hostnames, user creation, and password setup through the use of provisioners. If `backends` are specified but do not contain a valid provisioner, `azure-init` logs an error  and halts provisioning.
```
dongsupark, line 150: unnecessary whitespace at the end of line

doc/configuration.md
```diff
@@ -0,0 +1,156 @@
...
+  - `azure-init` skips querying `sshd -G` entirely
+  - The value in `authorized_keys_path` is used directly, without any dynamic path detection.
+
+### 3. Handling of Provisioners in `azure-init`
+
+The `azure-init` configuration allows for custom settings of hostnames, user creation, and password setup through the use of provisioners. If `backends` are specified but do not contain a valid provisioner, `azure-init` logs an error  and halts provisioning. 
+
+## Package Considerations
+
+To ensure smooth operation and compatibility across distrubtions, `azure-init`, should be packaged with a consistent configuration setup.
```
dongsupark, line 154:
```suggestion
To ensure smooth operation and compatibility across distributions, `azure-init`, should be packaged with a consistent configuration setup.
```

dongsupark: I did not run a spell checker, but it seems to worth running it for documents.

libazureinit/src/provision/mod.rs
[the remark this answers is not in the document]
anhvoms: @jeremycline This comment definitely sparked my interest some of these subtle differences between Rust constructs. It seems to me that other than slightly better readability, there's not much advantage to using `ok_or`.
I shamelessly took this [sample code ](https://medium.com/@techhara/rust-tip-and-trick-2-5a92641191f6#:~:text=The%20difference%20between%20Result%3A%3A,argument%2C%20rather%20than%20a%20value.) and modified it to use the same `thiserror` enum that we're using and ran a little experiment comparing `ok_or(Error::NoHostnameProvisioner.to_string())` vs `ok_or_else(|| Error::NoHostnameProvisioner.to_string()). `ok_or_else` outperformed `ok_or` by 30%-40% in release build and about 10% in debug build. 

This level of optimization doesn't really matter for azure-init (since we have many expensive operations like calling into IMDS, wireserver...) but I why wouldn't it be ok using `or_else` constructs everywhere just to make it simple?

libazureinit/src/config.rs
```diff
@@ -0,0 +1,1004 @@
...
+    /// * `base` - The base TOML table to merge into
+    /// * `additional` - The additional TOML table whose values take precedence
+    fn merge_toml(
+        mut base: toml::Value,
+        additional: toml::Value,
+    ) -> toml::Value {
+        if let (Some(base_table), Some(additional_table)) =
+            (base.as_table_mut(), additional.as_table())
+        {
+            for (key, additional_value) in additional_table {
```
anhvoms, line 415: As discussed offline, I think this implementation will have some issues when the users provide bad input. Consider the example:
```
base:
[foo]
a=b
[bar]
c=d

additional:
foo=a

The whole section [foo] of base will be overwritten
```

We can probably address this by ensuring that additional is a valid azure-init's config file before merging. However, my recommendation is to use an existing well known and proven library like [figment](https://crates.io/crates/figment)

anhvoms (changes requested):

doc/configuration.md
peytonr18, line 1: I used VSCode's Markdownlint and Code Spell Checker extensions to check / update this file. Those two extensions fixed any remaining spelling errors or whitespace issues - highly recommend.

libazureinit/src/config.rs
```diff
@@ -0,0 +1,820 @@
...
+pub enum HostnameProvisioner {
+    #[default]
+    Hostnamectl,
+    #[cfg(test)]
+    FakeHostnamectl,
+}
+
+#[derive(Default, Serialize, Deserialize, Debug, Clone, Copy, PartialEq)]
+#[serde(rename_all = "lowercase")]
+pub enum UserProvisioner {
```
jeremycline, line 33: Sorry, I should have explicitly called this out, but we'll want the same fix-up you did for https://github.com/Azure/azure-init/pull/137/files#r1854279256 to this enum as well as the PasswordProvisioner enum below.

peytonr18: No, that's on me! Should have caught that, thank you! Will fix them all:)

doc/configuration.md
```diff
@@ -0,0 +1,170 @@
...
+├── azure-init.toml                # Base configuration
+└── azure-init.toml.d/
+    ├── 01-network.toml            # Network configuration
+    ├── 02-ssh.toml                # SSH configuration
+    └── 99-overrides.toml          # Overrides
+```
+
+**Order of Merging:**
+
+1. Loads `azure-init.toml` as the base configuration.
```
anhvoms, line 78: This is not the order in the code. While I think they work out to be the same. The order in the code would actually populates the config with the default first, then it loads azure-init.toml, then it applies configs found in the directory.

doc/configuration.md
```diff
@@ -0,0 +1,170 @@
...
+## Configuration Hierarchy for `azure-init`
+
+Configuration can be set via a single file, a directory containing multiple files, or default values defined in the code.
+
+### Configuration Loading Order
+
+#### 1. CLI Override (`--config`)
+
+- The `--config` flag specifies the path to load the configuration from, which can be either a single file or a directory.
+  - **File:** If the path points to a file, only that file is loaded.
```
anhvoms, line 36: From reading the code, this is not true. This file is still merged with the base azure-init.toml and also with defaults.
The order of merging part should be applicable to both files and directory merging.

libazureinit/src/config.rs
```diff
@@ -0,0 +1,820 @@
...
+/// 2. **Main File**: `azure-init.toml`, if present.
+/// 3. **Directory Files**: `.toml` files in `azure-init.d`, sorted lexicographically.
+/// 4. **CLI Overrides**: A file or directory specified via the CLI.
+///
+/// Later sources override earlier ones in case of conflicts.
+impl Config {
+    pub fn load(path: Option<PathBuf>) -> Result<Config, Error> {
+        let mut figment = Figment::new()
+            .merge(Serialized::defaults(Config::default()))
+            .merge(Toml::file("azure-init.toml"));
```
anhvoms, line 275: what if this file azure-init.toml doesn't exist?

doc/configuration.md
```diff
@@ -0,0 +1,170 @@
...
+└── azure-init.toml.d/
+    ├── 01-network.toml            # Network configuration
+    ├── 02-ssh.toml                # SSH configuration
+    └── 99-overrides.toml          # Overrides
+```
+
+**Order of Merging:**
+
+1. Loads `azure-init.toml` as the base configuration.
+2. Merges `.d` files in alphabetical order:
```
anhvoms, line 79: We mention lexicographical order, but that will only dictate which file is being parses first (you sort them in some order and read them). We need to be explicit about whether the first value sticks, or last value sticks.

libazureinit/src/config.rs
```diff
@@ -0,0 +1,820 @@
...
+                    .filter_map(Result::ok)
+                    .map(|entry| entry.path())
+                    .filter(|path| {
+                        path.extension().map_or(false, |ext| ext == "toml")
+                    })
+                    .collect();
+
+                entries.sort();
+
+                for path in entries {
```
anhvoms, line 293: perhaps `for path_entry in entries` so that we don't override the same variable usage "path" (which is used in the let statement)

anhvoms (changes requested):

doc/configuration.md
peytonr18: Last file in the sorted order (in this example, it would be the `99-overrides.toml` file) sticks. I'll make that clearer in the document.

libazureinit/src/config.rs
```diff
@@ -0,0 +1,857 @@
...
+                format!("Configuration error: {:?}", e),
+            ))
+        })
+    }
+
+    /// Helper function to merge `.toml` files from a directory into the Figment configuration.
+    fn merge_toml_directory(
+        figment: &Figment,
+        dir_path: PathBuf,
+    ) -> Result<Figment, Error> {
```
jeremycline, line 313: An alternative pattern you can use here is:

```Rust
fn merge_toml_directory(
    mut figment: Figment,
    dir_path: PathBuf,
) -> Result<Figment, Error> {
```

This gives you ownership of the incoming Figment object, which you then modify before passing ownership _back_ to the caller. With this approach you don't need to do

```Rust
            let mut updated_figment = figment.clone();
            for path_entry in entries {
                tracing::info!("Merging configuration file: {:?}", path_entry);
                updated_figment = updated_figment.merge(Toml::file(path_entry));
            }
            Ok(updated_figment)
```

and can instead do

```Rust
            for path_entry in entries {
                tracing::info!("Merging configuration file: {:?}", path_entry);
                figment = figment.merge(Toml::file(path_entry));
            }
            Ok(figment)
```

It allows you to avoid an unnecessary clone. This is obviously not a performance-sensitive bit of code so avoiding a clone is not something we care much about, but it's a pattern you'll see lots of places (figment itself is doing it with the `merge` function) and it's nice to be aware of.

jeremycline (changes requested):
Looking really tidy, my only concern is the relative paths used for the config locations

doc/configuration.md
anhvoms: nit:
```suggestion
2. Merges `.d` files in lexicographical order:
```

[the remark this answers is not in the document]
anhvoms: The wording needs to be updated:
alphabetical -> lexicographical 

They're not `.d` files. They're toml files in the custom configuration directory

anhvoms (approved):

libazureinit/src/config.rs
```diff
@@ -0,0 +1,899 @@
...
+/// 3. **Directory Files**: `.toml` files in `azure-init.d`, sorted lexicographically.
+/// 4. **CLI Overrides**: A file or directory specified via the CLI.
+///
+/// Later sources override earlier ones in case of conflicts.
+impl Config {
+    const BASE_CONFIG: &'static str = "/etc/azure-init.toml";
+    const DROP_IN_CONFIG: &'static str = "/etc/azure-init.d/";
+
+    /// Public wrapper function that uses default locations
+    pub fn load(path: Option<PathBuf>) -> Result<Config, Error> {
```
jeremycline, line 276: The `///` renders it as documentation and since the fact that this is a wrapper is an internal implementation detail, we shouldn't call it out in public docs. Something like

```Rust
/// Load provisioning configuration.
///
/// In addition to the provided path, configuration will also be loaded from the default locations.
```

Let's developers know it's what they want to use to load config from the filesystem.

jeremycline (approved):
I have one very small nit-pick with a docblock, but otherwise I think this is done, great work!

PR #143, opened 2024-11-12
Cargo: remove MSRV-pinned crates for Rust 1.73

dongsupark:
Remove pinned versions of crates `predicates*`, as they are now able to be compiled with Rust 1.73, CI-specfied version of Rust.


dongsupark:
Too early to do that. Will revisit when Rust 1.83 is released with MSRV-aware cargo.

PR #146, opened 2024-11-22
Cargo: remove MSRV-pinned crates for Rust 1.74

dongsupark:
Remove pinned versions of crates `clap`, `predicates*`, as they are now able to be compiled with Rust 1.74, CI-specfied version of Rust.

Set `package.rust-version` to 1.74 to be used by other crates that would make use of [MSRV-aware resolver](https://blog.rust-lang.org/inside-rust/2024/10/31/this-development-cycle-in-cargo-1.83.html#msrv-aware-cargo) which is available in Rust 1.83 or newer.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,7 @@
 name = "azure-init"
 version = "0.1.1"
 edition = "2021"
+rust-version = "1.74"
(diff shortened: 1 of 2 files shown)


jeremycline (approved):
CI likes it, so everything seems to be good

Issue #121, opened 2024-09-04
[RFE] azure-init should send its version as part of User-Agent string for IMDS request


anhvoms:
## Current situation

Currently azure-init is sending the User-Agent string "azure-init test" when making imds query. azure-init should be sending the its version (e.g., "azure-init v0.1.1") for released binary and it should be sending the commit hash appended to existing version during nightly test.

## Impact

This allows easy troubleshooting and correlation by using platform telemetry. It will help us quickly narrow down which version of azure-init has issues etc...

John15321:
I cannot append my PR to this issue with the permissions that I have so here is just a link to my PR for this 
Just a draft for now tho 😄 
https://github.com/Azure/azure-init/pull/153

dongsupark:
Done by https://github.com/Azure/azure-init/pull/153. Thanks.

PR #153, opened 2025-01-16
Add Version and Commit Hash to User-Agent String

John15321:
This PR updates azure-init to send its version (e.g., `azure-init v0.1.1`) as part of the User-Agent string for released binaries. For nightly tests, it appends the commit hash to the existing version (e.g., `azure-init v0.1.1-abc1234`).

## Testing done

I have tested this change using a newly created local binary crate to ensure that the environment variable works as expected. However, I do not have the setup to test azure-init directly. Any guidance on testing `azure-init` would be appreciated.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -136,7 +136,7 @@ async fn main() -> ExitCode {
 async fn provision(config: Config, opts: Cli) -> Result<(), anyhow::Error> {
     let mut default_headers = header::HeaderMap::new();
     let user_agent = if cfg!(debug_assertions) {
-        format!("azure-init {}-{}", VERSION, COMMIT_HASH)
+        format!("azure-init v{}-{}", VERSION, COMMIT_HASH)
     } else {
         format!("azure-init {}", VERSION)
     };
--- a/src/main.rs
+++ b/src/main.rs
@@ -21,7 +21,9 @@ use tracing_subscriber::fmt::format::FmtSpan;
 use tracing_subscriber::layer::SubscriberExt;
 use tracing_subscriber::EnvFilter;
 
+// These should be set during the build process
 const VERSION: &str = env!("CARGO_PKG_VERSION");
+const COMMIT_HASH: &str = env!("GIT_COMMIT_HASH");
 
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -133,9 +135,12 @@ async fn main() -> ExitCode {
 #[instrument]
 async fn provision(config: Config, opts: Cli) -> Result<(), anyhow::Error> {
     let mut default_headers = header::HeaderMap::new();
-    let user_agent = header::HeaderValue::from_str(
-        format!("azure-init v{VERSION}").as_str(),
-    )?;
+    let user_agent = if cfg!(debug_assertions) {
+        format!("azure-init {}-{}", VERSION, COMMIT_HASH)
```
dongsupark, line 139:
```suggestion
        format!("azure-init v{}-{}", VERSION, COMMIT_HASH)
```

src/main.rs
```diff
@@ -133,9 +135,12 @@ async fn main() -> ExitCode {
...
+    } else {
+        format!("azure-init {}", VERSION)
```
dongsupark, line 141: same, missing `v`

dongsupark (commented):
Thanks for picking up the issue.

src/main.rs
John15321: Done :)

dongsupark (approved):
Looks good.

Just FYI, it would be great if the PR description could include `Fixes https://github.com/Azure/azure-init/issues/121` or `Closes https://github.com/Azure/azure-init/issues/121` or etc, to automatically close the issue when the PR got merged.

John15321:
> Looks good.
> 
> Just FYI, it would be great if the PR description could include `Fixes https://github.com/Azure/azure-init/issues/121` or `Closes https://github.com/Azure/azure-init/issues/121` or etc, to automatically close the issue when the PR got merged.

I was actually looking for that kind of an option in the PR settings. But I thought that I dont see them since Im not a maintainer. I didnt know that just mentioning it in the PR would close it. Thanks!

PR #154, opened 2025-01-23
Add logs when updating sshd config and user passwords

jeremycline:
Include informational logs about operations that succeed and error level logs when important operations fail, like writing the sshd config snippet or a subprocess fails to run.


libazureinit/src/provision/password.rs
```diff
@@ -25,18 +25,34 @@ impl PasswordProvisioner {
...
     let path_passwd = env!("PATH_PASSWD");
 
     if user.password.is_none() {
-        let status = Command::new(path_passwd)
-            .arg("-d")
-            .arg(&user.name)
-            .status()?;
+        let mut command = Command::new(path_passwd);
+        command.arg("-d").arg(&user.name);
+        tracing::trace!(?command, "About to remove user's password");
```
jeremycline, line 41: One thing worth noting is this would log the username (if TRACE logging is enabled, which it wouldn't be by default)

PR #156, opened 2025-01-23
Amend password and sshd logic to support Mariner OS

peytonr18:
This PR amends the password and SSHD configuration logic to support Mariner OS while retaining compatibility with Ubuntu.

Key changes:

1. Added an OS check to determine the appropriate SSHD configuration path. If the OS is Mariner, the script uses `/etc/ssh/sshd_config`; otherwise, it defaults to Ubuntu's `/etc/ssh/sshd_config.d/50-azure-init.conf`.

2. Modified `update_sshd_config` to address a cross-device link error on Mariner when using `fs::rename`. For Mariner, the logic now uses `fs::copy` to copy the temporary file to the target location and `fs::remove_file` to clean up the temporary file. For non-Mariner systems, the existing `fs::rename` logic remains unchanged.

### DIFF
--- a/libazureinit/src/provision/password.rs
+++ b/libazureinit/src/provision/password.rs
@@ -3,7 +3,7 @@
 
 use std::process::Command;
 
-use std::fs::read_to_string;
+use std::path::PathBuf;
 
 use tracing::instrument;
 
@@ -23,15 +23,15 @@ impl PasswordProvisioner {
     }
 }
 
-// Function to determine the SSH config path based on the OS
-// Default for Ubuntu is:  "/etc/ssh/sshd_config.d/50-azure-init.conf"
+// Determines the appropriate SSH configuration file path based on the filesystem.
+// If the "/etc/ssh/sshd_config.d" directory exists, it returns the path for a drop-in configuration file.
+// Otherwise, it defaults to the main SSH configuration file at "/etc/ssh/sshd_config".
 fn get_sshd_config_path() -> &'static str {
-    if let Ok(content) = read_to_string("/etc/os-release") {
-        if content.contains("ID=azurelinux") {
-            return "/etc/ssh/sshd_config";
-        }
+    if PathBuf::from("/etc/ssh/sshd_config.d").is_dir() {
+        "/etc/ssh/sshd_config.d/50-azure-init.conf"
+    } else {
+        "/etc/ssh/sshd_config"
     }
-    "/etc/ssh/sshd_config.d/50-azure-init.conf"
 }
 
 #[instrument(skip_all)]
--- a/libazureinit/src/provision/password.rs
+++ b/libazureinit/src/provision/password.rs
@@ -3,6 +3,8 @@
 
 use std::process::Command;
 
+use std::fs::read_to_string;
+
 use tracing::instrument;
 
 use crate::{error::Error, User};
@@ -21,10 +23,21 @@ impl PasswordProvisioner {
     }
 }
 
+// Function to determine the SSH config path based on the OS
+// Default for Ubuntu is:  "/etc/ssh/sshd_config.d/50-azure-init.conf"
+fn get_sshd_config_path() -> &'static str {
+    if let Ok(content) = read_to_string("/etc/os-release") {
+        if content.contains("ID=azurelinux") {
+            return "/etc/ssh/sshd_config";
+        }
+    }
+    "/etc/ssh/sshd_config.d/50-azure-init.conf"
+}
+
 #[instrument(skip_all)]
 fn passwd(user: &User) -> Result<(), Error> {
     // Update the sshd configuration to allow password authentication.
-    let sshd_config_path = "/etc/ssh/sshd_config.d/50-azure-init.conf";
+    let sshd_config_path = get_sshd_config_path();
     if let Err(error) = update_sshd_config(sshd_config_path) {
         tracing::error!(
             ?error,
--- a/libazureinit/src/provision/ssh.rs
+++ b/libazureinit/src/provision/ssh.rs
@@ -250,23 +250,12 @@ pub(crate) fn update_sshd_config(
         temp_file.write_all(modified_content.as_bytes())?;
         temp_file.set_permissions(fs::Permissions::from_mode(0o644))?;
 
-        // Conditional logic for Mariner
-        if sshd_config_path == PathBuf::from("/etc/ssh/sshd_config") {
-            // Use fs::copy and fs::remove_file for Mariner
-            fs::copy(temp_sshd_config_path, &sshd_config_path)?;
-            fs::remove_file(temp_sshd_config_path)?;
-            tracing::info!(
-                ?sshd_config_path,
-                "Updated existing sshd setting to allow password authentication (Mariner)"
-            );
-        } else {
-            // Use fs::rename for other systems
-            fs::rename(temp_sshd_config_path, &sshd_config_path)?;
-            tracing::info!(
-                ?sshd_config_path,
-                "Updated existing sshd setting to allow password authentication (default)"
-            );
-        }
+        fs::copy(temp_sshd_config_path, &sshd_config_path)?;
+        fs::remove_file(temp_sshd_config_path)?;
+        tracing::info!(
+            ?sshd_config_path,
+            "Updated existing sshd setting to allow password authentication (Mariner)"
+        )
     } else {
         let mut file =
             OpenOptions::new().append(true).open(&sshd_config_path)?;
--- a/libazureinit/src/provision/ssh.rs
+++ b/libazureinit/src/provision/ssh.rs
@@ -236,15 +236,16 @@ pub(crate) fn update_sshd_config(
 
     let re = &PASSWORD_REGEX;
     if re.is_match(&file_content) {
-        let modified_content =
-            re.replace_all(&file_content, "PasswordAuthentication yes\n");
+        let modified_content = re.replace_all(
+            &file_content,
+            "PasswordAuthentication yes # modified by azure-init\n",
+        );
 
         let mut sshd_config = OpenOptions::new()
             .write(true)
             .truncate(true)
             .open(&sshd_config_path)?;
         sshd_config.write_all(modified_content.as_bytes())?;
-        sshd_config.set_permissions(Permissions::from_mode(0o600))?;
 
         tracing::info!(
             ?sshd_config_path,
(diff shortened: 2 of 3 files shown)


libazureinit/src/provision/password.rs
```diff
@@ -21,10 +23,21 @@ impl PasswordProvisioner {
     }
 }
 
+// Function to determine the SSH config path based on the OS
+// Default for Ubuntu is:  "/etc/ssh/sshd_config.d/50-azure-init.conf"
+fn get_sshd_config_path() -> &'static str {
```
vmarcella, line 28: Will other operating systems also use different sshd configuration paths? Should the API parameterize this so that the caller (Implementor of libazureinit) is responsible for providing the path?

libazureinit/src/provision/ssh.rs
```diff
@@ -250,11 +250,23 @@ pub(crate) fn update_sshd_config(
...
-            ?sshd_config_path,
-            "Updated existing sshd setting to allow password authentication"
-        )
+        // Conditional logic for Mariner
+        if sshd_config_path == PathBuf::from("/etc/ssh/sshd_config") {
+            // Use fs::copy and fs::remove_file for Mariner
+            fs::copy(temp_sshd_config_path, &sshd_config_path)?;
+            fs::remove_file(temp_sshd_config_path)?;
+            tracing::info!(
+                ?sshd_config_path,
```
vmarcella, line 259: What does `?` in front of the variable do here?

libazureinit/src/provision/password.rs
```diff
@@ -21,10 +23,21 @@ impl PasswordProvisioner {
...
+    if let Ok(content) = read_to_string("/etc/os-release") {
+        if content.contains("ID=azurelinux") {
+            return "/etc/ssh/sshd_config";
+        }
+    }
+    "/etc/ssh/sshd_config.d/50-azure-init.conf"
```
vmarcella, line 34: NIT: could we use an explicit return statement as opposed to an implicit one? I know it's not technically considered idiomatic rust, but I personally think that explicit return statements read better and are more obvious at a first glance

libazureinit/src/provision/ssh.rs
```diff
@@ -250,11 +250,23 @@ pub(crate) fn update_sshd_config(
...
 
-        fs::rename(temp_sshd_config_path, &sshd_config_path)?;
-        tracing::info!(
-            ?sshd_config_path,
-            "Updated existing sshd setting to allow password authentication"
-        )
+        // Conditional logic for Mariner
+        if sshd_config_path == PathBuf::from("/etc/ssh/sshd_config") {
+            // Use fs::copy and fs::remove_file for Mariner
+            fs::copy(temp_sshd_config_path, &sshd_config_path)?;
```
cjp256, line 256: is this going to overwrite the whole config with just the fragment?

vmarcella (changes requested):
Left a few comments, but looks good for the most part

libazureinit/src/provision/password.rs
peytonr18: I think that might be the idea in the future -- Jeremy and I synced offline and instead of doing it this way based on /etc/os-release, it'd be better to try to use "/etc/ssh/sshd_config.d/50-azure-init.conf" and fallback to /etc/ssh/sshd_config rather than do it based on distro versioning. 


Ultimately, I do think we'll have to figure out maybe a better way as more distros are added, but for now, it's probably the best approach

libazureinit/src/provision/ssh.rs
peytonr18: It formats the variable using its Debug implementation rather than just the Display version- which makes it show up a lot better in the serial console logs.

libazureinit/src/provision/ssh.rs
peytonr18: It does -- I'll amend this so it just appends here instead of re-writing it all again.

libazureinit/src/provision/password.rs
peytonr18: I did try to add them, but the clippy linter throws an error for them as "unnecessary return statements."

libazureinit/src/provision/ssh.rs
```diff
@@ -250,10 +250,11 @@ pub(crate) fn update_sshd_config(
         temp_file.write_all(modified_content.as_bytes())?;
         temp_file.set_permissions(fs::Permissions::from_mode(0o644))?;
 
-        fs::rename(temp_sshd_config_path, &sshd_config_path)?;
+        fs::copy(temp_sshd_config_path, &sshd_config_path)?;
+        fs::remove_file(temp_sshd_config_path)?;
         tracing::info!(
             ?sshd_config_path,
-            "Updated existing sshd setting to allow password authentication"
+            "Updated existing sshd setting to allow password authentication (Mariner)"
```
jeremycline, line 257: I'd recommend leaving the log statement as-is since it could easily _not_ be Mariner/Azure Linux; this path gets hit anytime the file exists (the config snippet could exist from a previous provisioning, for example)

libazureinit/src/provision/ssh.rs
```diff
@@ -250,10 +250,11 @@ pub(crate) fn update_sshd_config(
         temp_file.write_all(modified_content.as_bytes())?;
         temp_file.set_permissions(fs::Permissions::from_mode(0o644))?;
 
-        fs::rename(temp_sshd_config_path, &sshd_config_path)?;
+        fs::copy(temp_sshd_config_path, &sshd_config_path)?;
+        fs::remove_file(temp_sshd_config_path)?;
```
jeremycline, line 254: I've ruminated about this a little bit and I think we'd be better off avoiding this tempfile copy dance by opening up `sshd_config_path` with `write(true)` and `truncate(true)`. `fs::copy` is [ultimately doing that anyway](https://doc.rust-lang.org/nightly/std/fs/fn.copy.html#platform-specific-behavior).

There's also some (extremely-to-the-point-of-absurdly unlikely in our use-case) security issues with using tempfiles in `/tmp` since it's a globally shared directory. I also noticed that we're setting file permissions to `0o644` which is more open than it should be; most distributions use `0o600`.

All that to say, let's just do:
```Rust
        let mut sshd_config = OpenOptions::new().write(true).truncate(true).open(&sshd_config_path)?;
        sshd_config.write_all(modified_content.as_bytes())?;
```

jeremycline (changes requested):

libazureinit/src/provision/password.rs
dongsupark: I agree with @vmarcella here, hard-coding a path with checking for `/etc/os-release` etc. is ideal.

> I think that might be the idea in the future -- Jeremy and I synced offline and instead of doing it this way based on /etc/os-release, it'd be better to try to use "/etc/ssh/sshd_config.d/50-azure-init.conf" and fallback to /etc/ssh/sshd_config rather than do it based on distro versioning.

@peytonr18 Can you please create a new issue for future improvements, summarizing what you discussed with Jeremy? That would help all of other folks keep track of discussions, without forgetting about what we should do next. Thanks.

jeremycline: The upshot of our discussion was that we can add a configuration option for the snippet location. I really don't think we want to be inspecting `/etc/os-release` and make hard-coded decisions off it. There's plenty of distribution derivatives popping up with tweaked `os-release` so unless we really have to, I'd prefer it to be configuration with reasonable defaults.

Nearly all distributions use `/etc/ssh/sshd_config` as the main config, and if they've set up an `Include` directive it's usually `/etc/ssh/sshd_config.d/`. If neither of those are there, the config is a nice escape hatch. I think doing the config as a follow-up is okay since with this adjustment we'll handle most distros.

jeremycline (approved):
You _should_ be able to drop the dependency on `tempfile` (but not the `dev-dependency`) now.

Everything else looks in good shape ✨

libazureinit/src/provision/password.rs
cjp256: I would think it would be best to maintain the /etc/ssh/sshd_config.d/50-azure-init.conf fragment, and if the directory doesn't exist, create it.

If `Include /etc/ssh/sshd_config.d/*.conf` is not in /etc/ssh/sshd_config.d, append `Include /etc/ssh/sshd_config.d/50-azure-init.conf` to /etc/ssh/sshd_config.

This way it's clear who wrote the config (override) and it's a simple append to the base config.

cjp256: Also, worth asking the Mariner folks to support the Include in the future.

libazureinit/src/provision/ssh.rs
```diff
@@ -240,29 +239,27 @@ pub(crate) fn update_sshd_config(
...
         tracing::info!(
             ?sshd_config_path,
             "Updated existing sshd setting to allow password authentication"
-        )
+        );
     } else {
         let mut file =
             OpenOptions::new().append(true).open(&sshd_config_path)?;
         file.write_all(b"PasswordAuthentication yes\n")?;
+        file.set_permissions(Permissions::from_mode(0o600))?;
```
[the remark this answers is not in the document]
cjp256: In many cases sshd config is world readable (644). +1 to removing.

libazureinit/src/provision/password.rs
jeremycline: I'm a little nervous about the indirection if the distribution doesn't set up a drop-in directory. Someone familiar with the distribution will know if there are drop-in configs normally and introducing them could cause that person to miss that password authentication is on. Given it's rather unwise to have password authentication on in the first place, any opportunity to miss that setting is a little scary.

Obviously, that's a bit contrived, unusual, and it's mostly a gut feeling so... Take it as you will. I like your suggestion to include a comment that we added the setting.

peytonr18: I went ahead and implemented that suggestion to add a comment that we added the setting. I'll add all of this to an issue so we can discuss it further there with a record of what ideas have already been shared.

libazureinit/src/provision/ssh.rs
```diff
@@ -240,29 +239,26 @@ pub(crate) fn update_sshd_config(
...
+        let mut sshd_config = OpenOptions::new()
             .write(true)
-            .create(true)
             .truncate(true)
-            .open(temp_sshd_config_path)?;
-        temp_file.write_all(modified_content.as_bytes())?;
-        temp_file.set_permissions(fs::Permissions::from_mode(0o644))?;
+            .open(&sshd_config_path)?;
+        sshd_config.write_all(modified_content.as_bytes())?;
+        sshd_config.set_permissions(Permissions::from_mode(0o600))?;
```
cjp256, line 247: did this sneak back in?

peytonr18: Super weird - I swore I deleted that and it stayed out. Maybe something weird happened locally -- will make sure it goes back out!

cjp256 (approved):
looks great, thanks!

vmarcella (approved):

PR #164, opened 2025-02-12
Adding Cached VM ID File for Provisioning Lifecycle Management 

### LINKED ISSUE #120: [RFE] azure-init should not attempt to provision the VM on reboot
azure-init pulls data from IMDS to configure the VM and reports health on every boot. This has some implications:
1) Unnecessary communication with wireserver (goalstate and health). If there is an issue with wireserver communication, this might unnecessarily delay boot if there are components that run after azure-init. azure-init. The extra health reporting is currently harmless, but should be avoided if possible.
2) If the user makes a change to the hostname, azure-init will update the hostname back to the value found in IMDS. This is often undesirable (however, the other side of this argument is that the customer can update the hostname in the VM, and they would like to see the VM hostname updated upon reboot). This should be addressed with a configurable option (cloud-init has has a similar [option](https://cloudinit.readthedocs.io/en/latest/reference/modules.html#set-hostname))

## **Implementation options

The easiest option is to write a file (/var/lib/azure-init/cache, e.g.,) with the existing vm id. vm id is available in IMDS, it's also available by reading the uuid provided by dmidecode -- see [here](https://github.com/canonical/cloud-init/blob/8bc3e42543d6be2624a439fe0b72b4546928667e/cloudinit/sources/azure/identity.py#L65). Using the id from dmidecode is preferred because it avoids unnecessary communication (IMDS), it also will work in the event that IMDS is down.  

Upon reboot, azure-init checks if the vm id has changed (which means that this is a new VM), and if yes, it should proceed with provision the VM.

peytonr18:
This PR adds a new `status.rs` module that helps manage the **provisioning lifecycle** for Azure Linux VMs.  
The core idea is to **create and check** a `.provisioned` file named after the VM ID when provisioning is completed, ensuring we skip provisioning on reboot if the file already exists. Additionally, this file can serve as an indicator that provisioning has been completed.

---

### **Provisioning File Logic**
- We now store a file at `\{provision_dir}/{vm_id}.provisioned`.
- If the file exists, provisioning is **considered complete**; otherwise, we run provisioning.
- The provisioning directory is configurable through `Config.provisioning_dir` and defaults to `/var/lib/azure-init/`.

### **VM ID Retrieval**
- Production retrieves the VM ID by reading `/sys/class/dmi/id/product_uuid` and byte-swapping if Gen1.
- `is_vm_gen1()` accepts a mock EFI path, allowing test scenarios to simulate different VM generations without relying on actual system files.

### **Integration with `main.rs`**
- `main.rs` checks `is_provisioning_complete()` to see if it needs to run provisioning.
- After successful provisioning, `mark_provisioning_complete()` creates the `.provisioned` file.
- On subsequent boots, if the file is found, provisioning is skipped.
- Additionally, main.rs retrieves the `vm_id` once and passes it through all necessary functions.

### **Unit Test Setup**
- **Environment variables** drive our partial mocking in tests:
  - `TEST_PROVISION_DIR` → points to a `TempDir` instead of `/var/lib/azure-init/`.
  -  tests pass an explicit file path for VM ID retrieval.

- **Tests verify** that:
  - **First run**: no provisioning file → provisioning needed.
  - **File is created**: `mark_provisioning_complete()` actually writes the file.
  - **Subsequent runs**: the `.provisioned` file is detected, so provisioning is skipped (simulated reboot test).

- **Linked Issues**:
  - Resolves #120 
  - Step 1 in integrating `azure-init` with WALinuxAgent

### DIFF
--- a/libazureinit/src/config.rs
+++ b/libazureinit/src/config.rs
@@ -224,6 +224,25 @@ impl Default for Telemetry {
     }
 }
 
+/// Provisioning directory configuration struct.
+///
+/// Configures settings for where azure-init should store provisioning-related files.
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(default)]
+pub struct ProvisioningDir {
+    /// Specifies the path used for storing provisioning status files.
+    /// Defaults to `/var/lib/azure-init/`.
+    pub path: PathBuf,
+}
+
+impl Default for ProvisioningDir {
+    fn default() -> Self {
+        Self {
+            path: PathBuf::from("/var/lib/azure-init/"),
+        }
+    }
+}
+
 /// General configuration struct for azure-init.
 ///
 /// Aggregates all configuration settings for managing SSH, provisioning, IMDS, media,
@@ -240,6 +259,7 @@ pub struct Config {
     pub azure_proxy_agent: AzureProxyAgent,
     pub wireserver: Wireserver,
     pub telemetry: Telemetry,
+    pub provisioning_dir: ProvisioningDir,
 }
 
 /// Implements `Display` for `Config`, formatting it as a readable TOML string.
@@ -597,6 +617,11 @@ mod tests {
 
         assert!(config.telemetry.kvp_diagnostics);
 
+        assert_eq!(
+            config.provisioning_dir.path.to_str().unwrap(),
+            "/var/lib/azure-init/",
+        );
+
         tracing::debug!("test_empty_config_file_uses_defaults_when_merged completed successfully.");
 
         Ok(())
@@ -630,6 +655,8 @@ mod tests {
         enable = false
         [telemetry]
         kvp_diagnostics = false
+        [provisioning_dir]
+        path = "/custom/provisioning/dir"
         "#
         )?;
 
@@ -685,6 +712,14 @@ mod tests {
 
         tracing::debug!("Verifying merged telemetry configuration...");
         assert!(!config.telemetry.kvp_diagnostics);
+
+        tracing::debug!(
+            "Verifying merged provisioning directory configuration..."
+        );
+        assert_eq!(
+            config.provisioning_dir.path.to_str().unwrap(),
+            "/custom/provisioning/dir"
+        );
 
         tracing::debug!(
             "test_load_and_merge_with_default_config completed successfully."
@@ -750,6 +785,14 @@ mod tests {
         tracing::debug!("Verifying default telemetry configuration...");
         assert!(config.telemetry.kvp_diagnostics);
 
+        tracing::debug!(
+            "Verifying default provisioning directory configuration..."
+        );
+        assert_eq!(
+            config.provisioning_dir.path.to_str().unwrap(),
+            "/var/lib/azure-init/"
+        );
+
         tracing::debug!("test_default_config completed successfully.");
 
         Ok(())
@@ -780,6 +823,8 @@ mod tests {
         enable = false
         [telemetry]
         kvp_diagnostics = false
+        [provisioning_dir]
+        path = "/cli-override-provisioning-dir"
         "#,
         )?;
 
@@ -822,6 +867,10 @@ mod tests {
         assert!(!config.provisioning_media.enable);
         assert!(!config.azure_proxy_agent.enable);
         assert!(!config.telemetry.kvp_diagnostics);
+        assert_eq!(
+            config.provisioning_dir.path.to_str().unwrap(),
+            "/cli-override-provisioning-dir"
+        );
 
         Ok(())
     }
--- a/libazureinit/src/config.rs
+++ b/libazureinit/src/config.rs
@@ -224,29 +224,29 @@ impl Default for Telemetry {
     }
 }
 
-/// The default directory for storing provisioning status files.
+/// The default directory for storing azure-init data files, such as the provisioning status file.
 ///
-/// This constant is declared outside its related struct so that both the `ProvisioningDir` struct
+/// This constant is declared outside its related struct so that both the `AzureInitDataDir` struct
 /// and other modules (like `status.rs`) can reference the same path without risking any mismatch.
-pub const DEFAULT_PROVISIONING_DIR: &str = "/var/lib/azure-init/";
+pub const DEFAULT_AZURE_INIT_DATA_DIR: &str = "/var/lib/azure-init/";
 
-/// Provisioning directory configuration struct.
+/// Azure-init data directory directory configuration struct.
 ///
-/// Configures settings for where azure-init should store provisioning-related files.
-/// If no custom path is provided, `ProvisioningDir::default()` uses
-/// [`DEFAULT_PROVISIONING_DIR`], ensuring a single source of truth.
+/// Configures settings for where azure-init should store data (especially provisioning-related) files.
+/// If no custom path is provided, `AzureInitDataDir::default()` uses
+/// [`DEFAULT_AZURE_INIT_DATA_DIR`], ensuring a single source of truth.
 #[derive(Serialize, Deserialize, Debug, Clone)]
 #[serde(default)]
-pub struct ProvisioningDir {
-    /// Specifies the path used for storing provisioning status files.
+pub struct AzureInitDataDir {
+    /// Specifies the path used for storing azure-init data files.
     /// Defaults to `/var/lib/azure-init/`.
     pub path: PathBuf,
 }
 
-impl Default for ProvisioningDir {
+impl Default for AzureInitDataDir {
     fn default() -> Self {
         Self {
-            path: PathBuf::from(DEFAULT_PROVISIONING_DIR),
+            path: PathBuf::from(DEFAULT_AZURE_INIT_DATA_DIR),
         }
     }
 }
@@ -267,7 +267,7 @@ pub struct Config {
     pub azure_proxy_agent: AzureProxyAgent,
     pub wireserver: Wireserver,
     pub telemetry: Telemetry,
-    pub provisioning_dir: ProvisioningDir,
+    pub azure_init_data_dir: AzureInitDataDir,
 }
 
 /// Implements `Display` for `Config`, formatting it as a readable TOML string.
@@ -626,7 +626,7 @@ mod tests {
         assert!(config.telemetry.kvp_diagnostics);
 
         assert_eq!(
-            config.provisioning_dir.path.to_str().unwrap(),
+            config.azure_init_data_dir.path.to_str().unwrap(),
             "/var/lib/azure-init/",
         );
 
@@ -663,8 +663,8 @@ mod tests {
         enable = false
         [telemetry]
         kvp_diagnostics = false
-        [provisioning_dir]
-        path = "/custom/provisioning/dir"
+        [azure_init_data_dir]
+        path = "/custom/azure-init-data-dir"
         "#
         )?;
 
@@ -722,11 +722,11 @@ mod tests {
         assert!(!config.telemetry.kvp_diagnostics);
 
         tracing::debug!(
-            "Verifying merged provisioning directory configuration..."
+            "Verifying merged azure-init data directory configuration..."
         );
         assert_eq!(
-            config.provisioning_dir.path.to_str().unwrap(),
-            "/custom/provisioning/dir"
+            config.azure_init_data_dir.path.to_str().unwrap(),
+            "/custom/azure-init-data-dir"
         );
 
         tracing::debug!(
@@ -794,10 +794,10 @@ mod tests {
         assert!(config.telemetry.kvp_diagnostics);
 
         tracing::debug!(
-            "Verifying default provisioning directory configuration..."
+            "Verifying default azure-init data directory configuration..."
         );
         assert_eq!(
-            config.provisioning_dir.path.to_str().unwrap(),
+            config.azure_init_data_dir.path.to_str().unwrap(),
             "/var/lib/azure-init/"
         );
 
@@ -831,8 +831,8 @@ mod tests {
         enable = false
         [telemetry]
         kvp_diagnostics = false
-        [provisioning_dir]
-        path = "/cli-override-provisioning-dir"
+        [azure_init_data_dir]
+        path = "/cli-override-azure-init-data-dir"
         "#,
         )?;
 
@@ -876,8 +876,8 @@ mod tests {
         assert!(!config.azure_proxy_agent.enable);
         assert!(!config.telemetry.kvp_diagnostics);
         assert_eq!(
-            config.provisioning_dir.path.to_str().unwrap(),
-            "/cli-override-provisioning-dir"
+            config.azure_init_data_dir.path.to_str().unwrap(),
+            "/cli-override-azure-init-data-dir"
         );
 
         Ok(())
--- a/libazureinit/src/config.rs
+++ b/libazureinit/src/config.rs
@@ -224,9 +224,17 @@ impl Default for Telemetry {
     }
 }
 
+/// The default directory for storing provisioning status files.
+///
+/// This constant is declared outside its related struct so that both the `ProvisioningDir` struct
+/// and other modules (like `status.rs`) can reference the same path without risking any mismatch.
+pub const DEFAULT_PROVISIONING_DIR: &str = "/var/lib/azure-init/";
+
 /// Provisioning directory configuration struct.
 ///
 /// Configures settings for where azure-init should store provisioning-related files.
+/// If no custom path is provided, `ProvisioningDir::default()` uses
+/// [`DEFAULT_PROVISIONING_DIR`], ensuring a single source of truth.
 #[derive(Serialize, Deserialize, Debug, Clone)]
 #[serde(default)]
 pub struct ProvisioningDir {
@@ -238,7 +246,7 @@ pub struct ProvisioningDir {
 impl Default for ProvisioningDir {
     fn default() -> Self {
         Self {
-            path: PathBuf::from("/var/lib/azure-init/"),
+            path: PathBuf::from(DEFAULT_PROVISIONING_DIR),
         }
     }
 }
--- a/libazureinit/src/status.rs
+++ b/libazureinit/src/status.rs
@@ -1,0 +1,187 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT License.
+
+//! Provisioning status management for azure-init.
+//!
+//! This module ensures that provisioning is performed only when necessary
+//! by tracking the VM ID. It stores a provisioning status file named after
+//! the VM ID in a persistent location (`/var/lib/azure-init/`).
+//!
+//! # Logic Overview
+//! - Retrieves the VM ID using `dmidecode`.
+//! - Determines if provisioning is required by checking if a status file exists.
+//! - Creates the provisioning status file upon successful provisioning.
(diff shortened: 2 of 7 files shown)


libazureinit/src/status.rs
```diff
@@ -0,0 +1,187 @@
...
+
+/// Retrieves the VM ID using `dmidecode`.
+///
+/// The VM ID is a unique system identifier that persists across reboots but changes
+/// when a VM is cloned or redeployed.
+///
+/// # Returns
+/// - `Some(String)` containing the VM ID if retrieval is successful.
+/// - `None` if `dmidecode` fails or the output is empty.
+fn get_vm_id() -> Option<String> {
```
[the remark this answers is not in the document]
peytonr18: Resolved dependency, no longer depends on dmidecode!

anhvoms:
Since dmidecode is not used anymore, please fix the description to remove reference to dmidecode

anhvoms:
I would recommend an additional cli feature to clear the provisioning directory. This might come in handy if we need to force azure-init to provision again.

libazureinit/src/status.rs
```diff
@@ -0,0 +1,254 @@
...
+///
+/// The VM ID is a unique system identifier that persists across reboots but changes
+/// when a VM is cloned or redeployed.
+///
+/// # Returns
+/// - `Some(String)` containing the VM ID if retrieval is successful.
+/// - `None` if something fails or the output is empty.
+pub fn get_vm_id() -> Option<String> {
+    // Test override check
+    if let Ok(mock_id) = std::env::var("MOCK_VM_ID") {
```
cjp256, line 91: TEST_PROVISION_DIR vs MOCK_VM_ID

if there are going to be env vars to support testing, maybe we need a scheme that's consistent.

e.g.  `AZURE_INIT_TEST_PROVISION_DIR`  `AZURE_INIT_TEST_VM_ID`

cjp256: it's possible to avoid this mock altogether by allowing an optional VM id param to 
- is_provisioning_complete()
- mark_provisioning_complete()

but to get the best testing, it may be best to allow overriding "/sys/class/dmi/id/product_uuid" so we can put whatever id in there we want and test the scenarios.  

is_vm_gen1() would some special care to mock or parameterized option for get_vm_id()

cjp256: it might be practical for testing if all paths derived from some configurable root path and any file structure is easily faked, but certainly beyond the scope of this PR

libazureinit/src/status.rs
```diff
@@ -0,0 +1,254 @@
...
+use byteorder::{BigEndian, ByteOrder, LittleEndian};
+use std::env;
+use std::fs::{self, File};
+use std::path::Path;
+use uuid::Uuid;
+
+use crate::error::Error;
+
+/// Directory where the provisioning status files are stored.
+const DEFAULT_PROVISION_DIR: &str = "/var/lib/azure-init/";
```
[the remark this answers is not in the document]
peytonr18: It's a good idea -- I'll put it there!

libazureinit/src/status.rs
peytonr18: I can easily make an issue for the last idea and circle back to it!

For this,  I updated `is_provisioning_complete()` and `mark_provisioning_complete()` to accept an optional vm_id, so we can pass in a test VM ID directly rather than relying on get_vm_id() or MOCK_VM_ID. In addition, I removed MOCK_VM_ID from get_vm_id() itself. Now it takes an optional path argument to override /sys/class/dmi/id/product_uuid, which avoids environment variables for testing.

I also updated is_vm_gen1() to accept an optional mock_efi_path, letting us simulate Gen1 or Gen2 in tests without touching real system files. 

That was definitely a little more complicated than I anticipated, so please take a look and let me know what you think!

peytonr18:
I amended the logic in `main.rs`, `config.rs` and `status.rs` so we use a configurable struct for the `provisioining_dir` instead of a hardcoded CONST env variable.  With this change though, we have to pass the config into `status.rs` and that made it necessary to create a copy of the config struct in main (because `Config` by nature doesn't allow for `Copy`).  

To me, it seemed better to do it as I show below because I anticipate needing a cloned version of config when I complete some of the issues related to toggling kvp telemetry on and off:

```rust
let clone_config: Config = config.clone();
match provision(config, opts).await {
mark_provisioning_complete(Some(&clone_config), None)
```

instead of doing this and cloning the config inline.

```rust
match provision(config.clone(), opts).await
mark_provisioning_complete(Some(&config), None)
```

This is one part of the PR that I would love feedback on if anyone else has a thought or a Rust standard we need to conform to!

libazureinit/src/status.rs
```diff
@@ -0,0 +1,342 @@
...
+    swapped[4] = (d2_le >> 8) as u8;
+    swapped[5] = (d2_le) as u8;
+
+    swapped[6] = (d3_le >> 8) as u8;
+    swapped[7] = (d3_le) as u8;
+
+    swapped[8..16].copy_from_slice(&bytes[8..16]);
+
+    swapped
+}
```
jeremycline, line 110: You should be able to drop the byteorder dependency and avoid some bit shifting with something like the following: 
```Rust
fn swap_uuid_to_little_endian(mut bytes: [u8; 16]) -> Uuid {
    let (d1, remainder) = bytes.split_at(std::mem::size_of::<u32>());
    let d1 = d1.try_into().map(u32::from_be_bytes).unwrap_or(0).to_ne_bytes();
    
    let (d2, remainder) = remainder.split_at(std::mem::size_of::<u16>());
    let d2 = d2.try_into().map(u16::from_be_bytes).unwrap_or(0).to_ne_bytes();

    let (d3, _) = remainder.split_at(std::mem::size_of::<u16>());
    let d3 = d3.try_into().map(u16::from_be_bytes).unwrap_or(0).to_ne_bytes();

    let native_endian= d1
        .into_iter()
        .chain(d2.into_iter())
        .chain(d3.into_iter())
        .collect::<Vec<_>>();
    debug_assert_eq!(native_endian.len(), 8);
    bytes[..native_endian.len()].copy_from_slice(&native_endian);
    uuid::Uuid::from_bytes(bytes)
}
```

Frustratingly, I can't find a great way to avoid the fallible calls; there's not a stable API to split an array into a known length. `split_at` gives us a slice, which even though we _know_ is the size of the u32, we have to use `try_into()` to attempt to convert it to `[u8; 4]`. I settled for using `unwrap_or(0)` although we should _never_ hit the or case.

libazureinit/src/status.rs
```diff
@@ -0,0 +1,342 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::fs;
+    use tempfile::TempDir;
+
+    #[test]
+    fn test_mark_provisioning_complete() {
```
jeremycline, line 231: I'm seeing this test fail occasionally locally. I believe this is because the cargo test runner is multi-threaded, so all your tests are stepping on top of each other. It's also worth noting that, unfortunately, setting environment variables in *nix is not [thread safe](https://doc.rust-lang.org/std/env/fn.set_var.html#safety). 

It's best to avoid setting an environment variable from your program if you can possibly help it. Even if you're absolutely sure there's no other threads, someone might come along in the future and add multi-threading without noticing the environment is getting modified.

src/main.rs
```diff
@@ -115,8 +116,26 @@ async fn main() -> ExitCode {
...
+        );
+        return ExitCode::SUCCESS;
+    }
+
+    let clone_config = config.clone();
     match provision(config, opts).await {
-        Ok(_) => ExitCode::SUCCESS,
+        Ok(_) => {
+            if let Err(e) =
+                mark_provisioning_complete(Some(&clone_config), None)
```
jeremycline, line 130: Ideally, I think, this function would be internal to the library and something it calls at the end of a successful provision. However, since the current `Provision` API has the user manually calling `goalstate::report_health` after it completes, perhaps we don't want this call to be done until after that?

We should consider moving the health reporting to be internal to the library as well and done for the user without additional calls, but if that's what we want to do, we can obviously do that separately.

src/main.rs
```diff
@@ -115,8 +116,26 @@ async fn main() -> ExitCode {
...
     };
 
+    if is_provisioning_complete(Some(&config), None) {
+        tracing::info!(
+            "Provisioning already completed earlier. Skipping provisioning."
+        );
+        return ExitCode::SUCCESS;
+    }
+
+    let clone_config = config.clone();
```
jeremycline, line 126: FWIW I don't think this is "odd" or breaking any Rust conventions I'm aware of so doing this or doing it inline are both fine

jeremycline (changes requested):
Unfortunately, I think you're going to need to adjust the tests a bit to avoid writing to the environment 😿

src/main.rs
peytonr18: I agree that having the library automatically call mark_provisioning_complete() once provisioning and health reporting are done would make the most sense! It'll probably be more intuitive to follow that logic too. I can file an issue so we can discuss there and decide if we want to wrap the entire flow (including report_health) inside the library? 

In the meantime, if it's good with you, I'll just move the mark_provisioning_complete() call to occur after report_health() so we’re at least marking provisioning complete at the very end of the process (which I agree makes more sense).

jeremycline: Works for me 👍🏼

libazureinit/src/status.rs
```diff
@@ -0,0 +1,337 @@
...
+use crate::error::Error;
+
+/// This function determines the effective provisioning directory.
+///
+/// If a [`Config`] is provided, this function returns `config.provisioning_dir.path`.
+/// Otherwise, it falls back to the default `/var/lib/azure-init/`.
+fn get_provisioning_dir(config: Option<&Config>) -> PathBuf {
+    config
+        .map(|cfg| cfg.provisioning_dir.path.clone())
+        .unwrap_or_else(|| PathBuf::from("/var/lib/azure-init/"))
```
anhvoms, line 36: Config already has a default. If there's a possibility that the config module will fail to provide a default, this value (`/var/lib/azure-init`) should be declared a const that both modules can access. This way there's only one possible default

peytonr18: Should've caught that in my look-over -- will update, thank you!

libazureinit/src/config.rs
```diff
@@ -224,6 +224,33 @@ impl Default for Telemetry {
     }
 }
 
+/// The default directory for storing provisioning status files.
+///
+/// This constant is declared outside its related struct so that both the `ProvisioningDir` struct
+/// and other modules (like `status.rs`) can reference the same path without risking any mismatch.
+pub const DEFAULT_PROVISIONING_DIR: &str = "/var/lib/azure-init/";
```
cjp256, line 231: consider calling it something like STATE_DIR or simar.  this path's scope is potentially wider than "provisioning" if azure-init wants non-provisioning state for some reason

libazureinit/src/status.rs
```diff
@@ -0,0 +1,337 @@
...
+    config: Option<&Config>,
+    vm_id: Option<String>,
+) -> Result<(), Error> {
+    check_provision_dir(config)?;
+
+    let vm_id = vm_id.or_else(|| get_vm_id(None, None));
+
+    if let Some(vm_id) = vm_id {
+        let file_path =
+            get_provisioning_dir(config).join(format!("{}.provisioned", vm_id));
```
cjp256, line 195: at some point we'd possibly want a similar marker file for pre-preprovisioned. I just wonder if it makes more sense to have state information in one file or to do these markers

libazureinit/src/config.rs
peytonr18: Should we change the config struct too then to something similar?
```rust
pub const DEFAULT_PERSISTENT_STATE_DIR: &str = "/var/lib/azure-init/";

#[derive(Serialize, Deserialize, Debug, Clone)]
#[serde(default)]
pub struct PersistentStateDir{
    /// Specifies the path used for storing provisioning status files.
    /// Defaults to `/var/lib/azure-init/`.
    pub path: PathBuf,
}

impl Default for PersistentStateDir {
    fn default() -> Self {
        Self {
            path: PathBuf::from(DEFAULT_PERSISTENT_STATE_DIR),
        }
    }
}
```

libazureinit/src/status.rs
```diff
@@ -0,0 +1,337 @@
...
+/// The VM ID is a unique system identifier that persists across reboots but changes
+/// when a VM is cloned or redeployed.
+///
+/// # Returns
+/// - `Some(String)` containing the VM ID if retrieval is successful.
+/// - `None` if something fails or the output is empty.
+pub fn get_vm_id(
+    custom_path: Option<&str>,
+    mock_efi_path: Option<&str>,
+) -> Option<String> {
```
jeremycline, line 116: Since you're exposing this as a public API, I would recommend keeping the test parameters private. The easiest way would be to do something like this:
 
```suggestion
/// Retrieves the VM ID by reading `/sys/class/dmi/id/product_uuid` and byte-swapping if Gen1.
///
/// The VM ID is a unique system identifier that persists across reboots but changes
/// when a VM is cloned or redeployed.
///
/// # Returns
/// - `Some(String)` containing the VM ID if retrieval is successful.
/// - `None` if something fails or the output is empty.
pub fn get_vm_id() -> Option<String> {
    private_get_vm_id(None, None)
}

fn private_get_vm_id(
    custom_path: Option<&str>,
    mock_efi_path: Option<&str>,
) -> Option<String> {
```

libazureinit/src/status.rs
```diff
@@ -0,0 +1,337 @@
...
+    } else {
+        tracing::info!(target: "libazureinit::status::retrieved_vm_id", "VM ID (Gen2, no swap): {}", system_uuid);
+        Some(system_uuid)
+    }
+}
+
+/// This function checks whether a provisioning status file exists for the current VM ID.
+/// If the file exists, provisioning has already been completed and should be skipped.
+/// If the file does not exist or the VM ID has changed, provisioning should proceed.
+///
```
jeremycline, line 161: Since this is a public function, it'd be good to expand the documentation slightly to describe what happens if either of the optional parameters isn't provided; if `config` is None, the default provisioning directory is used to check if things are complete, and if no VM id is provided, it'll be read.

If the parameters exist to make testing easier, you can use the same trick at above.

jeremycline (commented):
It's looking pretty solid, and my notes are less relevant if the provisioning marking/checking APIs become private and part of the provision API down the road.

libazureinit/src/status.rs
peytonr18: For this one, the parameters do actually matter beyond just testing --
`config` is how we determine the provisioning_dir (ie: where we put the vm_id.provisioned file). 
`vm_id` we need to write the correct filename.

But, I'll update the documentation to make everything clearer!

peytonr18: In the spirit of keeping this all consistent, I did the same for `mark_provisioning_complete()` and split it into a private and public function.

libazureinit/src/config.rs
peytonr18: And, looking at the rest of the code too, are you okay with keeping the functions in status.rs `get_provisioning_dir` and `check_provisioning_dir` or does it make more sense to change them too?

--> `get_persistent_state_dir` and `check_persistent_state_dir`

libazureinit/src/status.rs
```diff
@@ -0,0 +1,385 @@
...
+        !Path::new(path).exists()
+    } else {
+        if Path::new("/sys/firmware/efi").exists()
+            || Path::new("/dev/efi").exists()
+        {
+            return false;
+        }
+        true
+    }
+}
```
cjp256, line 70: While this works, I would suggest a mocking strategy that avoids new code paths for mocks.   This inherently guarantees we can't get coverage over the actual potential code paths.  While this function is relatively simple, it can be problematic for more complex cases.

Consider something like:

```rust
fn is_vm_gen1(sysfs_efi_path: Option<&str>, dev_efi_path: Option<&str>) -> bool {
    let sysfs_efi_path = sysfs_efi_path.unwrap_or("/sys/firmware/efi");
    let dev_efi_path = dev_efi_path.unwrap_or("/dev/efi");
    !Path::new(sysfs_efi_path).exists() && !Path::new(dev_efi_path).exists()
}

#[cfg(test)]
mod tests {
    use super::*;
    use std::fs::{create_dir, remove_dir};

    #[test]
    fn test_gen1_vm() {
        // Mock paths that do not exist
        assert!(is_vm_gen1(Some("/nonexistent/efi"), Some("/nonexistent/dev_efi")));
    }

    #[test]
    fn test_gen2_vm_with_efi() {
        let mock_path = "/tmp/mock_efi";
        create_dir(mock_path).ok(); // Create a mock directory
        assert!(!is_vm_gen1(Some(mock_path), Some("/nonexistent/dev_efi")));
        remove_dir(mock_path).ok(); // Clean up
    }

    #[test]
    fn test_gen2_vm_with_dev_efi() {
        let mock_path = "/tmp/mock_dev_efi";
        create_dir(mock_path).ok(); // Create a mock directory
        assert!(!is_vm_gen1(Some("/nonexistent/efi"), Some(mock_path)));
        remove_dir(mock_path).ok(); // Clean up
    }
}
```

This approach allows us to pretty much get 100% coverage.

libazureinit/src/status.rs
```diff
@@ -0,0 +1,385 @@
...
+///
+/// # Returns
+/// - `Some(String)` containing the VM ID if retrieval is successful.
+/// - `None` if something fails or the output is empty.
+pub fn get_vm_id() -> Option<String> {
+    private_get_vm_id(None, None)
+}
+
+fn private_get_vm_id(
+    custom_path: Option<&str>,
```
cjp256, line 118: ah this is better.  just a nit on name:

```suggestion
    product_uuid_path: Option<&str>,
    dev_efi_path: Option<&str>,
```

libazureinit/src/status.rs
cjp256: This is a good example of where the optional parameter would be useful, desirable even. If the VM ID is known earlier it can be reused with this parameter without having to query it again.

The main entry point should query for VM ID and pass it to logging -> KVP.

And then it can re-use when querying for provisioning status.

This would avoid multiple calls to get_vm_id().

libazureinit/src/status.rs
cjp256: I believe this was addressed, resolving.

libazureinit/src/status.rs
peytonr18: After some discussion offline, I've pushed a refactor to streamline how we handle `vm_id`. Instead of calling `get_vm_id()` in multiple places, we now retrieve it once in `main.rs` and pass it explicitly where needed (logging, provisioning, status etc.).
 - `vm_id` is now passed from `main.rs` → `setup_layers()` → `EmitKVPLayer`, so `EmitKVPLayer` no longer fetches it itself.
 - `is_provisioning_complete()` and `mark_provisioning_complete()` now take `vm_id` as a parameter from `main.rs` instead of each calling `get_vm_id()` internally.
 - Since `vm_id` is always passed explicitly, we no longer need the separate private/public function split for `is_provisioning_complete()` and `mark_provisioning_complete()`.

This removes redundant lookups with `get_vm_id()`. Let me know if anything looks off with that or anyone has any questions!

jeremycline (approved):
All my concerns have been addressed, thanks!

anhvoms (approved):
LGTM! Thanks for the work and all the testing for this PR

PR #168, opened 2025-03-04
Cargo: fix clippy errors around litemap 0.7.5 and zerofrom 0.1.6

dongsupark:
Avoid clippy errors from MSRV requirement 1.81 of `litemap` 0.7.5 as well as `zerofrom` 0.1.6.
Needed because azure-init CI still requires Rust 1.76+.

```
error: package `litemap v0.7.5` cannot be built because it requires rustc 1.81 or newer, while the currently active rustc version is 1.76.0
error: package `zerofrom v0.1.6` cannot be built because it requires rustc 1.81 or newer, while the currently active rustc version is 1.76.0
```
## Testing done

Local test passed.
```
$ rustup default 1.76.0
$ cargo clippy --all-targets --all-features
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,10 +26,16 @@ chrono = "0.4"
 
 [dev-dependencies]
 assert_cmd = "2.0.16"
+# Avoid clippy errors from MSRV requirement 1.81 of litemap 0.7.5.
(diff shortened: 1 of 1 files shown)


dongsupark:
Duplicate of https://github.com/Azure/azure-init/pull/173

PR #170, opened 2025-03-05
Report failures to Azure via Wireserver with KVP Fallback 

### LINKED ISSUE #58: [RFE] Report failures to Azure when there's an unrecoverable error
## Current situation

Currently azure-init does not report any failure to Azure. If it can't finish provisioning, it will return with an error code. From a user perspective, provisioning will eventually fail with OS provisioning timeout due to Azure platform not receiving a provisioning complete signal. 

In many cases the user might not be able to access the VM if provisioning fails and as such, might have a very hard time figuring out why provisioning failed

## Ideal future situation

Have the azure-init report failures to Azure, which will then fail provisioning with a useful error message indicating why provisioning failed.

## **Implementation options
These are not two mutually exclusive options, but rather complimenting each other.
 
1) Use wireserver to report errors to the platform. [Here](https://github.com/canonical/cloud-init/blob/79ee656236b626090becbf816b0ff1e961f04960/cloudinit/sources/helpers/azure.py#L755) is how cloud-init is doing it. Essentially azure-init will need to construct a health report similar to reporting provisioning complete, but indicating the report status as NotReady, a substatus of ProvisioningFailed, and a meaningful description that will eventually show up as an error message back to the user. 
I would strongly encourage azure-init to follow the error messages used by [cloud-init](https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/azure/errors.py), because we have post-processing, monitoring, and alerting mechanism built around the errors returned by cloud-init.
A sample error returned by cloud-init

`result=error|reason=http error querying IMDS|agent=Cloud-Init/23.3.3-0ubuntu0~20.04.1|http_code=410|duration=300.2051315307617|'exception=UrlError(''410 Client Error: Gone for url: http://169.254.169.254/metadata/instance?api-version=2021-08-01&extended=true'')'|url=http://169.254.169.254/metadata/instance?api-version=2021-08-01&extended=true|vm_id=e76f68ac-04a8-4069-be7c-7f04b01f520f|timestamp=2024-03-12T09:39:16.373226|documentation_url=https://aka.ms/linuxprovisioningerror`

2) The failure reporting via wireserver only works if azure-init can establish communication to wireserver and can successfully post the error. In the cases where it's not working, the other option is to write a KVP with the error and Azure platform will process it.
See [cloud-init](https://github.com/canonical/cloud-init/blob/79ee656236b626090becbf816b0ff1e961f04960/cloudinit/sources/azure/kvp.py#L27) implementation as reference

peytonr18:
This PR introduces failure reporting for provisioning. The changes ensure that if provisioning fails, the error is reported to Azure either via wireserver or, as a fallback, via KVP .
 
## Key Changes
- **XML Payload for Failure Reporting:**
  - Implemented `build_report_failure_file` and `report_failure()` functions.
  - These functions construct an XML payload similar to cloud‑init, where:
    - `<State>` is set to `"NotReady"`,
    - `<Substatus>` is set to `"ProvisioningFailed"`, and
    - `<Description>` contains a user-visible error message.

- **Integration in the Provisioning Flow:**
  - In the `provision()` function, when a provisioning error occurs:
    - The current goalstate is fetched from wireserver.
    - `report_failure()` is called with the appropriate error description.
    - If the wireserver connection fails and the error can't be posted, the error is logged via `tracing::error!` as a fallback.

- **config report_failure:**
  - An added `report_failure` call where the `Config` is parsed in allows us to force-fail for testing purposes. This will later be changed with the addition of `runcmd` and the new endpoint, but it's the best option to test that functionality for now. 

Resolves #58.

### DIFF
--- a/libazureinit/src/goalstate.rs
+++ b/libazureinit/src/goalstate.rs
@@ -252,10 +252,85 @@ fn build_report_health_file(goalstate: Goalstate) -> String {
     )
 }
 
+#[instrument(err, skip_all)]
+pub async fn report_failure(
+    client: &Client,
+    goalstate: Goalstate,
+    description: &str,
+    retry_interval: Duration,
+    total_timeout: Duration,
+    url: Option<&str>,
+) -> Result<(), Error> {
+    let mut headers = HeaderMap::new();
+    headers.insert("x-ms-agent-name", HeaderValue::from_static("azure-init"));
+    headers.insert("x-ms-version", HeaderValue::from_static("2012-11-30"));
+    headers.insert(
+        "Content-Type",
+        HeaderValue::from_static("text/xml;charset=utf-8"),
+    );
+
+    let request_timeout =
+        Duration::from_secs(http::WIRESERVER_HTTP_TIMEOUT_SEC);
+    let url = url.unwrap_or(DEFAULT_HEALTH_URL);
+
+    let post_request = build_report_failure_file(goalstate, description);
+
+    _ = http::post(
+        client,
+        headers,
+        post_request,
+        request_timeout,
+        retry_interval,
+        total_timeout,
+        url,
+    )
+    .await?;
+
+    Ok(())
+}
+
+fn build_report_failure_file(
+    goalstate: Goalstate,
+    description: &str,
+) -> String {
+    let post_request = "<?xml version=\"1.0\" encoding=\"utf-8\"?>\n\
+    <Health xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" xmlns:xsd=\"http://www.w3.org/2001/XMLSchema\">\n\
+        <GoalStateIncarnation>$GOAL_STATE_INCARNATION</GoalStateIncarnation>\n\
+        <Container>\n\
+            <ContainerId>$CONTAINER_ID</ContainerId>\n\
+            <RoleInstanceList>\n\
+                <Role>\n\
+                    <InstanceId>$INSTANCE_ID</InstanceId>\n\
+                    <Health>\n\
+                        <State>NotReady</State>\n\
+                        <Substatus>ProvisioningFailed</Substatus>\n\
+                        <Description>$DESCRIPTION</Description>\n\
+                    </Health>\n\
+                </Role>\n\
+            </RoleInstanceList>\n\
+        </Container>\n\
+    </Health>";
+
+    let post_request =
+        post_request.replace("$GOAL_STATE_INCARNATION", &goalstate.incarnation);
+    let post_request = post_request
+        .replace("$CONTAINER_ID", &goalstate.container.container_id);
+    let post_request = post_request.replace(
+        "$INSTANCE_ID",
+        &goalstate
+            .container
+            .role_instance_list
+            .role_instance
+            .instance_id,
+    );
+    post_request.replace("$DESCRIPTION", description)
+}
+
 #[cfg(test)]
 mod tests {
     use super::{
-        build_report_health_file, get_goalstate, report_health, Goalstate,
+        build_report_failure_file, build_report_health_file, get_goalstate,
+        report_health, Goalstate,
     };
 
     use reqwest::{header, Client, StatusCode};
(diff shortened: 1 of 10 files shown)


cjp256:
There is a new endpoint which does not require us to fetch goal state and uses json instead of xml.

sudo curl -X POST -H 'x-ms-guest-agent-name: azure-init/some-version' \
--header 'Content-Type: application/json' \
--data '{
    "state": "NotReady",
    "details": {
        "subStatus":  "ProvisioningFailed",
        "description": "The VM encountered an error during deployment."
    }
}' \
http://168.63.129.16/provisioning/health 

Example provisoning complete report without description
{
    "state": "Ready",
}

Example provisoning failure report
{
    "state": "NotReady",
    "details": {
        "subStatus":  "ProvisioningFailed",
        "description": "The provisioning client has something to report."
    }
}

Example provisioning in-progress report
{
    "state": "NotReady",
    "details": {

        "subStatus":  "Provisioning",
        "description": "The provisioning client has something to report."
    }
}

Response Code | Meaning
-- | --
201 | success
429 | too many requests. Rate limted
503 | GoalState is not ready yet, try again

We don't have to do it in this PR, but I would like to switch over to the new endpoint to avoid dealing with goalstate.

The one upside to fetching goalstate is getting the container id.

peytonr18:
Soon, we won't need goalstate at all and I can get rid of the CONFIG_HEALTH_MESSAGE and just be able to report_failure. But, that endpoint work will be in a separate PR so the current implementation is just a placeholder in the meantime.

src/kvp.rs
```diff
@@ -494,6 +563,76 @@ fn get_uptime() -> Duration {
...
+/// # Parameters
+/// - `status`: Specifies whether this is a success or failure report.
+/// - `reason`: For failure reports, the failure reason. Must be provided for failures.
+/// - `extra`: Optional additional key=value pairs.
+/// - `vm_id`: The VM ID.
+/// - `timestamp`: The precomputed timestamp in ISO 8601 format.
+///
+/// # Returns
+/// A tuple of (key, value) where key is fixed as "PROVISIONING_REPORT".
+fn build_health_kvp(
```
cjp256, line 600: I think it would be best to split this into two interfaces.

Avoid the need for HealthReportKind and split into report_success_kvp() / report_failure_kvp().  Is there value in keeping these together?

libazureinit/src/goalstate.rs
```diff
@@ -252,10 +252,85 @@ fn build_report_health_file(goalstate: Goalstate) -> String {
...
+pub async fn report_failure(
+    client: &Client,
+    goalstate: Goalstate,
+    description: &str,
+    retry_interval: Duration,
+    total_timeout: Duration,
+    url: Option<&str>,
+) -> Result<(), Error> {
+    let mut headers = HeaderMap::new();
+    headers.insert("x-ms-agent-name", HeaderValue::from_static("azure-init"));
```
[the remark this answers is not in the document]
cjp256: i see it below

let agent = format!("Azure-Init/{}", env!("CARGO_PKG_VERSION"));

maybe this should be made into a function somewhere

src/kvp.rs
cjp256: while using types, this is very much bool like.  if the boolean isn't trivial change in a moethod, I would consider making two methods which can each do their own thing with correct interfaces.  easier to read, and easier to maintain over time in my opinion.

libazureinit/src/goalstate.rs
```diff
@@ -252,10 +252,85 @@ fn build_report_health_file(goalstate: Goalstate) -> String {
     )
 }
 
+#[instrument(err, skip_all)]
+pub async fn report_failure(
```
cjp256, line 256: from a high level, goalstate is a dependency for health (for the goalstate incarnation id), but otherwise unrelated.  i would suggest moving this logic into health.rs

so we would have something like:

health::report_ready()
health::report_failure()

maybe someday:
heath::report_progress()

report ready is already here so do what you will, just a thought.

libazureinit/src/goalstate.rs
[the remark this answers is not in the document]
cjp256: i also see

 let mut default_headers = header::HeaderMap::new();
    let user_agent = if cfg!(debug_assertions) {
        format!("azure-init v{}-{}", VERSION, COMMIT_HASH)
    } else {
        format!("azure-init v{}", VERSION)
    };
    let user_agent = header::HeaderValue::from_str(user_agent.as_str())?;

PR #177, opened 2025-03-17
Output `azure-init.log` to Configurable Location

### LINKED ISSUE #64: [RFE] Output log to a known (perhaps configurable) location
All logs from azure-init currently are sent to console/journal log. We should channel logs to a file (INFO level logs can go to both file and console) so that it's easier to investigate issue should we end up having a complex issue.

Log can default to /var/log/azure-init.log. The exact file path should be configurable (once we support config).

peytonr18:
This PR makes the `azure-init log` path configurable via the `Config` struct while ensuring logging is properly initialized before configuration is loaded. 

To achieve this, the logging setup now follows a two-phase approach:
- Phase 1 (Minimal Logging Pre-Config)
 -- Initializes tracing with console (`stderr`), KVP (Hyper-V), and OpenTelemetry logging.
-- This ensures that logs from early configuration loading are captured.
- Phase 2 (Full Logging with Config)
-- Once `Config` is available, logging is reconfigured to also write to the configured `telemetry_log_path.`
-- The file path is taken from `config.telemetry_log_path`, falling back to `DEFAULT_TELEMETRY_LOG_PATH` if unspecified.

Resolves #64.

### DIFF
--- a/libazureinit/src/config.rs
+++ b/libazureinit/src/config.rs
@@ -252,23 +252,23 @@ impl Default for AzureInitDataDir {
 }
 
 /// The default directory for azure-init.log
-pub const DEFAULT_TELEMETRY_LOG_PATH: &str = "/var/log/azure-init.log";
+pub const DEFAULT_AZURE_INIT_LOG_PATH: &str = "/var/log/azure-init.log";
 /// Telemetry log (azure-init.log) struct.
 /// Configures settings for where azure-init should channel telemetry logs.
-/// If no custom path is provided, `TelemetryLogPath::default()` uses
-/// [`DEFAULT_TELEMETRY_LOG_PATH`].
+/// If no custom path is provided, `AzureInitLogPath ::default()` uses
+/// [`DEFAULT_AZURE_INIT_LOG_PATH`].
 #[derive(Serialize, Deserialize, Debug, Clone)]
 #[serde(default)]
-pub struct TelemetryLogPath {
+pub struct AzureInitLogPath {
     /// Specifies the path used to capture all telemetry logs.
     /// Defaults to `/var/log/azure-init.log`.
     pub path: PathBuf,
 }
 
-impl Default for TelemetryLogPath {
+impl Default for AzureInitLogPath {
     fn default() -> Self {
         Self {
-            path: PathBuf::from(DEFAULT_TELEMETRY_LOG_PATH),
+            path: PathBuf::from(DEFAULT_AZURE_INIT_LOG_PATH),
         }
     }
 }
@@ -290,7 +290,7 @@ pub struct Config {
     pub wireserver: Wireserver,
     pub telemetry: Telemetry,
     pub azure_init_data_dir: AzureInitDataDir,
-    pub telemetry_log_path: TelemetryLogPath,
+    pub azure_init_log_path: AzureInitLogPath,
 }
 
 /// Implements `Display` for `Config`, formatting it as a readable TOML string.
@@ -654,7 +654,7 @@ mod tests {
         );
 
         assert_eq!(
-            config.telemetry_log_path.path.to_str().unwrap(),
+            config.azure_init_log_path.path.to_str().unwrap(),
             "/var/log/azure-init.log"
         );
 
@@ -693,7 +693,7 @@ mod tests {
         kvp_diagnostics = false
         [azure_init_data_dir]
         path = "/custom/azure-init-data-dir"
-        [telemetry_log_path]
+        [azure_init_log_path]
         path = "/custom/path/azure-init.log"
         "#
         )?;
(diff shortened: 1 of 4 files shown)


libazureinit/src/config.rs
```diff
@@ -251,6 +251,28 @@ impl Default for AzureInitDataDir {
     }
 }
 
+/// The default directory for azure-init.log
+pub const DEFAULT_TELEMETRY_LOG_PATH: &str = "/var/log/azure-init.log";
```
anhvoms, line 255:
```suggestion
pub const DEFAULT_AZURE_INIT_LOG_PATH: &str = "/var/log/azure-init.log";
```

anhvoms: I would recommend replacing all the TelemetryLog with just Log (e.g., TelemetryLogPath -> LogPath)

libazureinit/src/config.rs
```diff
@@ -268,6 +290,7 @@ pub struct Config {
     pub wireserver: Wireserver,
     pub telemetry: Telemetry,
     pub azure_init_data_dir: AzureInitDataDir,
+    pub telemetry_log_path: TelemetryLogPath,
```
anhvoms, line 293: Maybe keep the naming convention consistent and use AzureInitLogPath ?

src/logging.rs
```diff
@@ -21,9 +24,19 @@ pub fn initialize_tracing() -> sdktrace::Tracer {
     provider.tracer("azure-kvp")
 }
 
+/// Builds a `tracing` subscriber that can optionally write azure-init.log to a specific location if `Some(&Config)` is provided.
```
anhvoms, line 27: long line formatting

src/logging.rs
```diff
@@ -60,10 +73,58 @@ pub fn setup_layers(
...
         .with_filter(EnvFilter::from_env("AZURE_INIT_LOG"));
 
+    let log_path = config
+        .map(|cfg| cfg.telemetry_log_path.path.clone())
+        .unwrap_or_else(|| PathBuf::from(DEFAULT_TELEMETRY_LOG_PATH));
+
+    let file_layer = match OpenOptions::new()
+        .create(true)
+        .append(true)
+        .open(&log_path)
```
anhvoms, line 83: One thing that we didn't discuss is the permission for azure-init datadir and azure-init logpath? Using cloud-init as a reference, I think different distros have taken different opinions on this, with Ubuntu's cloud-init.log being visible to all users while Suse has limited cloud-init.log access to only root user?

@cjp256 @jeremycline to comment

src/logging.rs
```diff
@@ -60,10 +73,58 @@ pub fn setup_layers(
...
+                .with_span_events(FmtSpan::NEW | FmtSpan::CLOSE)
+                .with_writer(file)
+                .with_filter(EnvFilter::from_env("AZURE_INIT_LOG")),
+        ),
+        Err(e) => {
+            event!(
+                Level::WARN,
+                "Could not open {}: {}. Falling back to default log path.",
+                log_path.display(),
+                e
```
anhvoms, line 96:
```suggestion
                "Could not open {}: {}. Falling back to default log path {}.",
                log_path.display(),
                e,
                DEFAULT_TELEMETRY_LOG_PATH
```

anhvoms (changes requested):

src/logging.rs
cjp256: would be best to ensure chmod 600 for that reason.

cloud-init.log is chmod 640, with group owner "adm" which gives a little more balance as any user with "adm" privileges can see it too.

libazureinit/src/config.rs
```diff
@@ -251,6 +251,28 @@ impl Default for AzureInitDataDir {
     }
 }
 
+/// The default directory for azure-init.log
+pub const DEFAULT_AZURE_INIT_LOG_PATH: &str = "/var/log/azure-init.log";
```
cjp256, line 255:
```suggestion
pub const DEFAULT_AZURE_INIT_LOG_PATH: &str = "/var/log/azure-init.log";

```

src/logging.rs
```diff
@@ -60,10 +75,84 @@ pub fn setup_layers(
...
+                fmt::layer()
+                    .with_span_events(FmtSpan::NEW | FmtSpan::CLOSE)
+                    .with_writer(file)
+                    .with_filter(EnvFilter::from_env("AZURE_INIT_LOG")),
+            )
+        }
+        Err(e) => {
+            event!(
+                Level::ERROR,
+                "Could not open {}: {}. Falling back to default log path {}.",
```
cjp256, line 108: I would suggest we simply emit this failure to syslog and not change configured file location.  In the future we can make sure critical events are available in syslog too.

peytonr18: Removed it! We don't have a specific syslog layer yet for tracing -- but that's something I can address in a future PR! For now, I just logged it via stderr/kvp.

anhvoms (approved):

anhvoms (approved):

PR #179, opened 2025-03-18
Allow for Enabling/Disabling KVP using Config

### LINKED ISSUE #152: Allow for Enabling/Disabling KVP using Config Struct
Issue for amending KVP functionality so that a user can enable/disable kvp telemetry emission via the config struct. 

Current idea is to pass the config to the subscriber setup and create the emit_kvp_layer only if kvp_diagnostics flag is enabled. Once we have KVP functionality, it can be amended to allow this. 

Related to PR #133.

peytonr18:
This PR introduces the ability to enable or disable KVP logging through the `telemetry.kvp_diagnostics` field in the configuration.

### Changes 
- The `setup_layers` function now checks `config.telemetry.kvp_diagnostics` before initializing `emit_kvp_layer`.
- If `kvp_diagnostics = false`, `emit_kvp_layer` is set to `None`, ensuring no logs are written to the EmitKVPStruct.
- If `config` is `None`, the default behavior (`true`) is maintained. 
- Added a unit test in `kvp.rs` to validate that no data is written when KVP is disabled.

### Key Point: Config Tracing & KVP Behavior
- Before the config is loaded, `setup_layers` is called once with `None`, meaning the config values will still be written to KVP. If `kvp_diagnostics = false`, KVP logging will stop after the second call to `setup_layers`, since the new subscriber overrides the previous one.

### Dependency on `probertson-cache-vm-id-logging-new`
This PR is based on `probertson-cache-vm-id-logging-new`, as that branch introduces passing `Config` into `setup_layers`, which is required for this logic.

### Next Steps
- Merge `probertson-cache-vm-id-logging-new` first.
- Then merge this PR to introduce configurable KVP diagnostics.

Resolves #152.

### DIFF
--- a/libazureinit/src/config.rs
+++ b/libazureinit/src/config.rs
@@ -251,6 +251,29 @@ impl Default for AzureInitDataDir {
     }
 }
 
+/// The default directory for azure-init.log
+pub const DEFAULT_AZURE_INIT_LOG_PATH: &str = "/var/log/azure-init.log";
+
+/// Telemetry log (azure-init.log) struct.
+/// Configures settings for where azure-init should channel telemetry logs.
+/// If no custom path is provided, `AzureInitLogPath ::default()` uses
+/// [`DEFAULT_AZURE_INIT_LOG_PATH`].
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(default)]
+pub struct AzureInitLogPath {
+    /// Specifies the path used to capture all telemetry logs.
+    /// Defaults to `/var/log/azure-init.log`.
+    pub path: PathBuf,
+}
+
+impl Default for AzureInitLogPath {
+    fn default() -> Self {
+        Self {
+            path: PathBuf::from(DEFAULT_AZURE_INIT_LOG_PATH),
+        }
+    }
+}
+
 /// General configuration struct for azure-init.
 ///
 /// Aggregates all configuration settings for managing SSH, provisioning, IMDS, media,
@@ -268,6 +291,7 @@ pub struct Config {
     pub wireserver: Wireserver,
     pub telemetry: Telemetry,
     pub azure_init_data_dir: AzureInitDataDir,
+    pub azure_init_log_path: AzureInitLogPath,
 }
 
 /// Implements `Display` for `Config`, formatting it as a readable TOML string.
@@ -630,6 +654,11 @@ mod tests {
             "/var/lib/azure-init/",
         );
 
+        assert_eq!(
+            config.azure_init_log_path.path.to_str().unwrap(),
+            "/var/log/azure-init.log"
+        );
+
         tracing::debug!("test_empty_config_file_uses_defaults_when_merged completed successfully.");
 
         Ok(())
@@ -665,6 +694,8 @@ mod tests {
         kvp_diagnostics = false
         [azure_init_data_dir]
         path = "/custom/azure-init-data-dir"
+        [azure_init_log_path]
+        path = "/custom/path/azure-init.log"
         "#
         )?;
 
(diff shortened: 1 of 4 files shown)


libazureinit/src/config.rs
```diff
@@ -251,6 +251,28 @@ impl Default for AzureInitDataDir {
     }
 }
 
+/// The default directory for azure-init.log
+pub const DEFAULT_AZURE_INIT_LOG_PATH: &str = "/var/log/azure-init.log";
+/// Telemetry log (azure-init.log) struct.
```
cjp256, line 256:
```suggestion

/// Telemetry log (azure-init.log) struct.
```

src/logging.rs
```diff
@@ -44,26 +59,112 @@ pub fn setup_layers(
...
+            Ok(layer) => Some(layer.with_filter(kvp_filter)),
+            Err(e) => {
+                event!(Level::ERROR, "Failed to initialize EmitKVPLayer: {}. Continuing without KVP logging.", e);
+                None
+            }
         }
+    } else {
+        event!(
+            Level::INFO,
+            "KVP diagnostics are disabled via config. Skipping KVP logging."
```
cjp256, line 80:
```suggestion
            "Hyper-V KVP diagnostics are disabled via config.  It is recommended to be enabled for support purposes."
```

src/logging.rs
```diff
@@ -44,26 +59,112 @@ pub fn setup_layers(
...
-        Err(e) => {
-            event!(Level::ERROR, "Failed to initialize EmitKVPLayer: {}. Continuing without KVP logging.", e);
-            None
+    let kvp_enabled = config
+        .map(|cfg| cfg.telemetry.kvp_diagnostics)
+        .unwrap_or(true);
+
+    let emit_kvp_layer = if kvp_enabled {
+        match EmitKVPLayer::new(
+            std::path::PathBuf::from("/var/lib/hyperv/.kvp_pool_1"),
```
cjp256, line 68: Slightly unrelated, but this reminds me we should update our service config to ensure kvpd is started:

After=hypervkvpd.service   (RHEL)
After=hv-kvp-daemon.service  (Ubuntu)

src/logging.rs
```diff
@@ -44,26 +59,112 @@ pub fn setup_layers(
...
+
+            Some(
+                fmt::layer()
+                    .with_span_events(FmtSpan::NEW | FmtSpan::CLOSE)
+                    .with_writer(file)
+                    .with_filter(EnvFilter::from_env("AZURE_INIT_LOG")),
+            )
+        }
+        Err(e) => {
+            event!(
```
cjp256, line 118: I think it would be best to just not log to the file in this case.  Maybe just make sure we emit important events via syslog?

cjp256: Or at least this failure into syslog.

src/logging.rs
peytonr18: Thank you!! Addressed in #180

anhvoms (approved):

libazureinit/src/config.rs
```diff
@@ -251,6 +251,29 @@ impl Default for AzureInitDataDir {
     }
 }
 
+/// The default directory for azure-init.log
+pub const DEFAULT_AZURE_INIT_LOG_PATH: &str = "/var/log/azure-init.log";
+
+/// Telemetry log (azure-init.log) struct.
+/// Configures settings for where azure-init should channel telemetry logs.
+/// If no custom path is provided, `AzureInitLogPath ::default()` uses
```
dongsupark, line 259:
```suggestion
/// If no custom path is provided, `AzureInitLogPath::default()` uses
```

dongsupark (approved):

cjp256 (approved):

PR #180, opened 2025-03-18
feat(azure-init.service): ensure azure-init wants and starts after kvpd

cjp256:
Ensure kvpd is started before azure-init and want it.

hypervkvpd.service is used by Fedora, RHEL, etc.
hv-kvp-daemon.service is used by Debian, Ubuntu, etc.

### DIFF
--- a/config/azure-init.service
+++ b/config/azure-init.service
@@ -1,5 +1,7 @@
 [Unit]
 Description=Azure-Init
+After=hypervkvpd.service hv-kvp-daemon.service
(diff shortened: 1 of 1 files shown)


peytonr18 (approved):
LGTM!

cjp256:
Need to run some full tests on this first.

KsenijaS:
> Need to run some full tests on this first.

Could we have a test that cover this scenario?

anhvoms (approved):

PR #187, opened 2025-04-03
Adding `runcmd` and `write_files` functionality

peytonr18:
This PR introduces two new features: `runcmd` and `write_files`.

- **runcmd**: Allows users to provide a list of commands (with arguments) that are executed in sequence during provisioning. This feature aims to simplify the process of running script-like instructions without requiring custom scripts or manual interventions.

- **write_files**: Enables the creation or modification of files on the provisioned machine. Users can specify file paths, contents, permissions, and owner/group settings within their configuration.

### Key Changes
1. **Schema Updates**: The config schema now includes sections for `runcmd` and `write_files`.  Additionally, the config parser now requires the `#azure-init-config` header at the top of any loaded file, validating it before merging config files or user-data. This is similar to what cloud-init does with `#cloud-config`.
3. **Execution Flow**: The provisioning logic has been updated to parse and execute commands from `runcmd`.  
4. **File Handling**: `write_files` ensures that each specified file is created with the correct contents, permissions, and ownership.  

### Additional Notes
- Relies on #188 , which contains the changes for report_failure  report_success logic that inspired adding `runcmd` and `write_files` for the sake of being able to force-fail provisioning during testing.


src/main.rs
```diff
@@ -186,68 +353,150 @@ async fn provision(
...
+    // If an error happens, capture the goalstate and call report_failure()
+    let provisioning_result: Result<(), anyhow::Error> = async {
+        // Username can be obtained either via fetching instance metadata from IMDS
+        // or mounting a local device for OVF environment file. It should not fail
+        // immediately in a single failure, instead it should fall back to the other
+        // mechanism. So it is not a good idea to use `?` for query() or
+        // get_environment().
+        let instance_metadata = imds::query(
+            &client,
+            Duration::from_secs(imds_http_retry_interval_sec),
```
KsenijaS, line 366: I wonder why we are passing timeout and retry interval as arguments to this function? I think those values should be constants or have a default value.

src/main.rs
```diff
@@ -186,68 +353,150 @@ async fn provision(
...
+                target: "azure_init",
+                health_report = "success",
+                "Provisioning completed successfully"
+            );
+        }
+        Err(e) => {
+            tracing::error!("Provisioning failed with error: {:?}", e);
+            // Report the provisioning failure via wireserver.
+            // If this fails, fallback to KVP reporting by logging the error.
+            if let Ok(vm_goalstate) = goalstate::get_goalstate(
```
KsenijaS, line 469: same comment here as for imds::query fn

libazureinit/src/goalstate.rs
```diff
@@ -252,10 +252,85 @@ fn build_report_health_file(goalstate: Goalstate) -> String {
     )
 }
 
+#[instrument(err, skip_all)]
+pub async fn report_failure(
+    client: &Client,
+    goalstate: Goalstate,
+    description: &str,
+    retry_interval: Duration,
+    total_timeout: Duration,
```
KsenijaS, line 261: can retry_interval and total_timeout be optional arguments?

peytonr18: I think we need retry_interval and total_timeout here so we can control the retry logic around posting to the wireserver. We use these to determine how often to retry and when to give up and fail provisioning if we can’t post goalstate (similar to how report_health works).

That said, we actually have a Wireserver config struct in config.rs with fields like `connection_timeout_secs` and `total_retry_timeout_secs`, and I’m not sure I remembered to utilize those when I merged in the config functionality (and we could use those here, I think).

src/main.rs
peytonr18: I didn’t write this function originally, I just moved it so it gets flagged as being "new" -- but my understanding is that the retry interval and total timeout are passed so we know how long to query IMDS and how often to retry until failure. 

In terms of making these constants, I agree with you -- that would be better. We do already define an IMDS config struct in config.rs with `connection_timeout_secs`, `read_timeout_secs`, and `total_retry_timeout_secs`, and I’m now realizing we might not be using those here like we should be... If that's the case, I can update this to pull from the config instead of being set manually.

libazureinit/src/config.rs
```diff
@@ -696,6 +801,19 @@ mod tests {
...
+        [runcmd]
+        commands = [
+            ["echo", "hello"],
+            ["apt-get", "update"]
+        ]
+        [[write_files]]
+        path = "/etc/custom-write-files"
+        owner = "test-user"
+        group = "test-user"
+        permissions = 384
```
cjp256, line 813: i assume this == 0o600, but also wonder if we should restrict it to one format to ensure correctness.

libazureinit/src/config.rs
```diff
@@ -345,6 +403,7 @@ impl Config {
 
         if base_path.exists() {
             tracing::info!(path=?base_path, "Loading base configuration file");
+            check_azure_init_header(&base_path)?;
```
cjp256, line 406: we only want this header in config coming from user-data / custom-data

local config files are azure-init configuration by definition.  it's the user data we want to make sure it meant for azure-init before attempting to load it

libazureinit/src/config.rs
```diff
@@ -421,6 +482,40 @@ impl Config {
...
+    })?;
+    let mut reader = std::io::BufReader::new(file);
+
+    let mut first_line = String::new();
+    let _ = reader.read_line(&mut first_line).map_err(|e| {
+        tracing::error!("Error reading first line of {:?}: {:?}", file_path, e);
+        Error::Io(e)
+    })?;
+
+    if !first_line.trim_start().starts_with("#azure-init-config") {
```
cjp256, line 505: i would suggest allowing any whitespace in between `#` and `aure-init-config` because I bet many will try `# azure-init-config`

cjp256: trim the end too.

libazureinit/src/config.rs
cjp256: because I would probably state permissions = 600 and be wrong.

src/main.rs
peytonr18: Resolved in #188.

libazureinit/src/config.rs
```diff
@@ -274,6 +287,61 @@ impl Default for AzureInitLogPath {
...
+}
+
+// Represents an instruction to write a file during provisioning.
+///
+/// # Fields
+/// - `path`: The destination path of the file.
+/// - `owner`: The username (e.g. "root") that will own the file. Defaults to `root`.`
+/// - `group`: The group name (e.g. "root") that will own the file. Defaults to `root`.
+/// - `permissions`: Numeric file permissions in octal (e.g. 0o600). Defaults to `o600`.
+/// - `content`: Raw file content in bytes (could be binary).
```
anhvoms, line 316: Perhaps not in this PR but we should offer the ability to write_files with format b64 or b64+gz, this will allow users to write files with characters that aren't human-readable

libazureinit/src/config.rs
```diff
@@ -274,6 +287,61 @@ impl Default for AzureInitLogPath {
...
+/// - `overwrite`: If `true`, overwrite existing files; otherwise skip if file already exists. Defaults to `true`.
```
anhvoms, line 317: Please add some examples here - or we should have some examples in the doc page

libazureinit/src/config.rs
```diff
@@ -421,6 +492,42 @@ impl Config {
...
+    let mut first_line = String::new();
+    let _ = reader.read_line(&mut first_line).map_err(|e| {
+        tracing::error!("Error reading first line of {:?}: {:?}", file_path, e);
+        Error::Io(e)
+    })?;
+
+    let trimmed = first_line.trim();
+    let re = Regex::new(r"^#\s*azure-init-config$").unwrap();
+    if !re.is_match(trimmed) {
+        tracing::error!(
```
anhvoms, line 518: This should not be an error but just an info. user-data can be used for many purposes and the customers don't necessarily want azure-init to handle it.

libazureinit/src/config.rs
```diff
@@ -421,6 +492,42 @@ impl Config {
...
+            "File {:?} missing proper header. Expected 'azure-init-config', found '{}'",
```
anhvoms, line 519: This should not be an error but just an info. user-data can be used for many purposes and the customers don't necessarily want azure-init to handle it.

libazureinit/src/config.rs
```diff
@@ -274,6 +287,61 @@ impl Default for AzureInitLogPath {
...
+/// Each command is specified as an array of string arguments,
+/// passed directly to the underlying executable. For example:
+/// ```yaml
+/// runcmd:
+///   commands:
+///     - ["echo", "hello", "world"]
+///     - ["apt-get", "update"]
+///     - ["bash", "-c", "echo 'Done!' && exit 0"]
+/// ```
+#[derive(Serialize, Deserialize, Debug, Clone, Default)]
```
anhvoms, line 301: Should we have a flag `fail_provisioning_upon_failure`? It should default to `True` but can be configured by the user

anhvoms: Same comment for write_files

anhvoms (changes requested):

libazureinit/src/config.rs
peytonr18: Went ahead and did it in this PR, we should support both of those formats now!

libazureinit/src/config.rs
cjp256: I personally would opt for keeping it simple in not configurable.  The config can be adjusted to avoid returning failure.   And allowing it just opens the door for unexpected behavior

peytonr18: @anhvoms Thoughts on this? Think this is the last big change to make/keep for this PR

anhvoms: What Chris mentioned works for me. The user indeed can change the command to not return failure

libazureinit/src/config.rs
```diff
@@ -274,6 +287,106 @@ impl Default for AzureInitLogPath {
...
+/// overwrite = true
+/// create_parents = true
+/// fail_provisioning_upon_failure = true
+///
+/// [[write_files]]
+/// path = "/etc/encoded.txt"
+/// owner = "test-user"
+/// group = "test-user"
+/// permissions = 0o600
+/// content = [97, 71, 86, 115, 98, 71, 56, 61]  # "aGVsbG8=" which is "hello"
```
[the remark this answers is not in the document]
peytonr18: Agreed wholeheartedly and addressed!

src/main.rs
```diff
@@ -100,6 +110,219 @@ fn get_username(
...
+    Ok(())
+}
+
+/// Applies a list of `WriteFiles` specifications, creating or overwriting each file
+/// according to its config. If `fail_provisioning_upon_failure` is `true` for a file
+/// and an error occurs, provisioning will fail. Otherwise, the error is logged and processing continues.
+///
+/// Returns a string describing the last successfully processed file (if any), or an error if one occurs
+/// with `fail_provisioning_upon_failure = true`.
+fn apply_write_files(
```
anhvoms, line 139: I would recommend organizing these methods into a different module and not stuff them into main.rs. This will allow future improvements to be limited to the module and easier to review. Perhaps `write_file.rs` and `runcmd.rs`

src/main.rs
```diff
@@ -100,6 +110,219 @@ fn get_username(
...
+    let mut builder = DirBuilder::new();
+    builder
+        .mode(mode)
+        .create(path)
+        .with_context(|| format!("Failed to create directory {:?}", path))?;
+    Ok(())
+}
+
+/// Applies a list of `WriteFiles` specifications, creating or overwriting each file
+/// according to its config. If `fail_provisioning_upon_failure` is `true` for a file
```
anhvoms, line 134: Please update doc if `fail_provisioning_upon_failure` is no longer an option

src/main.rs
```diff
@@ -100,6 +110,219 @@ fn get_username(
...
+                tracing::info!(
+                    "Successfully processed write_files entry: {:?}",
+                    written_path
+                );
+                last_path = written_path;
+            }
+            Err(e) => {
+                if wf.fail_provisioning_upon_failure {
+                    tracing::error!(
+                        "Aborting provisioning: write_files entry {:?} failed and `fail_provisioning_upon_failure` is true",
```
anhvoms, line 161: update message if `fail_provisioning_upon_failure` is no longer an option

src/main.rs
```diff
@@ -100,6 +110,219 @@ fn get_username(
...
+                        wf.path
+                    );
+                    return Err(e);
+                } else {
+                    tracing::warn!(
+                        "write_files entry {:?} failed, but continuing because `fail_provisioning_upon_failure` is false: {:?}",
```
anhvoms, line 167: same as above, update message

libazureinit/src/config.rs
```diff
@@ -274,6 +287,106 @@ impl Default for AzureInitLogPath {
...
+///
+/// # Fields
+/// - `path`: The destination path of the file.
+/// - `owner`: The username (e.g. "root") that will own the file. Defaults to `root`.`
+/// - `group`: The group name (e.g. "root") that will own the file. Defaults to `root`.
+/// - `permissions`: Numeric file permissions in octal (e.g. 0o600). Defaults to `o600`.
+/// - `content`: Raw file content as a string (e.g., `"hello"` or base64-encoded).
+/// - `encoding`: Optional encoding for the content. Supported values:
+///     - `"b64"`: Base64-encoded content.
+///     - `"b64+gz"`: Base64-encoded and gzipped content.
```
anhvoms, line 332: This should be `gz+b64` to be consistent with typical usage (as in, the content shall be gzipped, then base64 encoded).

libazureinit/src/config.rs
```diff
@@ -878,6 +1097,22 @@ mod tests {
         path = "/cli-override-azure-init-data-dir"
         [azure_init_log_path]
         path = "/custom/path/azure-init.log"
+        [runcmd]
+        commands = [
+            ["echo", "hello"],
+            ["apt-get", "update"]
+        ]
+        fail_provisioning_upon_failure = false
```
anhvoms, line 1105: can we add a sample and test case for `gz+b64`? Also add a test for multi-line content

anhvoms (changes requested):

src/main.rs
peytonr18: I've kept `fail_provisioning_upon_failure` in because I didn't think we had made a decision on keeping it or not keeping it.

jeremycline (changes requested):
Regarding the `fail_provisioning_upon_failure` option, which there seems to be some back-and-forth on: my take is I don't love it as a single boolean for a whole set of commands since that doesn't really grant me the flexibility I might need for _some_ commands to fail.

If the approach to run a set of complicated commands with error handling is expected to be "use write_files to add this shell script and then execute it" then maybe runcmd should always require commands to return zero to be considered a success. I see that files are written _after_ running commands, should we consider writing files out first?

Alternatively, each command could be a dictionary that includes an optional return code key (with 0 as the default) so users can say "I expect this command to return X". Sometimes commands return non-zero with particular codes to mean things other than failure (whether they _should_ is a separate discussion, of course). Of course, some terrible commands might return _multiple_ potential "acceptable" status codes, so this doesn't cover everything.

cjp256:
To be a bit of a pain, I'm asking to split out write_files functionality into its own PR then come back to runcmd.  There's a lot going on here and we should break it up.

Issue #186, opened 2025-04-01
Integrate New Status Endpoint 


peytonr18:
There is a new endpoint which does not require us to fetch goal state and uses json instead of xml.

```
sudo curl -X POST -H 'x-ms-guest-agent-name: azure-init/some-version' \
--header 'Content-Type: application/json' \
--data '{
    "state": "NotReady",
    "details": {
        "subStatus":  "ProvisioningFailed",
        "description": "The VM encountered an error during deployment."
    }
}' \
http://168.63.129.16/provisioning/health 

Example provisoning complete report without description
{
    "state": "Ready",
}

Example provisoning failure report
{
    "state": "NotReady",
    "details": {
        "subStatus":  "ProvisioningFailed",
        "description": "The provisioning client has something to report."
    }
}

Example provisioning in-progress report
{
    "state": "NotReady",
    "details": {

        "subStatus":  "Provisioning",
        "description": "The provisioning client has something to report."
    }
}
```

Response Code | Meaning
-- | --
201 | success
429 | too many requests. Rate limted
503 | GoalState is not ready yet, try again

We don't have to do it in this PR, but I would like to switch over to the new endpoint to avoid dealing with goalstate.

The one upside to fetching goalstate is getting the container id.

_Originally posted by @cjp256 in https://github.com/Azure/azure-init/issues/170#issuecomment-2707411674_

peytonr18:
Actively working on this, as part of the broader report_failure work.

PR #188, opened 2025-04-09
Introduce JSON-Based Provisioning Health Reporting and Reporting Failures to Wireserver/Azure

### LINKED ISSUE #58: [RFE] Report failures to Azure when there's an unrecoverable error

peytonr18:
This PR introduces failure reporting for provisioning and does so using a new JSON endpoint that does not rely on goalstate. The changes ensure that the provisioning status is reported to wireserver or, as a fallback, via KVP. Additionally, this utilizes a new JSON-based endpoint for reporting, rather than the old XML payload. 

## Key Changes
- **Removal of Legacy XML/Goalstate Logic:**
  - The old XML payload format and its associated goalstate logic have been completely removed.
  - The new endpoint accepts JSON and no longer requires goalstate information.

- **New Public APIs: `report_ready` & `report_failure`:**
  - Both build their own reqwest::Client using time‑outs from config.wireserver.
  - The shared `_report` helper handles JSON payload construction, headers, and the retry loop under the hood.

- **Integration Changes in Provisioning Flow:**
  - In `main()`, we now call `report_ready()` or `report_failure()` to send the health status on success or failure.
  - Early failure paths (e.g., a configuration load error) now use the new JSON health reporting rather than an XML blob (as was the case in #170). 

- **Unify Error Handling and Reporting:**
  - The existing `Error` enum has been extended to include the necessary formatting for errors that are reportable to wireserver.
  - All reportable errors, including supporting data, are now mapped to standardized KVP health reports via `Error::as_encoded_report`.
  - “Unhandled” and unexpected errors are now formatted and reported consistently for observability.

- **Tests and Test Helpers Updated:**
  - `unittest.rs` and `functional_tests.rs` have both been updated to reflect the new JSON endpoint. 

Initially based on #170, but that PR will be closed in favor of this one. 
Resolves #58 and #186.


libazureinit/src/config.rs
```diff
@@ -181,6 +181,10 @@ impl Default for AzureProxyAgent {
     }
 }
 
+/// Global default timeout for Wire server connections.
+pub const DEFAULT_WIRESERVER_TOTAL_RETRY_TIMEOUT_SECS: f64 = 1200.0;
+/// Global default retry interval for Wire server requests.
+pub const DEFAULT_WIRESERVER_CONNECTION_TIMEOUT_SECS: f64 = 2.0;
```
cjp256, line 187:
```suggestion
/// Retry wireserver up to 20 minutes.  The VM has most likely failed provisioning
/// due to timeout at this point.
pub const DEFAULT_WIRESERVER_TOTAL_RETRY_TIMEOUT_SECS: f64 = 1200.0;
pub const DEFAULT_WIRESERVER_CONNECTION_TIMEOUT_SECS: f64 = 60.0;
pub const DEFAULT_WIRESERVER_READ_TIMEOUT_SECS: f64 = 60.0;
```

src/main.rs
```diff
@@ -25,6 +26,11 @@ use std::time::Duration;
 use sysinfo::System;
 use tracing::instrument;
 
+use libazureinit::config::{
+    DEFAULT_WIRESERVER_CONNECTION_TIMEOUT_SECS,
```
cjp256, line 30: These should be disappeared and the config used.

libazureinit/src/config.rs
cjp256: I don't know that these need forward definitions, there should be no reference to this outside of config?

libazureinit/src/goalstate.rs
```diff
@@ -1,144 +1,126 @@
...
-///
-/// let res = libazureinit::goalstate::get_goalstate(
-///     &client, Duration::from_secs(1), Duration::from_secs(5),
-///     Some("http://127.0.0.1:8000/"),
-/// );
-/// ```
 #[instrument(err, skip_all)]
-pub async fn get_goalstate(
-    client: &Client,
+pub async fn report_provisioning_health(
```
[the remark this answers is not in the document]
peytonr18: This sounds good! The only thing I will need to make sure on passing the Config is that doing so doesn't "use/consume" the resource. Config isn't reusable, so passing it to each function may require us to do a lot of clones. I'll investigate that further and make sure that's the case - if not, I'll implement those examples.

src/main.rs
peytonr18: The reason I have it here is because we use the same timeout variable when we call report_provisioning_health if config fails (meaning, we won't be able to use any of the config values since there was an error with building the config). In that case, or in any cases where we call report_provisioning_health, we'd either need to hardcode a f64 value or use that global default that is defined in the config.

This is the situation I'm talking about: 
```rust
if let Err(report_error) = report_provisioning_health(
                "NotReady",
                Some("ProvisioningFailed"),
                Some("Invalid configuration schema"),
                Duration::from_secs_f64(
                    DEFAULT_WIRESERVER_CONNECTION_TIMEOUT_SECS,
                ),
                Duration::from_secs_f64(
                    DEFAULT_WIRESERVER_TOTAL_RETRY_TIMEOUT_SECS,
                ),
                None, // default wireserver URL 
```

libazureinit/src/goalstate.rs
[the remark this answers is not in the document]
peytonr18: So, the only advantage of having the individual knobs was in the case that `Config` fails, we could pass in the relevant global default values directly. 
In passing the full `Config`, we have to do something like this to report_failure when the config loading fails:
```rust
let config = match Config::load(opts.config.clone()) {
        Ok(config) => config,
        Err(error) => {
            eprintln!("Failed to load configuration: {error:?}");

            // Build temporary config to pass in wireserver defaults for report_failure
            let cfg = Config {
                wireserver: Wireserver {
                    connection_timeout_secs:
                        DEFAULT_WIRESERVER_CONNECTION_TIMEOUT_SECS,
                    read_timeout_secs: DEFAULT_WIRESERVER_READ_TIMEOUT_SECS,
                    total_retry_timeout_secs:
                        DEFAULT_WIRESERVER_TOTAL_RETRY_TIMEOUT_SECS,
                },
                ..Default::default()
            };

            if let Err(report_error) = report_failure(
                "Invalid configuration schema",
                &cfg,
                None, // default wireserver URL
            )
            .await
            {
                tracing::warn!(
                    "Failed to send provisioning failure report: {:?}",
                    report_error
                );
            }
```


Which, is totally fine with me! Just sharing for clarity.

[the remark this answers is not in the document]
peytonr18: Passing the whole Config in instead of relying on the specific knobs does require us to create a dummy default config and pass those values into report_failure when `Config` fails (as opposed to just passing the relevant global wireserver variables) but I'm fine with that! 

Just wanted to share for clarity. 

```rust
    let config = match Config::load(opts.config.clone()) {
        Ok(config) => config,
        Err(error) => {
            eprintln!("Failed to load configuration: {error:?}");

            // Build temporary config to pass in wireserver defaults for report_failure
            let cfg = Config::default();

            if let Err(report_error) = report_failure(
                "Invalid configuration schema",
                &cfg,
                None, // default wireserver URL
            )
            .await
            {
                tracing::warn!(
                    "Failed to send provisioning failure report: {:?}",
                    report_error
                );
            }
```

libazureinit/src/health.rs
```diff
@@ -0,0 +1,248 @@
...
+) -> Result<(), Error> {
+    tracing::info!("Reporting ready provisioning health");
+    _report("Ready", None, None, &config.wireserver, url_override).await
+}
+
+/// Report that provisioning has failed.
+pub async fn report_failure(
+    message: &str,
+    config: &Config,
+    url_override: Option<&str>,
```
cjp256, line 33: url_override may be better handled by putting wireserver_address into &config (i.e. default 168.63.129.16:80)

_technically_ the wireserver IP is presented in dhcp option 245 and could be different from 168.63.129.16

In practice, it won't happen anytime soon, or ever.

libazureinit/src/health.rs
```diff
@@ -0,0 +1,248 @@
...
+use serde_json::json;
+
+use crate::config::Config;
+use crate::error::Error;
+use crate::http;
+
+const DEFAULT_HEALTH_URL: &str = "http://168.63.129.16/provisioning/health";
+
+/// Report that provisioning is ready.
+pub async fn report_ready(
```
cjp256, line 21: I like the new APIs, thanks!

cjp256: For completeness, let's and report_in_progress(message,...) too.

libazureinit/src/health.rs
```diff
@@ -0,0 +1,248 @@
...
+                return Ok(());
+            }
+            StatusCode::TOO_MANY_REQUESTS => {
+                tracing::warn!("429 from wireserver, retrying…")
+            }
+            StatusCode::SERVICE_UNAVAILABLE => {
+                tracing::warn!("503 from wireserver, retrying…")
+            }
+            other => {
+                tracing::warn!("Non-retryable status, bailing out");
```
cjp256, line 121: can we add an info trace of every response from wireserver?  We can avoid this entire block with a simpler status code check

libazureinit/src/health.rs
peytonr18: When would that one be used? I can totally add it - just trying to think of use cases where we'd want to call report_in_progress or if there is a specific status code (429 or when there is a retry, etc) that we should base it on.

libazureinit/src/health.rs
peytonr18: I like this! I feel like it would be something we'd change later on anyways -- went ahead and made that change.

src/main.rs
peytonr18: Resolving this as it no longer applies and those have been removed!

libazureinit/src/health.rs
cjp256: I don't see immediate use for it, but it would be interesting to experiment with a bit.

My thought is to (perhaps optionally) use it is to report in progress at the start of provisioning to check in, or to include the result of a long-running operation prior to reporting ready (e.g. disk format, runcmd, etc.).  That way CRP knows (or could know) what the last operation to start was and blame it on OSPTO.  Today we leverage KVP diagnostics, but if we have networking up and wireserver is responding, it'd be a better choice.

src/kvp.rs
```diff
@@ -494,6 +576,71 @@ fn get_uptime() -> Duration {
...
+/// # Returns
+/// A tuple of (key, value) where key is fixed as "PROVISIONING_REPORT".
+fn build_success_health_report(
+    agent: &str,
+    optional_key: Option<&str>,
+    vm_id: &str,
+    timestamp: &str,
+) -> (String, String) {
+    let mut value = format!(
+        "result=success|agent={}|pps_type=None|vm_id={}|timestamp={}",
```
anhvoms, line 599: Can you file issue to correctly report the pps_type? - this can be retrieved from the extended IMDS call (which we're already using)

anhvoms (changes requested):

src/kvp.rs
peytonr18: Definitely, Issue #192

libazureinit/src/health.rs
```diff
@@ -0,0 +1,305 @@
...
+/// Report that provisioning has failed.
+pub async fn report_failure(
+    message: &str,
+    config: &Config,
+) -> Result<(), Error> {
+    tracing::info!(failure_reason=%message, "Reporting provisioning failed with the message {}", message);
+    _report(
+        "NotReady",
+        Some("ProvisioningFailed"),
+        Some(message),
```
cjp256, line 33: the failure report sent should match what we're doing with kvp. - reason=message

cjp256: the way we keep this consistent in cloud-init is we have a ReportableError class that formats it correctly and every reported failure implements a subclass of it.

https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/azure/errors.py

peytonr18: Fixed the failure formatting as requested. For the ReportableError idea, maybe we consider it as a future improvement if you think that'd ultimately be helpful?

libazureinit/src/health.rs
```diff
@@ -0,0 +1,310 @@
...
+    message: &str,
+    config: &Config,
+) -> Result<(), Error> {
+    let formatted_message = format!("reason={}", message);
+    tracing::info!(
+        failure_reason = %formatted_message,
+        "Reporting provisioning failed with the message: {}",
+        message
+    );
+    _report(
```
anhvoms, line 35: The formatted_message here should be the same message that would have been reported via KVP. It doesn't need to happen in this PR, but please file an issue for it.

anhvoms (changes requested):

libazureinit/src/health.rs
```diff
@@ -0,0 +1,310 @@
...
+                    .to_str()
+                    .ok()
+                    .and_then(|s| s.parse::<u64>().ok())
+                {
+                    tracing::warn!(
+                        "Retryable HTTP status {} received with Retry-After: {}s. Sleeping before retrying...",
+                        status,
+                        delay_secs
+                    );
+                    tokio::time::sleep(Duration::from_secs(delay_secs)).await;
```
jeremycline, line 148: This introduces an accounting issue with the `total_retry_timeout_secs`. The `remaining` variable is updated from the result of `http::post`, but in the event that the server gives us a `Retry-After` header, however long it tells us to wait is not included in the `remaining` time. One option would be to use `Duration::saturating_sub` to account for this, but this introduces another (less likely) issue.

In the worst case, the server is buggy and tells us to sleep for `MAX_INT` or something absurd, in which case this will appear to hang forever (although we will get a helpful log message with the duration, so it would be pretty obvious what happened). What I would recommend doing instead of the `while !remaining.is_zero()` is to use `loop {}` and wrap it in a `tokio::time::timeout` like the `http::request` function does. This way, no matter what happens, no matter what the server tells you to do, you will stop after `total_retry_timeout_secs`.

libazureinit/src/health.rs
```diff
@@ -0,0 +1,310 @@
...
+        &config.wireserver,
+    )
+    .await
+}
+
+/// Internal helper: all of the header setup, JSON‐body construction, retry loop, etc.
+#[instrument(err, skip_all)]
+async fn _report(
+    state: &str,
+    substatus: Option<&str>,
```
jeremycline, line 62: Very much optional, especially since this is an internal API, but you could define an `enum` for the known valid states and substates, and then impl a `to_str` method on them (or `ToString`, or `Into`, or...) so you can't mis-type the string states and it makes it clear to developers what is expected at a glance. 

I'm not sure it's worth it given the number of states and the fact that it's internal, so choose your own adventure.

libazureinit/src/health.rs
```diff
@@ -0,0 +1,310 @@
...
+
+        remaining = new_remaining;
+    }
+
+    tracing::warn!("Report '{}' timed out", state);
+    Err(Error::Timeout)
+}
+
+#[cfg(test)]
+mod tests {
```
jeremycline, line 180: I recommend adding a test case to cover `Retry-After` as well as a variant with a buggy/odd `Retry-After` that's greater than `total_retry_timeout_secs`

jeremycline (changes requested):

libazureinit/src/goalstate.rs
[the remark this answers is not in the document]
cjp256: thanks!

libazureinit/src/health.rs
cjp256: thanks!

libazureinit/src/health.rs
cjp256: Doesn't look like wireserver responds with Retry-After, even when throttled:

* Connected to 168.63.129.16 (168.63.129.16) port 80
> GET /machine/?comp=goalstate HTTP/1.1
> Host: 168.63.129.16
> User-Agent: curl/8.5.0
> Accept: */*
> x-ms-version:2012-11-30
>
< HTTP/1.1 403 Forbidden
< Content-Type: text/xml; charset=utf-8
< Server: Microsoft-IIS/10.0
< Date: Fri, 02 May 2025 20:28:09 GMT
< Content-Length: 357
<
<?xml version="1.0" encoding="utf-8"?>
<Error xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
    <Code>RequestRateLimitExceeded</Code>
    <Message>The request was denied because of too many requests by this client. Please retry the request after a few seconds.</Message>
    <Details></Details>
* Connection #0 to host 168.63.129.16 left intact
</Error>

libazureinit/src/health.rs
cjp256: I do like this idea.

libazureinit/src/health.rs
cjp256: I would remove Retry-After support and ensure we log response headers to give us data in the future if wireserver ever uses it and it seems legit.  Sleeping one second is very reasonable in this situation.

libazureinit/src/health.rs
cjp256: This will be the next PR.

libazureinit/src/health.rs
cjp256: Yes, this will be the next PR before we start adding more error conditions with new functionality.

libazureinit/src/http.rs
```diff
@@ -125,7 +123,7 @@ async fn request(
 
                         match response.error_for_status() {
                             Ok(response) => {
-                                if statuscode == StatusCode::OK {
+                                if statuscode == StatusCode::OK || statuscode == StatusCode::CREATED {
```
cjp256, line 126: i think we can remove this if statement altogether.  We're in this Ok() block so any status code we get should be safe to consider OK (2xx)

libazureinit/src/health.rs
peytonr18: Really liked this idea, implemented it in the most recent commit!

libazureinit/src/health.rs
peytonr18: Done in the most recent commit actually!

bpryan99 (approved):
LGTM

libazureinit/src/health.rs
```diff
@@ -0,0 +1,418 @@
...
+        HeaderValue::from_str(&format!("azure-init v{}", version)).unwrap(),
+    );
+    headers.insert(
+        "x-ms-guest-agent-name",
+        HeaderValue::from_str(&format!("azure-init/{}", version)).unwrap(),
+    );
+    headers
+        .insert("content-type", HeaderValue::from_static("application/json"));
+
+    tracing::info!(?headers, "Prepared HTTP headers");
```
jeremycline, line 173: IMO info is a little high for this. I'd recommend debug or even trace level since the header content won't be particularly interesting in most cases.

libazureinit/src/health.rs
```diff
@@ -0,0 +1,418 @@
...
+            tracing::info!(
+                target: "libazureinit::health::status",
+                "Report '{}' succeeded",
+                state
+            );
+            return Ok(());
+        }
+
+        if status == StatusCode::TOO_MANY_REQUESTS
+            || status == StatusCode::SERVICE_UNAVAILABLE
```
jeremycline, line 222: It might not hurt to retry on `INTERNAL_SERVER_ERROR` as well. I don't know if WireServer returns `INTERNAL_SERVER_ERROR` or any other 5XX codes, but lots of services _do_ even if they don't document that it's a possibility.

src/kvp.rs
```diff
@@ -205,6 +205,66 @@ impl EmitKVPLayer {
...
+                    reason.as_deref().unwrap_or("No failure reason provided"),
+                    Some("stuff"),
+                    &self.vm_id,
+                    &timestamp,
+                )),
+                other => Err(format!("Invalid health report kind: {}", other)),
+            };
+
+        let (key, value) =
+            kvp_result.expect("Failed to build health key-value pair");
```
jeremycline, line 264: If I understand correctly, this is a case of programmer error where they used a particular key in a trace without the expected value. Given that, I think panicking might be okay assuming it'll definitely happen in the test suite, although I would include some help in the panic so the developer knows what to do to fix their oversight, something like "health_report value must be 'success' or 'failure'".

jeremycline (commented):
All my concerns have been addressed. 

I looked through the KVP reporting bits, but didn't have much feedback as I'm less familiar with what it's "supposed" to output so all I can say is the code itself looks reasonable. There's one little note that's really just a nitpick around the error message.

src/kvp.rs
```diff
@@ -494,6 +576,71 @@ fn get_uptime() -> Duration {
...
+/// - `agent`: The agent version string.
+/// - `optional_key`: Optional additional key=value pairs.
+/// - `vm_id`: The VM ID.
+/// - `timestamp`: The precomputed timestamp in ISO 8601 format.
+///
+/// # Returns
+/// A tuple of (key, value) where key is fixed as "PROVISIONING_REPORT".
+fn build_success_health_report(
+    agent: &str,
+    optional_key: Option<&str>,
```
anhvoms, line 594: I'm having a little bit of trouble interpreting this `optional_key`. Are we allowing the user to pass in whatever key they want, as in, they can pass in something like `fruit=apple`, or are fixing the key part as `optional_key` and allow the customer to pass in some string xyz, then we send `optional_key=xyz` to the report?

If the latter, why not using "extra" to match the report failure?

src/kvp.rs
```diff
@@ -205,6 +205,66 @@ impl EmitKVPLayer {
         let encoded_kvp_flattened: Vec<u8> = encoded_kvp.concat();
         self.send_event(encoded_kvp_flattened);
     }
+
+    /// Processes and sends a health report event based on the provided `health_report` field.
+    ///
+    /// This function extracts the `reason` field (if present)
+    /// and constructs a properly formatted health report key-value pair (KVP) using
+    /// [`build_health_kvp`]. The resulting KVP is then encoded and sent via `send_event()`.
```
anhvoms, line 213: did you mean encode_kvp_item

src/kvp.rs
```diff
@@ -205,6 +205,66 @@ impl EmitKVPLayer {
...
+        let timestamp = DateTime::<Utc>::from(UNIX_EPOCH + event_time)
+            .format("%Y-%m-%dT%H:%M:%S%.3f")
+            .to_string();
+
+        let agent = format!("Azure-Init/{}", env!("CARGO_PKG_VERSION"));
+        let kvp_result: Result<(String, String), String> =
+            match health_report.as_str() {
+                "success" => Ok(build_success_health_report(
+                    &agent,
+                    Some("optional_value"),
```
anhvoms, line 249: why are we passing these literal strings of "optional_value" to KVP?

src/kvp.rs
peytonr18: Truthfully, I don't know what we would want the customer to pass in here - but, I kept it because it exists as an option in the cloud-init format. For whatever reason, the success report signature has optional_key and the failure report signature has extra.

peytonr18:
Refactored the health reporting logic to support structured extra and optional_key fields. In cloud-init, extra is effectively a dictionary of arbitrary key/value pairs (called supporting_data), so we updated our implementation to reflect that. It now accepts either a JSON string or a semicolon-delimited format like step=write_keys;component=ssh, and flattens the result inline in the failure report.

optional_key is supported only in success reports, which mirrors how cloud-init handles it — success and failure reports use slightly different formats for whatever reason. But the key point is that it is no longer a hardcoded variable. 

Also updated docs to clarify supported formats, added test coverage, treated 500s as retryable, and moved header logs to debug to reduce noise.

Let me know if there are any concerns with how I set this up!

libazureinit/src/config.rs
```diff
@@ -195,14 +200,20 @@ pub struct Wireserver {
...
     fn default() -> Self {
         Self {
-            connection_timeout_secs: 2.0,
-            read_timeout_secs: 60.0,
-            total_retry_timeout_secs: 1200.0,
+            connection_timeout_secs: DEFAULT_WIRESERVER_CONNECTION_TIMEOUT_SECS,
+            read_timeout_secs: DEFAULT_WIRESERVER_READ_TIMEOUT_SECS,
+            total_retry_timeout_secs:
+                DEFAULT_WIRESERVER_TOTAL_RETRY_TIMEOUT_SECS,
+            health_endpoint: "http://168.63.129.16/provisioning/health"
```
cjp256, line 215: should this be a const too for consistency with others?

cjp256:
Ideally I would prefer to see that the reportable errors are aligned with cloud-init's:
https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/azure/errors.py

If we can consolidate the reportable errors into one module for easy enumeration, that would be great.

src/kvp.rs
```diff
@@ -205,6 +206,159 @@ impl EmitKVPLayer {
...
+    ///     health_report = "failure",
+    ///     reason = "Network timeout",
+    ///     extra = r#"{"step":"imds_query","region":"westus"}"#,
+    ///     "Provisioning failed"
+    /// );
+    /// ```
+    ///
+    /// Both formats will be parsed into key-value pairs and included in the final KVP report.
+    fn handle_health_report(
+        &self,
```
cjp256, line 272: this should just be taking the encoded string that is output from ReportableError

libazureinit/src/config.rs
peytonr18: Initially didn't since it's not a variable referenced outside the config - but I think it makes the most sense for consistency so it's a const now!

libazureinit/src/health.rs
```diff
@@ -18,6 +18,58 @@ use crate::config::Config;
 use crate::error::Error;
 use crate::http;
 
+/// Provides  error categories for provisioning failures that match cloud-init.
+///
+/// Used to supply a standardized human-readable reason for provisioning errors.
+#[derive(Debug, Clone)]
+pub enum ProvisioningErrorKind {
```
cjp256, line 25: I think this would be a bit cleaner if we surfaced ReportableError constructors with the desired fields for supporting data, instead of enums, e.g.:

```
impl ReportableError {
    pub fn dhcp_interface_not_found(duration: f64) -> Self {
        let mut supporting_data = HashMap::new();
        supporting_data.insert("duration".into(), duration.to_string());
        Self::new("failure to find DHCP interface", Some(supporting_data))
    }

    pub fn dhcp_lease(duration: f64, interface: Option<String>) -> Self {
        let mut supporting_data= HashMap::new();
        supporting_data.insert("duration".into(), duration.to_string());
        supporting_data.insert("interface".into(), interface.unwrap_or_default());
        Self::new("failure to obtain DHCP lease", Some(supporting_data))
    }
}
```

Only implement what we use.  We don't need to sync some list with cloud-init's errors, but I want to ensure we align with them as we add them.

libazureinit/src/health.rs
```diff
@@ -78,76 +130,140 @@ impl ReportableError {
...
-        self.extra.insert(key.to_string(), value.to_string());
+    /// Adds a key-value pair to the supporting data for the report.
+    pub fn with_supporting_data(mut self, key: &str, value: &str) -> Self {
+        self.supporting_data
+            .insert(key.to_string(), value.to_string());
         self
     }
+
+    /// Serializes the reportable error as a pipe-delimited KVP entry.
+    pub fn as_encoded_report(&self) -> String {
```
cjp256, line 145: this should take the parameters that may be unknown from the caller creating the reportable error.

e.g. vm_id, pps_type, etc.

src/main.rs
```diff
@@ -227,6 +229,30 @@ async fn main() -> ExitCode {
             Err(error) => {
                 eprintln!("Failed to load configuration: {error:?}");
                 eprintln!("Example configuration:\n\n{}", Config::default());
+
+                // Build temporary config to pass in wireserver defaults for report_failure
+                let cfg = Config::default();
+
+                let report_result = report_failure(
+                    &ReportableError::unhandled_exception().to_string(),
```
cjp256, line 237: this should be replaced with an error specific to this, e.g. reason=failure to load configuration and see what we can include from this error

libazureinit/src/http.rs
```diff
@@ -125,10 +123,8 @@ async fn request(
 
                         match response.error_for_status() {
                             Ok(response) => {
-                                if statuscode == StatusCode::OK {
                                     tracing::info!(target: "libazureinit::http::success", "HTTP response succeeded with status {}", statuscode);
```
cjp256, line 126: should rustfmt be in CI?  I'm thinking this indent should have changed

peytonr18: I ran it pre-commit but maybe it didn't catch/save something - I'll make sure I run it again.

src/main.rs
```diff
@@ -260,11 +285,43 @@ async fn main() -> ExitCode {
...
+                "Provisioning completed successfully"
+            );
+
+            ExitCode::SUCCESS
+        }
         Err(e) => {
             tracing::error!("Provisioning failed with error: {:?}", e);
             eprintln!("{e:?}");
+
+            if let Err(report_err) = report_failure(
```
cjp256, line 312: discussed offline that `e` should be the reportable error that bubbles up

libazureinit/src/health.rs
```diff
@@ -0,0 +1,578 @@
...
+        data.push("pps_type=None".to_string());
+        data.push(format!("vm_id={vm_id}"));
+        data.push(format!("timestamp={}", timestamp.to_rfc3339()));
+        data.push(format!("documentation_url={}", self.documentation_url));
+        encode_report(&data)
+    }
+
+    /// Constructor for “failure to find DHCP interface”.
+    pub fn dhcp_interface_not_found(duration_secs: f64) -> Self {
+        let mut err = ReportableError::new("failure to find DHCP interface");
```
cjp256, line 114: we can lose these constructors and just have teh caller do them unless perhaps reused in multiple places

src/main.rs
peytonr18: Just implemented this. if `e` is already a `ReportableError`, we bubble it up as-is for health reporting; otherwise, we wrap it as an unhandled exception

peytonr18: Nevermind - synced offline and going to rework ReportableError to be integrated with our existing Error type in error.rs.

libazureinit/src/error.rs
```diff
@@ -72,10 +75,181 @@ pub enum Error {
     Timeout,
     #[error("Failed to update the sshd configuration")]
     UpdateSshdConfig,
+    #[error("Failed to load sshd config: {details}")]
+    ConfigLoadFailure { details: String },
+    #[error("Unhandled exception")]
```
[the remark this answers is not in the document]
cadejacobson: Building off of that: it looks like the error outputs the details in ConfigLoadFailure, do we want to do that for the unhandled error as well just to give information to the user?
`#[error("Unhandled error: {details}")]`

libazureinit/src/error.rs
```diff
@@ -72,10 +75,181 @@ pub enum Error {
...
 impl From<tokio::time::error::Elapsed> for Error {
     fn from(_: tokio::time::error::Elapsed) -> Self {
         Self::Timeout
     }
 }
+
+/// Implement reportable formatting for `Error` to be used in health reporting.
+impl Error {
+    /// Returns a human-readable summary describing the error variant.
+    pub fn reason(&self) -> String {
```
cjp256, line 93: i'm inclined to ask that reason=lower case unless ACRONYM to be consistent with existing reasons used in cloud-init

libazureinit/src/health.rs
```diff
@@ -0,0 +1,414 @@
...
+    Client,
+};
+use std::time::Duration;
+use tracing::instrument;
+
+use chrono::Utc;
+use reqwest::StatusCode;
+// use serde::Serialize;
+use serde_json::json;
+// use std::collections::HashMap;
```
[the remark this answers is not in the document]
peytonr18: Just saw that, great catch! Need a cargo command to catch anything I have commented out along the way!

src/main.rs
```diff
@@ -12,7 +12,8 @@ use libazureinit::imds::InstanceMetadata;
 use libazureinit::User;
 use libazureinit::{
     error::Error as LibError,
-    goalstate, imds, media,
+    health::{report_failure, report_ready},
+    imds, media,
     media::{get_mount_device, Environment},
```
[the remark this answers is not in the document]
peytonr18: Good question -- we don't need both. And the same is the case for `imds`, so consolidated `libazureinit` imports into one import statement.

cadejacobson (approved):
Thanks!

libazureinit/src/health.rs
```diff
@@ -0,0 +1,412 @@
...
+            }
+        };
+        write!(f, "{s}")
+    }
+}
+
+/// Constructs a KVP entry representing a successful provisioning event.
+pub fn encoded_success_report(
+    vm_id: &str,
+    optional_key_value: Option<(&str, &str)>,
```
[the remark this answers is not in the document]
cjp256: let's just remove it for now

jeremycline (commented):
Not quite done reviewing, but it's dinner time for me. I'll finish up tomorrow morning.

src/kvp.rs
```diff
@@ -178,6 +187,8 @@ impl EmitKVPLayer {
                 eprintln!("Failed to flush the log file: {e}");
             }
         }
+        // The receiver might have been dropped if the main thread doesn't wait
+        let _ = completion_tx.send(());
```
jeremycline, line 191: You're on the right track here, but I believe there's one more piece we need to ensure the KVP write can be gracefully shut down.

The problem is we only send this signal after we break from the `while let Some(encoded_kvp) = events.recv().await` loop.

From the `recv()` docs:
> This method returns None if the channel has been closed and there are no remaining messages in the channel’s buffer. This indicates that no further values can ever be received from this Receiver. The channel is closed when all senders have been dropped, or when [close](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.UnboundedReceiver.html#method.close) is called.

We need either all the senders to be dropped, or to call close on the `UnboundReceiver`. The problem is that the only sender is owned by `EmitKVPLayer` (and used via `send_event()`), and that object ultimately gets owned by the global tracing subscriber. This, in turn, means we have no way to drop the sender, so we'll never exit the while loop, and the user will wait on the oneshot signal forever. The alternative, calling close ourselves, is difficult because we don't know when to do that.

So... We need to be able communicate to the task somehow that we want it to finish up, and we can't do that by dropping all the senders since the tracing layer _needs_ to have a handle to that and it also needs to live for the whole life of the program. We _could_ provide another oneshot that the user triggers to start the shutdown, but there's a convenient structure in `tokio_util` for this sort of thing: [CancellationToken](https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html). 

The way this would work is that the caller provides a token when creating the `EmitKVPLayer`. Then, when they know it's time to shut down, they call cancel on that token to let the tokio task know it's time to close the `UnboundReceiver` and drain any existing messages. You also give the caller the `JoinHandle` from `tokio::spawn` which they can `await` after they call cancel. The function signature would look something like this:

```rust
    pub fn new(
        file_path: std::path::PathBuf,
        vm_id: &str,
        graceful_shutdown: tokio_util::sync::CancellationToken,
    ) -> Result<(Self, tokio::task::JoinHandle<_>), anyhow::Error> {
```

Note that I do think it's a tad un-idomatic to return anything besides `Self` or `Result<Self, Error>` from a `new()` function, but it's not the end of the world. An alternative would be to add a new type that has the layer as one field and the task as another, so that the user can split them out into two separate owned objects. So the public API might look something like:
```Rust
pub struct Kvp {
    pub tracing_layer: EmitKVPLayer,
    pub writer: JoinHandle<_>,
}

impl Kvp {
    pub fn new(
        file_path: std::path::PathBuf,
        vm_id: &str,
        graceful_shutdown: tokio_util::sync::CancellationToken,
    ) -> Result<Self, anyhow::Error> {
}
```

src/kvp.rs
```diff
@@ -31,17 +32,127 @@ use tracing_subscriber::{
...
+                            eprintln!("Failed to write to log file during shutdown: {e}");
+                        }
+                        if let Err(e) = file.flush() {
+                            eprintln!("Failed to flush the log file during shutdown: {e}");
+                        }
+                    }
+                    // All messages are drained, exit the loop.
+                    break;
+                }
+            }
```
jeremycline, line 150: Very neatly done, this looks right to me!

jeremycline (approved):
Once KVP gets moved to the library we may want to tweak some of the things we're marking `pub` but I don't think we should deal with that in this PR.

I'm not an expert on what should be going out via KVP so I'll leave that to Anh and Chris, but I don't spot any issues with the approach. Great work!

libazureinit/src/health.rs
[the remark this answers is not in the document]
cjp256: It's fine to keep or lose it.  It's more of an optional feature in the spec.  I'd actually think pps_type should only be set if not None to conserve space.

anhvoms (approved):

PR #190, opened 2025-04-15
Refactor logging to avoid double-init panic and support tracing layer reloads

peytonr18:
## Reason

`main.rs` previously called `tracing::subscriber::set_global_default(...)` more than once during our two-phase logging setup (pre- and post-config). This caused a race condition and occasional panics in `tracing-subscriber` on Ubuntu 20.04.

This PR refactors that logic so the global subscriber is installed exactly once. The `logging.rs` module now dynamically reloads only the KVP and file layers once the config is available.

## Changes

This PR updates the two-phase logging setup to use a single global subscriber and reloadable layers (KVP and file) for config-driven updates.

Previously, calling `set_global_default` twice resulted in a panic and this error message:

> `assertion failed: self.replace(val).is_none()`

With these changes:

- The global subscriber is installed once at startup using fallback defaults.
- After the config is loaded, only the KVP and file layers are reloaded using values from `config.rs`.
- Behavior remains the same (KVP output, file logging), but we eliminate the risk of panics and race conditions.


src/logging.rs
```diff
@@ -1,21 +1,69 @@
...
+    layer::SubscriberExt,
+    reload::{self, Handle},
+    EnvFilter, Layer, Registry,
 };
 
 use crate::kvp::EmitKVPLayer;
 use libazureinit::config::{Config, DEFAULT_AZURE_INIT_LOG_PATH};
 
+/// A global static that holds reload handles & writer state
+static LOGGING_STATE: OnceCell<LoggingState> = OnceCell::new();
```
jeremycline, line 25: I think we can avoid a lot of the tricky bits of this PR by changing the main function to do something like this:

```Rust
#[tokio::main]
async fn main() -> ExitCode {
    let opts = Cli::parse();

    let tracer = initialize_tracing();
    let vm_id: String = get_vm_id()
        .unwrap_or_else(|| "00000000-0000-0000-0000-000000000000".to_string());

    // where subscriber is one with acceptable defaults to cover anything emitted while we load the config that includes
    // the desired logging config options
    let config = tracing::subscriber::with_default(subscriber, || Config::load(opts.config.clone()));
    let config = match config {
        Ok(config) => config,
        Err(error) => {
            eprintln!("Failed to load configuration: {error:?}");
            return ExitCode::FAILURE;
        }
    };

    if let Err(e) = setup_layers(tracer, &vm_id, Some(&config)) {
        tracing::error!("Failed to set final logging subscriber: {e:?}");
    }
    ...
```

This drops the initial call to `setup_layers` in favor of applying a temporary tracing subscriber while running the configuration loading function. That temporary subscriber can be configured however you like, ideally as minimally as possible to ensure it doesn't have a lot of complicated setup, but enough to ensure a malformed config is reported in a clear way.

After you've got your config loaded, you can set the global default. This way you have just one place where the global default is set and you can avoid accounting for current state.

peytonr18: This is a much easier fix. Took this approach and just added a tracing call to emit the finalized `Config` into the kvp layer after the global default has been set so we have it there if needed.

src/logging.rs
```diff
@@ -87,9 +84,7 @@ pub fn setup_layers(
         .with_writer(std::io::stderr)
         .with_filter(EnvFilter::from_env("AZURE_INIT_LOG"));
 
-    let log_path = config
-        .map(|cfg| cfg.azure_init_log_path.path.clone())
-        .unwrap_or_else(|| PathBuf::from(DEFAULT_AZURE_INIT_LOG_PATH));
+    let log_path = config.azure_init_log_path.path.clone();
```
jeremycline, line 87: Total nitpick, but since you're no longer using `DEFAULT_AZURE_INIT_LOG_PATH` as a fall back, you don't really need this clone since all the users of it take references; you can just replace `log_path` with `config.azure_init_log_path.path` references.

src/logging.rs
```diff
@@ -59,9 +58,7 @@ pub fn setup_layers(
         .join(","),
     )?;
 
-    let kvp_enabled = config
-        .map(|cfg| cfg.telemetry.kvp_diagnostics)
-        .unwrap_or(true);
+    let kvp_enabled = config.telemetry.kvp_diagnostics;
```
jeremycline, line 61: It's fine as it is, but since you don't have a fallback anymore you can just use `if config.telemetry.kvp_diagnostics {` directly

jeremycline (approved):
Other than a few no-longer-necessary variables which don't hurt anything, this looks good to me!

cjp256 (approved):
LGTM but would also recommend @jeremycline's advise

jeremycline (approved):

PR #193, opened 2025-04-24
Add `clean` CLI subcommand to remove provisioning marker and allow re-provisioning

### LINKED ISSUE #167: Create a CLI feature to clear the provisioning directory
Create an additional cli feature to clear the provisioning directory. This might come in handy if we need to force azure-init to provision again. For any software that has a "cache", it should offer some feature to clear the cache

_Originally posted by @anhvoms in https://github.com/Azure/azure-init/issues/164#issuecomment-2666532760_

peytonr18:
This PR introduces a new azure-init `clean` subcommand to allow users to manually remove the provisioning marker file associated with a VM. This is useful in scenarios where a user wants to re-provision a VM.

## Changes
- Added a clean subcommand to the CLI definition.
- Implemented `clear_provisioning_status()`, which removes the .provisioned file associated with the VM ID.
- Integrated the logic into main.rs, returning `ExitCode::SUCCESS` if cleared or if the .provisioned file isn't found, and `ExitCode::FAILURE` if deletion fails unexpectedly.

## Requesting Feedback
- This functionality has been added into `main.rs` but would it be better to put the `clear_provisioning_status()` into another file or possibly its own file? Was not sure what the best practice here would be. 

Resolves #167

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -64,12 +64,9 @@ struct Cli {
 
 #[derive(Subcommand, Debug)]
 enum Command {
-    /// Clean up files created by azure-init
+    /// By default, this removes provisioning state data. Optional flags can be used
+    /// to clean logs or additional generated files.
     Clean {
-        /// Cleans the provisioning state for the current VM
-        #[arg(long, default_value_t = true)]
-        provision: bool,
-
         /// Cleans the log files as defined in the configuration file
         #[arg(long)]
         logs: bool,
@@ -210,7 +207,6 @@ async fn main() -> ExitCode {
     let tracer = initialize_tracing();
     let vm_id: String = get_vm_id()
         .unwrap_or_else(|| "00000000-0000-0000-0000-000000000000".to_string());
-
     let opts = Cli::parse();
 
     let temp_layer = tracing_subscriber::fmt::layer()
@@ -245,8 +241,8 @@ async fn main() -> ExitCode {
         config
     );
 
-    if let Some(Command::Clean { provision, logs }) = opts.command {
-        if provision && clean_provisioning_status(&config).is_err() {
+    if let Some(Command::Clean { logs }) = opts.command {
+        if clean_provisioning_status(&config).is_err() {
             return ExitCode::FAILURE;
         }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -118,35 +118,63 @@ fn get_username(
         })
 }
 
-/// Cleans the provisioning state marker for the current VM.
+/// Cleans all provisioning state marker files from the azure-init data directory.
 ///
-/// This removes the `.provisioned` file named after the VM ID from the
-/// configured azure-init data directory (typically `/var/lib/azure-init`).
-/// Removing this file allows provisioning to run again on the next startup.
+/// This removes all files ending in `.provisioned` from the directory specified
+/// by `azure_init_data_dir` (typically `/var/lib/azure-init`). These marker files
+/// indicate that provisioning has completed. Removing them allows azure-init to
+/// re-run provisioning logic on the next boot.
 #[instrument]
-fn clean_provisioning_status(
-    config: &Config,
-    vm_id: &str,
-) -> Result<(), std::io::Error> {
-    let provisioned_marker = &config
-        .azure_init_data_dir
-        .path
-        .join(format!("{vm_id}.provisioned"));
+fn clean_provisioning_status(config: &Config) -> Result<(), std::io::Error> {
+    let data_dir = &config.azure_init_data_dir.path;
+    let mut found = false;
 
-    match std::fs::remove_file(provisioned_marker) {
-        Ok(_) => {
-            tracing::info!(
-                "Successfully clear provisioning state at: {:?}",
-                provisioned_marker
-            );
+    for entry in std::fs::read_dir(data_dir)? {
+        let path = match entry {
+            Ok(e) => e.path(),
+            Err(e) => {
+                tracing::error!(
+                    "Failed to read directory entry in {:?}: {:?}",
+                    data_dir,
+                    e
+                );
+                return Err(e);
+            }
+        };
+
+        if path.extension().is_some_and(|ext| ext == "provisioned") {
+            found = true;
+
+            match std::fs::remove_file(&path) {
+                Ok(_) => {
+                    tracing::info!(
+                        "Successfully removed provisioning state at: {:?}",
+                        path
+                    );
+                }
+                Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
+                    tracing::info!(
+                        "No provisioning marker found at: {:?}",
+                        path
+                    );
+                }
+                Err(e) => {
+                    tracing::error!(
+                        "Failed to clean provisioning marker {:?}: {:?}",
+                        path,
+                        e
+                    );
+                    return Err(e);
+                }
+            }
         }
-        Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
-            tracing::info!(
-                "No provisioning state to clear at: {:?}",
-                provisioned_marker
-            );
-        }
-        Err(e) => return Err(e),
+    }
+
+    if !found {
+        tracing::info!(
+            "No provisioning marker files (*.provisioned) found in {:?}",
+            data_dir
+        );
     }
 
     Ok(())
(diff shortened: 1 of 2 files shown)


anhvoms:
Consider allowing the user to specify whether log should be cleaned. There are many instances where we just want to clear the data directory to allow re-provisioning, but the logs should be kept for diagnostic purposes. cloud-init allows `clean --logs` to indicate that one would want to clean logs as well.

src/main.rs
```diff
@@ -56,6 +56,20 @@ struct Cli {
...
+
+#[derive(Subcommand, Debug)]
+enum Command {
+    Clean {
+        #[arg(long, default_value_t = true)]
+        provision: bool,
+
+        #[arg(long)]
+        logs: bool,
+    },
```
jeremycline, line 72: If you add a docblock to this enum variant, it will render has help text in the CLI. For example:
```suggestion
    /// Clean up files created by azure-init
    Clean {
        /// Cleans the provisioning state for the current VM
        #[arg(long, default_value_t = true)]
        provision: bool,

        /// Cleans the log files as defined in the configuration file
        #[arg(long)]
        logs: bool,
    },
```

Will cause the CLI help text to be:

```
❯ cargo run --bin=azure-init -- --help
Minimal provisioning agent for Azure

Create a user, add SSH public keys, and set the hostname.

Arguments provided via command-line arguments override any arguments provided via environment variables.

Usage: azure-init [OPTIONS] [COMMAND]

Commands:
  clean  Clean up files created by azure-init
  help   Print this message or the help of the given subcommand(s)

<snip>
```

and the command help to be:

```
❯ cargo run --bin=azure-init -- clean --help
Clean up files created by azure-init

Usage: azure-init clean [OPTIONS]

Options:
      --provision  Cleans the provisioning state for the current VM
      --logs       Cleans the log files as defined in the configuration file
  -h, --help       Print help
```

It might also be a good idea to explicitly note in the `struct Cli` docblock that if run without sub-commands, it'll provision the host.

src/main.rs
```diff
@@ -100,6 +114,62 @@ fn get_username(
         })
 }
 
+/// Cleans the provisioning state marker for the current VM.
+///
+/// This removes the `.provisioned` file named after the VM ID from the
+/// configured azure-init data directory (typically `/var/lib/azure-init`).
+/// Removing this file allows provisioning to run again on the next startup.
+#[instrument]
+fn clear_provisioning_status(
```
jeremycline, line 123: I'd recommend updating the function names to use `clean` instead of `clear` for consistency

src/main.rs
```diff
@@ -140,6 +210,26 @@ async fn main() -> ExitCode {
         config
     );
 
+    if let Some(Command::Clean { provision, logs }) = &opts.command {
```
jeremycline, line 213: nitpick, but I _believe_ you can drop the reference here and use `opts.command` so you later don't need to dereference the bools in your if statements.

jeremycline (commented):
A few little polish notes, but otherwise looks good to me

peytonr18:
> Consider allowing the user to specify whether log should be cleaned. There are many instances where we just want to clear the data directory to allow re-provisioning, but the logs should be kept for diagnostic purposes. cloud-init allows `clean --logs` to indicate that one would want to clean logs as well.

I should have added support for this! Using--logs, users can choose to explicitly remove the azure-init.log file during a clean operation. Let me know if there are additional log files you’d like included in this behavior!

src/main.rs
```diff
@@ -100,6 +118,62 @@ fn get_username(
...
+/// Removing this file allows provisioning to run again on the next startup.
+#[instrument]
+fn clean_provisioning_status(
+    config: &Config,
+    vm_id: &str,
+) -> Result<(), std::io::Error> {
+    let provisioned_marker = &config
+        .azure_init_data_dir
+        .path
+        .join(format!("{vm_id}.provisioned"));
```
cjp256, line 134: This should be a method so there's only one place this logic is maintained.

cjp256: OTOH it perhaps should delete *.provisioned

src/main.rs
```diff
@@ -140,6 +214,26 @@ async fn main() -> ExitCode {
         config
     );
 
+    if let Some(Command::Clean { provision, logs }) = opts.command {
+        if provision {
+            match clean_provisioning_status(&config, &vm_id) {
+                Ok(_) => tracing::info!("Provisioning state cleared."),
+                Err(e) => {
+                    tracing::error!("Failed to clear provisioning state: {e:?}")
```
cjp256, line 222:
```suggestion
                    tracing::error!("Failed to clean provisioning state: {e:?}")
```

src/main.rs
```diff
@@ -140,6 +214,26 @@ async fn main() -> ExitCode {
         config
     );
 
+    if let Some(Command::Clean { provision, logs }) = opts.command {
+        if provision {
+            match clean_provisioning_status(&config, &vm_id) {
+                Ok(_) => tracing::info!("Provisioning state cleared."),
```
cjp256, line 220:
```suggestion
                Ok(_) => tracing::info!("Provisioning state cleaned."),
```

src/main.rs
```diff
@@ -140,6 +214,26 @@ async fn main() -> ExitCode {
...
+                Err(e) => {
+                    tracing::error!("Failed to clear provisioning state: {e:?}")
+                }
+            }
+        }
+
+        if logs {
+            match clean_log_file(&config) {
+                Ok(_) => tracing::info!("Log file cleared."),
+                Err(e) => tracing::error!("Failed to clear log file: {e:?}"),
```
cjp256, line 230:
```suggestion
                Err(e) => tracing::error!("Failed to delete log file: {e:?}"),
```

src/main.rs
```diff
@@ -100,6 +118,62 @@ fn get_username(
...
+/// This removes the log file at the path configured by `azure_init_log_path`,
+/// which defaults to `/var/log/azure-init.log`. If the file does not exist,
+/// a message is logged but no error is returned.
+#[instrument]
+fn clean_log_file(config: &Config) -> Result<(), std::io::Error> {
+    let log_path = &config.azure_init_log_path.path;
+
+    match std::fs::remove_file(log_path) {
+        Ok(_) => {
+            tracing::info!("Successfully cleared log file at: {:?}", log_path);
```
cjp256, line 166:
```suggestion
            tracing::info!("Successfully removed log file at: {:?}", log_path);
```

src/main.rs
```diff
@@ -100,6 +118,62 @@ fn get_username(
...
+        }
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
+            tracing::warn!("No log file found to clear at: {:?}", log_path);
```
cjp256, line 169: This should probably be at debug level as the case is harmless.

cjp256, line 169:
```suggestion
            tracing::debug!("No log file found at: {:?}", log_path);
```

src/main.rs
```diff
@@ -100,6 +118,62 @@ fn get_username(
...
+    vm_id: &str,
+) -> Result<(), std::io::Error> {
+    let provisioned_marker = &config
+        .azure_init_data_dir
+        .path
+        .join(format!("{vm_id}.provisioned"));
+
+    match std::fs::remove_file(provisioned_marker) {
+        Ok(_) => {
+            tracing::info!(
```
cjp256, line 138:
```suggestion
            tracing::debug!(
```

src/main.rs
```diff
@@ -140,6 +214,26 @@ async fn main() -> ExitCode {
...
+            match clean_provisioning_status(&config, &vm_id) {
+                Ok(_) => tracing::info!("Provisioning state cleared."),
+                Err(e) => {
+                    tracing::error!("Failed to clear provisioning state: {e:?}")
+                }
+            }
+        }
+
+        if logs {
+            match clean_log_file(&config) {
```
cjp256, line 228: don't log anything here as it should already be done in clean_log_file().

Exit with failure code instead of success on error.

src/main.rs
cjp256, line 222: make sure we exit with error in this case

src/main.rs
```diff
@@ -100,6 +118,62 @@ fn get_username(
...
+    let log_path = &config.azure_init_log_path.path;
+
+    match std::fs::remove_file(log_path) {
+        Ok(_) => {
+            tracing::info!("Successfully cleared log file at: {:?}", log_path);
+        }
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => {
+            tracing::warn!("No log file found to clear at: {:?}", log_path);
+        }
+        Err(e) => return Err(e),
```
cjp256, line 171: same, log here and remove traces from caller to avoid redundant logs

jeremycline (commented):
It looks like you addressed everything I noted, although I realized there's not test coverage for this and we should probably have that.

For this kind of thing, I usually write an [integration test](https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html) using [assert_cmd](https://crates.io/crates/assert_cmd). Since everything the command touches is configurable, your test can create a temp directory with fake logs/provision files, write out a config file pointing to that directory, and then feed it all into the command line and assert return values, standard output, etc.

src/main.rs
```diff
@@ -56,6 +57,23 @@ struct Cli {
...
+    command: Option<Command>,
+}
+
+#[derive(Subcommand, Debug)]
+enum Command {
+    /// Clean up files created by azure-init
+    Clean {
+        /// Cleans the provisioning state for the current VM
+        #[arg(long, default_value_t = true)]
+        provision: bool,
```
anhvoms, line 71: Maybe call this `data` or `cache` instead of `provision`

cjp256 (approved):
Everything looks good, the only thing I recommend is splitting the test `clean_removes_provision_and_log_files` into two separate tests.  One without log cleanup and one with.

jeremycline (approved):

PR #194, opened 2025-04-28
build(Makefile): add install target

cjp256:
Add make install target to install azure-init package to DESTDIR.

Allow for toggling debug vs. release builds with optional BUILD_MODE and a default for debug.

- Install azure-init to /usr/bin/azure-init.
- Install service to /usr/lib/systemd/system/azure-init.service.
- Drop ConditionFileIsExecutable.
- Update demo script.

Example:

`make install DESTDIR=/tmp/foo`


peytonr18 (approved):
LGTM!

PR #200, opened 2025-05-20
.github: ignore automatic updates of idna_adapter

dongsupark:
Ignore automatic updates of `idna_adapter`, to make it compatible with recent changes in https://github.com/Azure/azure-init/pull/198.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,8 @@ updates:
     schedule:
       interval: "weekly"
     ignore:
+      - dependency-name: "idna_adapter"
+        # stay with <= 1.2.0 for Rust 1.78
       - dependency-name: "litemap"
         # stay with <= 0.7.4 for Rust 1.76
       - dependency-name: "zerofrom"


.github/dependabot.yml
```diff
@@ -9,6 +9,8 @@ updates:
     schedule:
       interval: "weekly"
     ignore:
+      - dependency-name: "idna_adapter"
+        # stay with <= 1.2.0 for Rust 1.78
       - dependency-name: "litemap"
         # stay with <= 0.7.4 for Rust 1.76
```
anhvoms, line 15: MSRV is 1.78, is the `litemap` ignore still correct?

anhvoms (approved):

.github/dependabot.yml
dongsupark: According to [comments](https://github.com/Azure/azure-init/blob/628d80f2c012f8b762c51f5a7cf090a879fe8677/libazureinit/Cargo.toml#L29), they should be kept until Sep. 2025.

Issue #206, opened 2025-06-09
Split CLI Tests in Two


peytonr18:
Everything looks good, the only thing I recommend is splitting the test `clean_removes_provision_and_log_files` into two separate tests.  One without log cleanup and one with.

_Originally posted by @cjp256 in https://github.com/Azure/azure-init/pull/193#pullrequestreview-2898233357_

dongsupark:
Done, https://github.com/Azure/azure-init/pull/208

PR #208, opened 2025-07-01
Split provision and log file test in two

cadejacobson:
This PR fixes issue #206 and splits `clean_removes_provision_and_log_files` into two tests. The first test, `provision_and_log_files_are_created` first ensures that the provisioning and log files can be created, and then `clean_removes_provision_and_log_files` actually deletes these files after they have been created. 

## How to use
This can be used by running unit tests in the workspace with the below commands.

## Testing done
As this is just a unit testing change, I have not done any image tests. I have run: 
```
cargo test --all-features --workspace
cargo fmt --all --check
```

### DIFF
--- a/tests/cli.rs
+++ b/tests/cli.rs
@@ -20,9 +20,16 @@ fn help_groups() -> Result<(), Box<dyn std::error::Error>> {
     Ok(())
 }
 
-#[test]
-fn clean_removes_only_provision_files_without_log_arg(
-) -> Result<(), Box<dyn std::error::Error>> {
+// Helper function to set up the log and provision files for cleaning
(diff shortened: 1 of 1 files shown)


cadejacobson:
> Thanks! To clarify, what I was hoping for was two tests for clean:
> 
> (A) cmd.args(["--config", config_path.to_str().unwrap(), "clean", "--logs"]); (B) cmd.args(["--config", config_path.to_str().unwrap(), "clean"]);
> 
> this verifies that clean without --logs works as expected.

This should be fixed now! Thanks for the review.

dongsupark (approved):
Looks good.
As this PR touches only unit tests, I think there is no need to run further tests.

Issue #210, opened 2025-07-02
Create unit test coverage for passwd()


cadejacobson:
Currently, there is no test coverage for ensuring that the `passwd()` function is doing what it should. The `passwd()` function interacts with system-level components, so this should be mocked.

The `passwd()` function currently contains some sshd configuration checks, so we should wait until the PR #211 is completed to implement these tests.

cadejacobson:
This was completed with PR #215.

PR #211, opened 2025-07-02
Utilize IMDS disablePasswordAuthentication to Configure Password Authentication

peytonr18:
This PR wires through the `disablePasswordAuthentication` flag from Azure instance metadata (IMDS) to ensure the `sshd_config` file correctly reflects the intended authentication state.

**Key changes:** 
- Extracted the flag from IMDS in main.rs and passed it to `Provision::new(...)`.
- Modified `update_sshd_config()` to take a bool and write the appropriate value.
- Updated `PasswordProvisioner::set()` to pass the flag through to `update_sshd_config()`.
- Updated all existing unit tests. 
- Added a new test case to verify "PasswordAuthentication no" is written when disabled.

**Logic Note:**
I chose to add the `disable_password_authentication` flag to the Provision struct so that azure-init could pass the exact value directly from the IMDS metadata at runtime. If there are objections to that, I can reconfigure this!


anhvoms (approved):
LGTM!

libazureinit/src/provision/password.rs
```diff
@@ -26,7 +24,7 @@ impl PasswordProvisioner {
 // Determines the appropriate SSH configuration file path based on the filesystem.
 // If the "/etc/ssh/sshd_config.d" directory exists, it returns the path for a drop-in configuration file.
 // Otherwise, it defaults to the main SSH configuration file at "/etc/ssh/sshd_config".
-fn get_sshd_config_path() -> &'static str {
+pub(crate) fn get_sshd_config_path() -> &'static str {
```
cadejacobson, line 27: Would it make sense to move this function to the `mod.rs` file as well or even the `ssh.rs` file? Now that the `update_sshd_config()` logic has been moved out of `password.rs`, we are not using this function in `password.rs`.

libazureinit/src/provision/mod.rs
```diff
@@ -85,6 +88,29 @@ impl Provision {
...
+        // update sshd_config based on IMDS disablePasswordAuthentication value.
+        let ssh_config_update_required = self
+            .config
+            .password_provisioners
+            .backends
+            .first()
+            .is_some_and(|b| matches!(b, PasswordProvisioner::Passwd));
+
+        if ssh_config_update_required {
+            let sshd_config_path = password::get_sshd_config_path();
```
cadejacobson, line 100: We would then look for the function in the scope `ssh::` which would be a little more intuitive I think.

cadejacobson (commented):
This will be super useful! It is going to make mocking the `passwd()` function **_so much_** easier for #210. Thank you!

cadejacobson (approved):
Looks good to me! Thanks

anhvoms (approved):

PR #217, opened 2025-07-29
New Signal Endpoint Example

peytonr18:
To avoid the condition where main exits via `EXITCODE::SUCCESS/FAILURE` before the `kvp_writer` finishes writing the health report to the `.kvp_pool_1` file, this draft PR example introduces a `oneshot` signal. 

Channel Creation: 
-  The `tokio::sync::oneshot` channel is a lightweight channel perfect for sending a single completion signal.
- The `EmitKVPLayer::new` function now creates this channel. It gives the sending end to the kvp_writer background task and returns the receiving end to the caller.
- When the kvp_writer task finishes its work (i.e., its input channel closes and all messages are flushed), it sends a signal through the oneshot channel just before it exits.

Signal Receiver (src/logging.rs):
- The `setup_layers` function, which initializes the logging system, was updated to receive the oneshot receiver from the EmitKVPLayer.
- It then passes this receiver up to the main function, making the completion signal available to the application's root.
- I also created a type alias LoggingSetup to simplify the complex return signature and improve code readability.

Wait / Receive Signal: 
- The main function now receives the oneshot receiver after setting up logging.
- After all provisioning logic is complete (whether it succeeds or fails), main calls .await on the receiver.
- This .await blocks the main function from exiting until the signal is received from the kvp_writer task, guaranteeing that the final log has been flushed to disk.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -220,11 +220,17 @@ async fn main() -> ExitCode {
     let temp_subscriber =
         tracing_subscriber::Registry::default().with(temp_layer);
 
-    let config =
+    let (config, kvp_completion_rx) =
         match tracing::subscriber::with_default(temp_subscriber, || {
-            Config::load(opts.config.clone())
+            let config = Config::load(opts.config.clone())?;
+            let (subscriber, rx) = setup_layers(tracer, &vm_id, &config)?;
+            if let Err(e) = tracing::subscriber::set_global_default(subscriber)
+            {
+                eprintln!("Failed to set global default subscriber: {e}");
+            }
+            Ok::<_, anyhow::Error>((config, rx))
         }) {
-            Ok(cfg) => cfg,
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -279,90 +282,97 @@ async fn main() -> ExitCode {
...
+                    Some(LibError::NonEmptyPassword) => ExitCode::from(config),
+                    Some(_) | None => ExitCode::FAILURE,
                 }
-                Some(LibError::NonEmptyPassword) => ExitCode::from(config),
-                Some(_) | None => ExitCode::FAILURE,
             }
         }
+    };
+
+    if let Some(rx) = kvp_completion_rx {
```
cadejacobson, line 370: I like this solution a lot! This definitely beats a static `sleep()` call in my book.

PR #218, opened 2025-08-01
Fix `Mutex Poisoning` Error Caused by Double Insert Panics

peytonr18:
This PR resolves an intermittent `Mutex poisoned` panic that occurred during VM provisioning. The root cause was a crash in the `tracing` library, triggered when the telemetry layer tried to mark the same operation as "failed" twice. 

This fix prevents this by: 
1. Introducing a `set_span_as_failed()` helper method that ensures that the tracing layer only inserts the failure status once per span. 
2. Added unit tests that reproduce the original panic (confirming the root cause and providing regression testing) and verifying the fix solves the original panic.

### DIFF
--- a/src/kvp.rs
+++ b/src/kvp.rs
@@ -26,7 +26,9 @@ use tracing::{
 };
 
 use tracing_subscriber::{
-    layer::Context as TracingContext, registry::LookupSpan, Layer,
+    layer::Context as TracingContext,
+    registry::{LookupSpan, SpanRef},
+    Layer,
 };
 
 use sysinfo::System;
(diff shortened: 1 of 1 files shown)


src/kvp.rs
```diff
@@ -734,4 +746,59 @@ mod tests {
...
+                event!(Level::ERROR, "This is the first error.");
+                event!(Level::ERROR, "This is the second error.");
+            }
+            a_failing_operation();
+        });
+        // The test passes if it completes without panicking.
+    }
+    #[test]
+    #[should_panic(expected = "assertion failed: self.replace(val).is_none()")]
+    fn test_repro_insert_panic() {
```
cjp256, line 776: This test is awesome for confirming the behavior of the library and highlighting the root cause.  Since it doesn't really add any coverage to azure-init though, I would suggest removing it and using it as documentation in the PR/commit message.

peytonr18: Done, that makes more sense!!

cadejacobson (approved):
Looks awesome

PR #223, opened 2025-08-18
Refactor library: move logging.rs and kvp.rs into libazureinit crate

peytonr18:
This commit refactors the codebase by relocating `kvp.rs` and `logging.rs` from the main crate into the libazureinit crate. This change is an architectural step toward consolidating KVP event formatting and emission within the core provisioning library and for streamlining the report_ready logic. 

This is also required for Afterburn integration. 

# Changes:
- The `kvp.rs` and `logging.rs` modules were moved from the main azure-init binary crate into the libazureinit library.
- The logic for generating health report strings was centralized in `health.rs`. The KVP logging mechanism was also simplified, removing redundant logic that was previously spread between `kvp.rs` and `health.rs`. Now, `health.rs` is now the single source of truth for what a health report contains, and it emits the properly formatted health report directly to `kvp.rs` as a value. 
- A few other minor changes were made to ensure `azure-init.log` contains what is expected. During the refactor, logs previously sent to `azure-init.log` were suppressed. Minor fixes to the filtering setup were made to address this and unit tests were added to prevent future regression.

### DIFF
--- a/libazureinit/src/logging.rs
+++ b/libazureinit/src/logging.rs
@@ -78,7 +78,7 @@ pub struct Kvp<S: Subscriber> {
     /// The `JoinHandle` for the background task responsible for writing
     /// KVP data to the file. The caller can use this handle to wait for
     /// the writer to finish.
-    pub writer: JoinHandle<std::io::Result<()>>,
+    writer: JoinHandle<std::io::Result<()>>,
     shutdown: CancellationToken,
 }
 
--- a/libazureinit/src/logging.rs
+++ b/libazureinit/src/logging.rs
@@ -14,8 +14,8 @@ use tracing_subscriber::{
     fmt, layer::SubscriberExt, EnvFilter, Layer, Registry,
 };
 
+use crate::config::Config;
 use crate::kvp::Kvp;
-use libazureinit::config::Config;
 
 pub type LoggingSetup = (
     Box<dyn Subscriber + Send + Sync + 'static>,
--- a/libazureinit/src/logging.rs
+++ b/libazureinit/src/logging.rs
@@ -22,7 +22,7 @@ pub type LoggingSetup = (
     Option<JoinHandle<std::io::Result<()>>>,
 );
 
-pub fn initialize_tracing() -> sdktrace::Tracer {
+fn initialize_tracing() -> sdktrace::Tracer {
     let provider = SdkTracerProvider::builder()
         .with_sampler(Sampler::AlwaysOn)
         .build();
@@ -42,11 +42,11 @@ pub fn initialize_tracing() -> sdktrace::Tracer {
 ///   with `config`, adding file logging to `config.azure_init_log_path.path` or
 ///   falling back to `DEFAULT_AZURE_INIT_LOG_PATH` if unspecified.
 pub fn setup_layers(
-    tracer: sdktrace::Tracer,
     vm_id: &str,
     config: &Config,
     graceful_shutdown: CancellationToken,
 ) -> Result<LoggingSetup, anyhow::Error> {
+    let tracer = initialize_tracing();
     let otel_layer = OpenTelemetryLayer::new(tracer).with_filter(
         EnvFilter::try_from_env("AZURE_INIT_LOG")
             .unwrap_or_else(|_| EnvFilter::new("info")),
@@ -167,12 +167,11 @@ mod tests {
         config.azure_init_log_path.path = log_path.clone();
         config.telemetry.kvp_diagnostics = false;
 
-        let tracer = initialize_tracing();
         let vm_id = "test-vm-id-for-logging";
         let graceful_shutdown = CancellationToken::new();
 
         let (subscriber, _kvp_handle) =
-            setup_layers(tracer, vm_id, &config, graceful_shutdown.clone())
+            setup_layers(vm_id, &config, graceful_shutdown.clone())
                 .expect("Failed to setup layers");
 
         tracing::subscriber::with_default(subscriber, || {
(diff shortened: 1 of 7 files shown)


libazureinit/src/logging.rs
```diff
@@ -0,0 +1,264 @@
...
+/// to a specific location if `Some(&Config)` is provided.
+///
+/// This function follows a two-phase initialization:
+/// - Minimal Setup (Pre-Config): When called initially, it sets up basic logging
+///   to console (`stderr`), KVP (Hyper-V), and OpenTelemetry without file logging.
+///
+/// - Full Setup (Post-Config): After the configuration is loaded, it is called again
+///   with `config`, adding file logging to `config.azure_init_log_path.path` or
+///   falling back to `DEFAULT_AZURE_INIT_LOG_PATH` if unspecified.
+pub fn setup_layers(
```
jeremycline, line 44: I think providing this function as an easy path for users that don't otherwise have existing logging is okay, but I suspect that Afterburn and any other users may want to configure logging differently (different filters, or other synthetic event settings than `FmtSpan::NEW | FmtSpan::CLOSE` for example).

I'd recommend also exposing a way to get just the KVP stuff and leave it up to the caller to configure their tracing stack. Consider doing something like this (apologies for the length, I got a bit carried away here):
```Rust
struct KvpLayer<S: Subscriber>(Filtered<EmitKVPLayer, EnvFilter, S>);

/// Emit tracing data to the Hyper-V KVP.
///
/// # Example
///
/// ```no_run
/// # use libazureinit::logging::Kvp;
/// use tracing_subscriber::layer::SubscriberExt;
/// 
/// # #[tokio::main]
/// # async fn main() -> anyhow::Result<()> {
/// // In production scenarios, use libazureinit::get_vm_id() to get the VM ID
/// let mut kvp = Kvp::new("a-unique-id")?;
/// let registry = tracing_subscriber::Registry::default().with(kvp.layer());
///
/// // When it's time to shut down, doing this ensures all writes are flushed
/// kvp.halt().await?;
/// # Ok(())
/// # }
/// ```
pub struct Kvp<S: Subscriber> {
    layer: Option<KvpLayer<S>>,
    /// The `JoinHandle` for the background task responsible for writing
    /// KVP data to the file. The caller can use this handle to wait for
    /// the writer to finish.
    writer: JoinHandle<std::io::Result<()>>,
    shutdown: CancellationToken,
}

impl<S: Subscriber + for<'lookup> LookupSpan<'lookup>> Kvp<S> {

    pub fn new<T: AsRef<str>>(vm_id: T) -> Result<Self, anyhow::Error> {
        let shutdown = CancellationToken::new();
        let inner = crate::kvp::Kvp::new(
            std::path::PathBuf::from("/var/lib/hyperv/.kvp_pool_1"),
            vm_id.as_ref(),
            shutdown.clone(),
        )?;

        let kvp_filter = EnvFilter::builder().parse(
            [
                "WARN",
                "azure_init=INFO",
                "libazureinit::config::success",
                "libazureinit::http::received",
                "libazureinit::http::success",
                "libazureinit::ssh::authorized_keys",
                "libazureinit::ssh::success",
                "libazureinit::user::add",
                "libazureinit::status::success",
                "libazureinit::status::retrieved_vm_id",
                "libazureinit::health::status",
                "libazureinit::health::report",
            ]
            .join(","),
        )?;
        let layer = Some(KvpLayer(inner.tracing_layer.with_filter(kvp_filter)));

        Ok(Self {
            layer,
            writer: inner.writer,
            shutdown,
        })
    }

    /// Get a tracing [`Layer`] to use with a [`Registry`].
    ///
    /// # Panics
    ///
    /// If this function is called more than once.
    pub fn layer(&mut self) -> Filtered<EmitKVPLayer, EnvFilter, S> {
        assert!(
            self.layer.is_some(),
            "Kvp::layer cannot be called multiple times!"
        );
        self.layer.take().unwrap().0
    }

    /// Gracefully shut down the KVP writer.
    ///
    /// This will stop new KVP logs from being queued and wait for all pending writes to the KVP
    /// pool to complete.  After this returns, no further logs will be written to KVP.
    pub async fn halt(self) -> Result<(), anyhow::Error> {
        self.shutdown.cancel();
        self.writer.await??;
        Ok(())
    }
}
```

Also, with the above it's a bit weird since it immediately decomposes `crate::kvp::Kvp` and has very similar fields, but since that type is private, we could easily change that to be different later.

I don't _know_ that Afterburn would want to tweak settings, so if this is exactly what they want then no need to expose more stuff.

cadejacobson (approved):

libazureinit/src/logging.rs
```diff
@@ -31,6 +32,100 @@ fn initialize_tracing() -> sdktrace::Tracer {
...
+/// kvp.halt().await?;
+/// # Ok(())
+/// # }
+/// ```
+pub struct Kvp<S: Subscriber> {
+    layer: Option<KvpLayer<S>>,
+    /// The `JoinHandle` for the background task responsible for writing
+    /// KVP data to the file. The caller can use this handle to wait for
+    /// the writer to finish.
+    pub writer: JoinHandle<std::io::Result<()>>,
```
jeremycline, line 81: I don't think this should be a public field since the cancellation token is private. Anyone who awaits this writer will never complete since the type isn't clone-able and the only way to have the cancellation token used is via `halt()` which takes `self`, consuming the instance.

jeremycline (changes requested):

jeremycline (approved):

cadejacobson (approved):
lgtm!

PR #225, opened 2025-08-25
Make KVP Filter Configurable For External Projects Using libazureinit

peytonr18:
Currently, `libazureinit` has a hardcoded KVP (Key-Value Pair) filter that only captures events at the `WARN` level and above. This creates a limitation for external projects (like Afterburn) that want to use `libazureinit` as a dependency and configure their own tracing levels.

For example, if Afterburn wants to capture `DEBUG` level events through the KVP system (or have their own filters in general on what telemetry is emitted), they currently cannot override the default filter without modifying `libazureinit` source code.

## Solution
This PR introduces a configurable KVP filter via the `AZURE_INIT_KVP_FILTER` environment variable:
- If the environment variable is not set, the existing default behavior is preserved
- External projects can set their own filter rules (e.g., `AZURE_INIT_KVP_FILTER=debug` or `AZURE_INIT_KVP_FILTER=my_crate=debug,other_crate=info`)
- Invalid filter strings are logged as warnings and fall back to the default filter
- There are examples in the code documentation that show how other projects can use this feature.

##  Testing
Added unit tests that verify:
- Default filter behavior when no environment variable is set
- Environment variable override functionality
- Fallback to default when invalid filter is provided

This change enables `libazureinit` to be more flexible as a library dependency while maintaining its current behavior for the main azure-init project.

### DIFF
--- a/libazureinit/src/config.rs
+++ b/libazureinit/src/config.rs
@@ -232,6 +232,33 @@ pub struct Telemetry {
     /// these directives are parsed using `tracing_subscriber::EnvFilter` and
     /// applied to the KVP layer unless overridden by the `AZURE_INIT_KVP_FILTER`
     /// environment variable. When not set, defaults tailored for azure-init are used.
+    ///
+    /// **Precedence**: Environment variable `AZURE_INIT_KVP_FILTER` takes precedence
+    /// over this config value. If neither is set, azure-init-specific defaults are used.
+    ///
+    /// The value must be a string that follows the syntax for
+    /// `tracing_subscriber::EnvFilter`, which is a comma-separated list of
+    /// logging directives. For example: `warn,my_crate=debug`.
+    ///
+    /// ### Examples of acceptable values:
+    ///
+    /// - **Capture `INFO` level and above for all crates:**
+    ///   ```toml
+    ///   kvp_filter = "info"
+    ///   ```
+    ///
+    /// - **Capture `DEBUG` from your crate and `WARN` from others:**
+    ///   ```toml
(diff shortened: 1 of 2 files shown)


anhvoms:
We should add this configuration to libazureinit config

libazureinit/src/config.rs
```diff
@@ -223,15 +223,23 @@ impl Default for Wireserver {
...
+#[serde(default)]
 pub struct Telemetry {
     /// Flag to enable or disable KVP diagnostics. Defaults to `true`.
     pub kvp_diagnostics: bool,
+
+    /// Optional filter directives for the KVP tracing layer. When set,
+    /// these directives are parsed using `tracing_subscriber::EnvFilter` and
+    /// applied to the KVP layer unless overridden by the `AZURE_INIT_KVP_FILTER`
+    /// environment variable. When not set, defaults tailored for azure-init are used.
+    pub kvp_filter: Option<String>,
```
anhvoms, line 235: provide examples of acceptable values (can use similar examples as the ones found in kvp.rs)

anhvoms (approved):

cjp256:
A little late here, but my thoughts:

(a) removing env var option unless we can deem it necessary.  I'd avoid adding multiple ways to configure it if possible.  Work on making Config easily customized from a caller perspective.

(b) `std::thread::sleep(std::time::Duration::from_millis(50));` seems suspect and prone to race conditions.  If something needs flushing or waiting on, maybe there is a more precise operation we can do?

(c) the test strings look good to me, but ideally they would be assigned to a variable (written once) to avoid typos/sync issues.

PR #229, opened 2025-09-09
Adding password functionality to `libazureinit`

### LINKED ISSUE #52: [RFE] Disable provisioning with password

peytonr18:
## Goal
- Adding password functionality to `libazureinit` as part of the ongoing work with Flatcar/Afterburn. 

## Changes
- Implement password setting in `PasswordProvisioner` using `chpasswd` with piped stdin (no secrets in argv/logs)
- Preserve account locking when no password via passwd -l using `PATH_PASSWD` (which was part of existing functionality). 
- Add rustdoc to `password.rs` documenting behavior and usage example
- No change to `azure-init` binary behavior: it does not set passwords; external consumers can opt-in via `User::with_password(...)`

## Update unit tests:
- `password.rs`: mock-based tests for set and lock flows (no external commands)
- `mod.rs:` assert sshd update gating unchanged (true with `Passwd`, false with `FakePasswd`)
- `tests/cli.rs`: ensure CLI exposes no password-related flags

Resolves #52

### DIFF
--- a/libazureinit/src/provision/password.rs
+++ b/libazureinit/src/provision/password.rs
@@ -54,8 +54,11 @@ impl PasswordProvisioner {
 /// - If no password is provided, it disables password-based login by locking the
 ///   account's password using `passwd -l`.
 ///
-/// Note: While `libazureinit` supports password provisioning, the reference
-/// `azure-init` binary does not use this feature and only creates locked accounts.
+/// Reference `azure-init` behavior: it never calls `User::with_password`.
+/// Therefore `user.password` is `None`, and this function always locks the
+/// account via `passwd -l` (there is no alternate locking path). Library
+/// consumers that want a password must explicitly call `User::with_password`.
+/// See `doc/azure_init_behavior.md` for details.
 #[instrument(skip_all)]
 fn passwd(user: &User) -> Result<(), Error> {
     if let Some(ref password) = user.password {
--- a/libazureinit/src/provision/password.rs
+++ b/libazureinit/src/provision/password.rs
@@ -49,12 +49,12 @@ impl PasswordProvisioner {
 /// Manages the user's password during provisioning.
 ///
 /// This function supports two modes of operation:
-/// - If a password is provided in the `User` object, it sets the password securely 
+/// - If a password is provided in the `User` object, it sets the password securely
 ///   using the `chpasswd` utility via stdin to avoid exposing secrets.
-/// - If no password is provided, it disables password-based login by locking the 
+/// - If no password is provided, it disables password-based login by locking the
 ///   account's password using `passwd -l`.
 ///
-/// Note: While `libazureinit` supports password provisioning, the reference 
+/// Note: While `libazureinit` supports password provisioning, the reference
 /// `azure-init` binary does not use this feature and only creates locked accounts.
 #[instrument(skip_all)]
 fn passwd(user: &User) -> Result<(), Error> {
--- a/libazureinit/src/provision/password.rs
+++ b/libazureinit/src/provision/password.rs
@@ -58,6 +58,7 @@ use std::io::Write;
 use std::process::{Command, Stdio};
 
 use tracing::instrument;
+use zeroize::Zeroize;
 
 use crate::{error::Error, User};
 
@@ -88,7 +89,8 @@ impl PasswordProvisioner {
 ///
 /// # Security
 /// The password is passed securely to `chpasswd` via stdin to avoid
-/// exposing secrets in process arguments or logs.
+/// exposing secrets in process arguments or logs. The password is also
+/// securely cleared from memory after use using zeroization.
 #[instrument(skip_all)]
 pub fn set_user_password(user: &str, password: &str) -> Result<(), Error> {
     // Basic input validation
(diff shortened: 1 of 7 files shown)


cadejacobson (approved):
This looks correct to me! Password logic only applies to libazureinit, and azure-init broadly still blocks this feature.

anhvoms (approved):
LGTM!

libazureinit/src/provision/password.rs
```diff
@@ -21,30 +48,48 @@ impl PasswordProvisioner {
...
+/// - If a password is provided in the `User` object, it sets the password securely
+///   using the `chpasswd` utility via stdin to avoid exposing secrets.
+/// - If no password is provided, it disables password-based login by locking the
+///   account's password using `passwd -l`.
 ///
-/// If the `User` object contains a password, this function will return a
-/// `NonEmptyPassword` error, as provisioning with a password is not a
-/// supported feature.
+/// Note: While `libazureinit` supports password provisioning, the reference
+/// `azure-init` binary does not use this feature and only creates locked accounts.
```
cjp256, line 58:
```
only creates locked accounts.
```

on what condition? is it simply not specifying a user.password and always incurs -l? or not calling it at all and being locked by another mechanism?

fwiw I think it is fine that the API does _not_ document specific usage, but that should be documented somewhere else.   Let's consider starting an azure-init behavior doc?

peytonr18: For the sake of being thorough, I added a few lines to the API to better explain behavior, as well as an `azure_init_behavior.md` document. I do think this document will ultimately be ingested by Cade's documentation PR (See #233), but it's there as an independent doc in the meantime. 

The current behavior is: 
- Simply not specifying `User.password` is enough to invoke `-l`. During `Provision::provision`, the `PasswordProvisioner::Passwd` backend runs and if `user.password.is_none()`, it calls `passwd -l` to lock the account. 
- As for "not calling it at all", the binary always runs password provisioning. An external project could choose to omit password provisioning backends and then libazureinit would not perform the lock step. With the default config though, `passwd` is present and the lock mechanism occurs deterministically whenever `User.password` is absent.

peytonr18:
RE: @cjp256 
> consider breaking the API up into something like:
> 
> ` set_user_password(user, password)`
> 
> and
> 
> `lock_user(user)`
> 
> This would provide better building blocks without coupling specific behaviors (which can be also exported as additional interfaces to be used by flatcar and azure-init, and documented accordingly).
> 
> relatedly - flatcar's walinuxagent today does this for users with passwords:
> 
> ```
> $ cat /etc/ssh/sshd_config.d/80-flatcar-walinuxagent.conf
> PasswordAuthentication yes
> ChallengeResponseAuthentication yes
> ClientAliveInterval 180
> ```
> 
> without password specified:
> 
> ```
> $ cat /etc/ssh/sshd_config.d/80-flatcar-walinuxagent.conf
> PasswordAuthentication no
> ChallengeResponseAuthentication no
> ClientAliveInterval 180
> ```
> 
> The sshd config is not in scope for this PR but an example of coupled behaviors, where we maybe want to export equivalents if needed.

I wanted to circle back on this with a more in-depth comment about future design for SSHD config. 

#### Changes Made in this PR
I've implemented the decoupled password API building blocks as requested:
- **`set_user_password(user, password)`** - Sets password via `chpasswd`
- **`lock_user(user)`** - Locks account via `passwd -l`
- **Refactored `PasswordProvisioner`** to use these APIs internally
- **Maintained backward compatibility** - existing `Provision::provision()` works unchanged

#### SSH Config Changes - Not Yet Implemented
Per your comment about SSH config being "not in scope for this PR," I didn't make any SSH configuration changes yet. But, I think we should eventually do the following (whether in this PR or a follow up one) to finalize the SSH and password decoupling: 

**Current SSH behavior:** The provisioning flow automatically updates `/etc/ssh/sshd_config.d/50-azure-init.conf` (or `/etc/ssh/sshd_config`) to set `PasswordAuthentication yes/no` based on whether a password provisioner is configured. This happens immediately after password provisioning in `Provision::provision()`.

**Proposed SSH decoupling:** Move this SSH configuration logic out of the automatic provisioning flow and into a separate function/step. This would mean:
- `set_user_password()` and `lock_user()` have zero SSH impact
- SSH configuration becomes an independent choice for consumers
- WALinuxAgent/Flatcar can handle their own SSH config logic as needed
- The traditional `Provision::provision()` flow could still update SSH config, but as a separate step from the password logic.

libazureinit/src/provision/password.rs
```diff
@@ -19,48 +74,125 @@ impl PasswordProvisioner {
...
+///
+/// # Returns
+/// * `Ok(())` on success
+/// * `Err(Error)` if the account locking fails
 #[instrument(skip_all)]
-#[cfg(test)]
-fn mock_passwd(user: &User) -> Result<(), Error> {
-    if user.password.is_none() {
-        Ok(())
+pub fn lock_user(user: &str) -> Result<(), Error> {
```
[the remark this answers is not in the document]
cadejacobson: set_user_password also does this declaration of `user: str`

Issue #174, opened 2025-03-05
Unpin zerofrom and litemap when our MSRV is 1.81


jeremycline:
We had to pin both zerofrom and litemap, dependencies of ours via `reqwest`, to 0.1.5 and 0.7.4 respectively in https://github.com/Azure/azure-init/pull/173. This is because they started using `core::error` which stabilized in 1.81.

We should unpin these when our MSRV hits 1.81, which should be on or around September 5th, 2025.

PR #231, opened 2025-09-12
Cargo: pin sysinfo to 0.36

dongsupark:
Pin `sysinfo` to `=0.36` for MSRV issues with Rust < 1.85.0. This should be unpinned around 2026-02-17 if we continue with our ~1 year MSRV policy.

Without the pinning, CI with Rust 1.81 fails like:

```
Downloaded sysinfo v0.37.0
error: failed to parse manifest at `/home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/sysinfo-0.37.0/Cargo.toml`

Caused by:
  feature `edition2024` is required

  The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.81.0 (eeb90cda1 2024-09-04)).
  Consider trying a newer version of Cargo (this may require the nightly release).
  See https://doc.rust-lang.org/night
```

Also do not pin `litemap`, `zerofrom` any more, as the current MSRV for the last 1 year becomes 1.81.0, which does not require to pin specific versions of litemap or zerofrom.

Fixes https://github.com/Azure/azure-init/issues/174

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,5 +13,7 @@ updates:
         # stay with <= 1.2.0 for Rust 1.78
       - dependency-name: "litemap"
         # stay with <= 0.7.4 for Rust 1.76
+      - dependency-name: "sysinfo"
+        # stay with <= 0.36 for Rust < 1.85
       - dependency-name: "zerofrom"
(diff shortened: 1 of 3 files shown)


peytonr18:
Resolves #174

PR #234, opened 2025-09-23
Add --version flag to CLI

peytonr18:
### Summary:
Introduces dynamic support for the `--version` and `-V` flags in the Azure-init CLI, enabling real-time display of the current version. For development builds, the version string now includes Git metadata, and an optional packaging override allows for customized version output.

### How version is determined
At runtime, the version string is composed as:
1. `AZURE_INIT_VERSION` (if set by packaging; used verbatim), else
2. `CARGO_PKG_VERSION-VERGEN_GIT_DESCRIBE` (if available), else
3. `CARGO_PKG_VERSION-<short VERGEN_GIT_SHA>` (if available), else
4. `CARGO_PKG_VERSION`

Example outputs:
- Git checkout: `azure-init 0.1.1-v0.1.1-103-gf6729bb-dirty` or `azure-init 0.1.1-f6729bb`
- Packaging override: `azure-init 0.1.1-1ubuntu2404` or `azure-init 0.1.1-el8`
- No git/no override: `azure-init 0.1.1`

### Implementation details
- Use `vergen` + `vergen-gitcl` in `build.rs` to emit `VERGEN_GIT_*` env vars.
- Compute the final version string in `src/main.rs` at runtime (no leaks).
- Use a custom `--version`/`-V` flag (clap’s built-in version flag disabled) to print and exit cleanly.
- Logs and the HTTP User-Agent use the same computed version string.

### Usage
```bash
azure-init --version
# Output (git tree): azure-init 0.1.1-f6729bb
azure-init -V
# Output (packaging override): azure-init 0.1.1-1ubuntu2404
```
This aligns with the PR request to use `vergen` for including git details in dev builds and to optionally allow OS packaging to pass a version at build time.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -30,14 +30,15 @@ fn version_string() -> String {
     if let Some(v) = option_env!("AZURE_INIT_VERSION") {
         return v.to_string();
     }
+
     if let Some(desc) = option_env!("VERGEN_GIT_DESCRIBE") {
-        return format!("{PKG_VERSION}-{desc}");
-    }
-    if let Some(sha) = option_env!("VERGEN_GIT_SHA") {
+        format!("{PKG_VERSION}-{desc}")
+    } else if let Some(sha) = option_env!("VERGEN_GIT_SHA") {
         let short = &sha[..std::cmp::min(7, sha.len())];
-        return format!("{PKG_VERSION}-{short}");
+        format!("{PKG_VERSION}-{short}")
+    } else {
+        PKG_VERSION.to_string()
     }
-    PKG_VERSION.to_string()
 }
 
 /// Minimal provisioning agent for Azure
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,6 +63,5 @@ systemd_linux = ["passwd", "hostnamectl", "useradd"]
 default = ["systemd_linux"]
 
 [build-dependencies]
-# Use vergen to emit build, cargo, rustc, sysinfo, and git metadata
 vergen = { version = "9.0.6", features = ["build", "cargo", "rustc", "si"] }
 vergen-gitcl = "1"
--- a/build.rs
+++ b/build.rs
@@ -5,8 +5,15 @@ use vergen_gitcl::GitclBuilder;
 fn main() {
     // Re-run if the packaging version override changes
     println!("cargo:rerun-if-env-changed=AZURE_INIT_VERSION");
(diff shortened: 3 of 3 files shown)


libazureinit/src/lib.rs
```diff
@@ -1,5 +1,9 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT License.
+
+/// The version of the azure-init package
```
cadejacobson, line 4: libazureinit and azure-init can have unrelated versions due to them each having a variable in their respective cargo.toml so I would update this to 
```suggestion
/// The version of the libazureinit package
```

src/main.rs
```diff
@@ -25,7 +25,7 @@ use tracing::instrument;
 use tracing_subscriber::{prelude::*, Layer};
 
 // These should be set during the build process
-const VERSION: &str = env!("CARGO_PKG_VERSION");
+use libazureinit::VERSION;
```
cadejacobson, line 28: The previous comment brings up an interesting dilemma.

When running `azure-init --version`, are we expecting to see the azure-init version or the libazureinit version? Currently, the command for azure-init seems to output libazureinit's version.

These are technically independent versions since each has a version number found in Cargo.toml. If we are planning to keep these 2 values identical, this isn't an issue, but if we re-released libazureinit for the first time (version 1.0) and never updated azure-init's version from v0.1.1, it feels like this output could deviate.

cadejacobson: If possible, I think it would be best to leave `libazureinit::VERSION` populated in case anyone wants to access it from the library, but I think `azure-init --version` should output azure-init's version. Because of the open source nature of this project, we can then take the specific azure-init version and look back at the source code to find what specific libazureinit version is being used if that is ever needed for whatever reason.

peytonr18: Done!

cjp256 (approved):
Awesome, thanks!
```
cpatterson@devbox-eastus:~/git/azure-init$ cargo build
...
cpatterson@devbox-eastus:~/git/azure-init$ ./target/debug/azure-init --version
azure-init 0.1.1-4d96baf

cpatterson@devbox-eastus:~/git/azure-init$ AZURE_INIT_VERSION=99-mycustomversion cargo build
...
cpatterson@devbox-eastus:~/git/azure-init$ ./target/debug/azure-init --version
azure-init 99-mycustomversion
```

Issue #242, opened 2025-10-06
[RFE] Add improved customization to testinit responses to Azure-init


cadejacobson:
>At some point it would be nice to feed a "recipe/playbook" to the mock IMDS server (via a file passed to command line) and ask it to do stuff like
"Respond with 500 for the first request (test http 500 handling), sleep on the next request (test read timeout handling), respond with 404, respond with partial data, respond with full data"

>In normal mode you can make the server keep responding with good data, but in a "playbook" mode you can use it to test azure-init handling of various IMDS/wireserver condition

REF: Anh Vo in PR #230

PR #244, opened 2025-10-07
Add recipes for IMDS responses

cadejacobson:
## How to use
Customization of the E2E Testing pipeline can be done by creating or modifying a JSON file in the `/api-responses/imds/` directory. Once the order of responses is set, add `--imds-responses /path/to/response-file` as an argument to `python3 test_server.py` in the start_server_docker.sh script.

Example:
```bash
# Start the server and capture its PID
python3 test_server.py --imds-responses /api-responses/imds/internal-error-500.json &
SERVER_PID=$!
```

This PR also ensures that the script pipeline now waits for the azure-init service to wait for completion before running.

This work satisfies #245 and introduces the framework for half of issue #242.

## Testing done
This PR has been tested via the e2e testing pipeline in CI/CD.

### DIFF
--- a/testinit/testing-server/imds_handler.py
+++ b/testinit/testing-server/imds_handler.py
@@ -49,7 +49,7 @@ class IMDSHandler(BaseHTTPRequestHandler):
         response_body = current_response.get('response', {})
         self.wfile.write(json.dumps(response_body).encode())
 
-        logger.info(f"Returning response: {current_response}, from position: {self.__class__._response_position}")
+        logger.info(f"Returning response: {json.dumps(current_response, indent=2)}, from position: {self.__class__._response_position}")
         self.__class__._response_position += 1
         return
     


testinit/testing-server/api-responses/imds/timeout-408.json
```diff
@@ -0,0 +1,11 @@
+{
+  "responses": [
+    {
+      "status_code": 408,
```
anhvoms, line 4: can we simulate the more common timeout scenario where it's returning 200, but it is taking like 30 seconds (or we can pick a value of 5 seconds, but set azure-init imds timeout to 2s)

cadejacobson: Definitely! We do currently have the ability to set a universal IMDS or wireserver delay with environment variables so that every response takes at least that amount of time, but I could easily add an optional value in these JSON structures to induce a specific delay on any single response.

anhvoms: Yes, I would prefer it as a configurable option per response (an environment variable will affect all responses in the same playbook)

PR #247, opened 2025-10-09
Make set_hostname function available to libazureinit consumers

cadejacobson:
This PR exposes the set_hostname function to the the libazureinit package for third party consumption by projects that add the libazureinit crate. This PR also adds tests that specifically cover the expected behavior of the `set_hostname` function.

## How to use
No difference in usage.

## Testing done
This PR has been tested with the testinit process in the CI/CD pipeline.

### DIFF
--- a/libazureinit/src/provision/mod.rs
+++ b/libazureinit/src/provision/mod.rs
@@ -2,7 +2,7 @@
 // Licensed under the MIT License.
 pub mod hostname;
 pub mod password;
-pub(crate) mod ssh;
+pub mod ssh;
 pub mod user;
 
 use crate::config::{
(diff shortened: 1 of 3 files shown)


libazureinit/src/provision/mod.rs
```diff
@@ -338,8 +355,57 @@ mod tests {
             "create_user should fail with no user provisioners"
         );
         assert!(
-            matches!(result.unwrap_err(), error::Error::NoUserProvisioner),
+            matches!(result.unwrap_err(), Error::NoUserProvisioner),
             "Should return NoUserProvisioner error"
         );
     }
+
+    #[test]
```
KsenijaS: why are the tests in the same file as the code? Why are they not under tests folder?

cadejacobson: Good question. Rust is **_super_** picky about what code they allow you to access, even to the point that they actively compile documentation comments when you run `cargo build` to see if there are any pieces of code in them that should not be there. Your documentation can give compiler errors, which some Rust devs get shocked by. 

With that said, Cargo (Rust's compiler) also has expectations on where tests go. Their documentation expects unit tests and documentation tests to be `in the src files` and any broader integration tests should go into a `tests/` directory. We primarily have unit tests for this project, along with the recent addition of the testinit which lives outside of the Rust system, but by leaving the tests in the specific module they work on, we are better able to ensure access to the private functions that Rust would otherwise lock down. 
https://doc.rust-lang.org/cargo/guide/tests.html

PR #269, opened 2025-10-22
Expose the create-user function to libazureinit

cadejacobson:
This PR adds the public `create_user()` function which is now accessible to third party Rust projects for consumption.

## How to use
No usage changes.

## Testing done
This PR has been tested with the testinit process in the CI/CD pipeline.

### DIFF
--- a/libazureinit/src/provision/mod.rs
+++ b/libazureinit/src/provision/mod.rs
@@ -53,7 +53,7 @@ impl Provision {
     /// Returns [`Error::NoUserProvisioner`] if no user provisioner backends are
     /// configured or if all backends fail to create the user.
     #[instrument(skip_all)]
-    pub fn create_user(self) -> Result<(), Error> {
+    pub fn create_user(&self) -> Result<(), Error> {
         self.config
             .user_provisioners
             .backends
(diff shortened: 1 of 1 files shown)


libazureinit/src/provision/mod.rs
```diff
@@ -43,6 +43,31 @@ impl Provision {
...
+    /// the user with the first backend that succeeds. Currently supported
+    /// backends include:
+    /// - Useradd
+    ///
+    /// # Errors
+    ///
+    /// Returns [`Error::NoUserProvisioner`] if no user provisioner backends are
+    /// configured or if all backends fail to create the user.
+    #[instrument(skip_all)]
+    pub fn create_user(self) -> Result<(), Error> {
```
cjp256: why self and not &self?

Can you use &self and avoid the clone?

cadejacobson: Using a reference with &self or using a clone should both solve the issue here. In my personal dealings with Rust, I have run into more issues with maintaining variable lifetimes than I have with allocating too many deep copies, so I have used `.clone()` more than is maybe considered ideal as a method to really avoid that. 

Since this will be operating in systems with limited space, though, I think it's a good idea to go with &self to save any amount of bytes we can for other programs. Will get this updated.

peytonr18 (approved):
This looks good to me!

PR #286, opened 2026-02-19
Add configurable KVP client API and harden concurrent telemetry writes

### LINKED ISSUE #239: [RFE] Support Concurrent KVP Telemetry

peytonr18:
## Summary
Expands KVP telemetry beyond the original event-prefix change by introducing a public, external-caller-friendly KVP client API, while strengthening concurrency behavior for startup truncation and documenting the end-to-end design.

This enables multiple emitters/libraries to coexist more safely and gives callers explicit control over KVP emission lifecycle and identity.

### New public KVP client API (`kvp.rs`, re-exported via `logging.rs`)
- Added `KvpOptions` with builder-style configuration:
  - `vm_id`
  - `event_prefix`
  - `file_path`
  - `truncate_on_start`
- Added `Kvp::new()` (default options path).
- Added `Kvp::with_options(...)` for explicit caller configuration.
- Added direct emission APIs:
  - `emit(message, level)`
  - `emit_health_report(report)`
- Added `close().await` for graceful shutdown and queue drain.

### Internal wiring updates (`logging.rs`)
- `setup_layers` continues to use tracing-layer emission internally.
- Internal construction now goes through `Kvp::new_internal(...)` with explicit options-derived inputs.
- Existing azure-init tracing behavior remains intact.

### Concurrency hardening for startup truncation (`kvp.rs`)
- Truncation path now uses file locking to avoid check-then-truncate races.
- Lock contention during truncation is treated as non-fatal (expected in multi-client environments), allowing initialization to continue safely.

## Breaking Changes
This PR intentionally changes the public KVP API shape:
- Removed old subscriber-wrapper usage pattern (`Kvp<S>::new(...), .layer(), .halt()`).
- Replaced with direct client API (`Kvp::new/with_options`, `emit`, `emit_health_report`, `close`).

## Goal
- Make KVP emission straightforward for external callers without requiring `tracing_subscriber` plumbing.
- Support concurrent/multi-emitter scenarios more safely.
- Keep internal azure-init tracing integration while exposing a cleaner public surface.

Resolves #239

### DIFF
--- a/libazureinit/src/kvp.rs
+++ b/libazureinit/src/kvp.rs
@@ -66,6 +66,14 @@ pub struct Kvp {
 }
 
 impl Kvp {
+    /// Opens a KVP guest pool file for appending.
+    ///
+    /// All callers that need a file handle for KVP writes should go through
+    /// this method so that the open-mode is consistent everywhere.
+    fn open_kvp_file(path: &Path) -> io::Result<File> {
+        OpenOptions::new().append(true).create(true).open(path)
+    }
+
     /// Creates a new `Kvp` instance, spawning a background task for writing
     /// KVP telemetry data to a file.
     ///
@@ -82,10 +90,7 @@ impl Kvp {
     ) -> Result<Self, anyhow::Error> {
         truncate_guest_pool_file(&file_path)?;
 
-        let file = OpenOptions::new()
-            .append(true)
-            .create(true)
-            .open(&file_path)?;
+        let file = Self::open_kvp_file(&file_path)?;
 
         let (events_tx, events_rx): (
             UnboundedSender<Vec<u8>>,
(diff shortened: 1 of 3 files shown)


cadejacobson (approved):
This looks awesome to me! Good work

libazureinit/src/kvp.rs
```diff
@@ -532,37 +555,50 @@ pub fn decode_kvp_item(
...
+                .context("Failed to open KVP file for truncation"));
         }
-    }
+    };
+
+    // Hold an exclusive lock for the metadata-check + truncate window so
+    // that two concurrent callers cannot both decide the file is stale and
+    // truncate data the other has already written.
+    FileExt::lock_exclusive(&file).map_err(|e| {
+        anyhow::Error::from(e).context("Failed to lock KVP file for truncation")
```
anhvoms: if this is expected (that only one client is expected to do the truncation), this should not be an error. We can easily conclude an error situation by seeing all clients failing, but no successful truncation (from the log a few lines below)

PR #288, opened 2026-03-09
feat(kvp): add store trait and Hyper-V KVP storage crate

peytonr18:
## What this PR introduces

- Adds workspace crate `libazureinit-kvp`.
- Adds `libazureinit-kvp/src/lib.rs`, `libazureinit-kvp/src/store.rs`, and `libazureinit-kvp/src/error.rs`.
- Adds crate dependencies: `libc`, `tracing`. Dev-dependencies: `rstest`, `tempfile`.
- This is step 1 of the abstraction — the crate exists and is exercised by its own tests, but is not yet wired into `libazureinit`. A follow-up PR will replace `libazureinit/src/kvp.rs` with this implementation.

## KVP design

Single concrete type — no trait indirection in the public API.

- `KvpPoolStore`: file-backed KVP pool store.
- `KvpPool`: enum identifying one of the five Hyper-V pool indices.
- `PoolMode`: policy for write-path size limits.
- `KvpError`: error type.

### `KvpPoolStore` API

Construction:
- `KvpPoolStore::new(pool, mode)` — uses the default directory `/var/lib/hyperv`.
- `KvpPoolStore::new_in(pool, dir, mode)` — uses a caller-supplied directory (file name derived from `pool`).

Reads / queries (do not enforce mode size limits on the key argument; a `Safe`-mode store can read keys written in `Unsafe` mode):
- `read(key) -> Result<Option<String>, KvpError>` — last-write-wins on duplicate keys.
- `entries() -> Result<HashMap<String, String>, KvpError>` (deduplicated, last-write-wins).
- `dump() -> Result<Vec<(String, String)>, KvpError>` (on-disk order, preserves duplicates).
- `len() -> Result<usize, KvpError>` (on-disk record count, not unique keys).
- `is_empty() -> Result<bool, KvpError>`
- `is_stale() -> Result<bool, KvpError>`

Writes (enforce `PoolMode` key and value size limits):
- `insert(key, value) -> Result<(), KvpError>` — upsert; overwrites first match in place, removes any further duplicates, and enforces the 1024 unique-key cap for new keys.
- `append(key, value) -> Result<(), KvpError>` — blind append; no duplicate check, no unique-key cap. Intended for log/telemetry workloads.
- `populate(records) -> Result<(), KvpError>` — atomic replace-all. Validates the full batch (including the unique-key cap) before locking, then truncates and rewrites under one exclusive lock. Preserves duplicates in iteration order. Inverse of `dump`.
- `delete(key) -> Result<bool, KvpError>` — removes all records matching the key; returns `true` if any were present.
- `clear() -> Result<(), KvpError>` — truncates the file.
- `clear_if_stale() -> Result<(), KvpError>` — reads boot time, takes the exclusive lock, then truncates only if the file's mtime is `<=` boot time. The check happens under the write lock so a concurrent writer can't extend mtime between the staleness check and the truncate.

Metadata accessors:
- `pool() -> KvpPool`
- `mode() -> PoolMode`
- `path() -> &Path`
- `max_key_size() -> usize`
- `max_value_size() -> usize`

## Pool identity

`KvpPool` enumerates the five standard Hyper-V KVP pool indices:

| Variant | File | Role |
|---|---|---|
| `External` | `.kvp_pool_0` | Host-to-guest, pushed by host admin |
| `Guest` | `.kvp_pool_1` | Guest-to-host; where cloud-init / azure-init write |
| `Auto` | `.kvp_pool_2` | Guest intrinsics generated by the daemon |
| `AutoExternal` | `.kvp_pool_3` | Host-originated data describing the host |
| `AutoInternal` | `.kvp_pool_4` | Undocumented; no pool file exists |

`KvpPool::default_dir()` returns `/var/lib/hyperv`. `KvpPool::default_path()` returns the full default path for a pool.

## On-disk format and limits

Hyper-V wire format: fixed-size records, identical across environments.

- key field: 512 bytes (zero-padded UTF-8)
- value field: 2048 bytes (zero-padded UTF-8)
- record size: 2560 bytes
- maximum unique keys: 1024 (enforced by `insert` and `populate`, not `append`)

All sizes are UTF-8 byte counts measured against the Linux on-disk layout. The Windows-side Hyper-V spec expresses limits in UTF-16 code units (256 / 1024), which equal the 512 / 2048-byte fields used by `hv_kvp_daemon` on Linux.

`PoolMode` controls write-side size limits:

| Mode | Max key | Max value | Notes |
|---|---|---|---|
| `Safe` | 254 bytes | 1022 bytes | Recommended for Linux kernel compatibility (2 bytes under the kernel `HV_KVP_EXCHANGE_MAX_*` maximums to leave room for a NUL terminator on either side of the boundary) |
| `Unsafe` | 512 bytes | 2048 bytes | Full wire-format maximums |

Read-side operations do not enforce these limits, so a `Safe`-mode store can read or delete keys written in `Unsafe` mode.

## Locking: both `flock` and `fcntl` OFD

The store takes **both** `flock` and `fcntl` open file description (OFD) locks for every operation:

- Acquire `flock` first (`LOCK_EX` for writes, `LOCK_SH` for reads). If the subsequent `fcntl` lock fails, the `flock` is released before returning the error (all-or-nothing acquisition).
- Then `fcntl(F_OFD_SETLKW, ...)` with `F_WRLCK` for writes or `F_RDLCK` for reads, range `l_start = 0, l_len = 0` (whole file).
- Linux `flock(2)` and `fcntl(2)` byte-range locks live in **disjoint kernel domains** and do not coordinate with each other. `hv_kvp_daemon` uses `fcntl`; cloud-init uses `flock(LOCK_EX)`. Holding both is the only way to coordinate with either.
- OFD `fcntl` locks (`F_OFD_SETLKW`) share a domain with classic POSIX `fcntl` byte-range locks but are per-fd and thread-safe — compatible with `hv_kvp_daemon` while remaining safe across threads sharing a process.
- Successful locks are released when the file descriptor closes (via `Drop` on the handle), so there is no explicit unlock in normal paths.

## File I/O behavior

Standard file I/O via `OpenOptions`, `read_exact`, `write_all`, `set_len`, `seek`, `flush`.

- `open_read_write_create()` sets `.truncate(false)` to avoid implicit truncation on open and `.mode(0o600)` so newly created pool files are private to the owning user.
- Iteration is handled by an internal `KvpPoolIter` that holds the file and keeps the lock for its lifetime, and supports in-place value overwrite and swap-remove. The lock is released when the iterator is dropped.

Flows:

- `insert` — open read+write+create, take exclusive locks, iterate records (collecting unique keys into a set). If the key exists, overwrite the first match in place and remove any further duplicates via swap-with-last + truncate. If the key is new, enforce the 1024 unique-key cap and append at EOF. Flush; locks released on drop.
- `append` — open read+write+create, take exclusive locks, seek to EOF, write a new record. No duplicate check, no unique-key cap. Flush.
- `populate` — validate the full batch up front (key/value sizes and unique-key cap), then open read+write+create, take exclusive locks, `set_len(0)`, and write every record. An I/O error mid-write may leave the file partially written.
- `delete` — open read+write, take exclusive locks, iterate records. For each match, swap with the final record and truncate by one. Flush only if at least one record was removed.
- `clear` — open read+write, take exclusive locks, `set_len(0)`.
- `read` / `entries` / `dump` — open read-only, take shared locks, decode records.

All missing-file paths return well-defined sentinels (`None`, `false`, empty collections, `0`, `false`) instead of erroring.

Malformed-file handling:

- A file whose size is not a multiple of the 2560-byte record size is rejected with an `Io` error from `record_count_from_len` on any operation that opens it for iteration or calls `len()`. Recover by calling `clear()`.
- Records whose key or value fields are not valid UTF-8 surface as `Io(InvalidData)` from `decode_record`. Trailing zero bytes in either field are trimmed during decode.

## Stale-data handling

- `is_stale()` compares the pool file's mtime to the system boot time, derived by parsing the `btime` field of `/proc/stat`.
- `clear_if_stale()` is a convenience that clears the file only when stale.
- Stale detection is explicit and caller-controlled — no automatic truncation occurs during construction.

## Validation

Key and value validation are implemented as free functions:

- `validate_key(key, max)` — rejects empty keys, null bytes, and lengths greater than `max`.
- `validate_value(value, max)` — rejects null bytes and lengths greater than `max`.

`insert`, `append`, and `populate` validate using limits derived from the store's `PoolMode`. `read` and `delete` validate the key's content (non-empty, no nulls) but pass `usize::MAX` for the size cap so they don't reject otherwise-valid keys that were written in a different `PoolMode`.

## Error model

`KvpError` variants:

- `EmptyKey`
- `KeyContainsNull`
- `KeyTooLarge { max, actual }`
- `ValueContainsNull`
- `ValueTooLarge { max, actual }`
- `MaxUniqueKeysExceeded { max }`
- `Io(io::Error)`

Implements `Display`, `Error` (with `source()` chaining to the inner `io::Error`), and `From<io::Error>`.

## Testability: internal `SysOps` / `Handle` traits

To exercise error paths that are difficult to trigger with real syscalls (lock failures, mid-write I/O errors, transient filesystem failures, deterministic boot times), the store is implemented against two internal traits:

- `SysOps` — filesystem opens, `path_metadata`, `boot_time`.
- `Handle` — per-file `read`, `write`, `seek`, `set_len`, `metadata`, `flock_*`, `fcntl_*`.

`OsSysOps` / `OsHandle` are the production implementations and are constructed by default by `new` / `new_in`. Tests can substitute a fault-injecting backend via the crate-private `KvpPoolStore::with_ops` constructor. The public API exposes only the concrete `KvpPoolStore` (which is `Clone + Debug`; the inner `SysOps` is held behind an `Arc`).

## Test coverage

Comprehensive unit tests cover validation, lock interactions, fault injection via the `SysOps` / `Handle` traits, concurrent reader/writer scenarios, stale-data detection, and all documented error paths. 100% line coverage on Codecov.

image


cadejacobson (changes requested):

libazureinit-kvp/src/lib.rs
```diff
@@ -0,0 +1,121 @@
...
+
+pub use hyperv::HyperVKvpStore;
+
+/// Hyper-V key limit in bytes (policy/default preset).
+pub const HYPERV_MAX_KEY_BYTES: usize = 512;
+/// Hyper-V value limit in bytes (policy/default preset).
+pub const HYPERV_MAX_VALUE_BYTES: usize = 2048;
+/// Azure key limit in bytes (policy/default preset).
+pub const AZURE_MAX_KEY_BYTES: usize = 512;
+/// Azure value limit in bytes (UTF-16: 511 characters + null terminator).
```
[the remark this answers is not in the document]
peytonr18: I was basing this off of what we have in our existing `kvp.rs` -- 

```
const HV_KVP_EXCHANGE_MAX_KEY_SIZE: usize = 512;
const HV_KVP_EXCHANGE_MAX_VALUE_SIZE: usize = 2048;
const HV_KVP_AZURE_MAX_VALUE_SIZE: usize = 1022;
```

and my old notes from that KVP process, but I'll look for the official documentation where it says that!

libazureinit-kvp/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+
+use std::collections::HashMap;
+use std::io;
+
+pub mod hyperv;
+
+pub use hyperv::HyperVKvpStore;
+
+/// Hyper-V key limit in bytes (policy/default preset).
+pub const HYPERV_MAX_KEY_BYTES: usize = 512;
```
cjp256: shouldn't be here (dupe of hyperv)

libazureinit-kvp/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+pub const HYPERV_MAX_VALUE_BYTES: usize = 2048;
+/// Azure key limit in bytes (policy/default preset).
+pub const AZURE_MAX_KEY_BYTES: usize = 512;
+/// Azure value limit in bytes, matching Azure host behavior.
+pub const AZURE_MAX_VALUE_BYTES: usize = 1022;
+
+/// Storage abstraction for KVP backends.
+///
+/// Semantics:
+/// - `write`: stores one key/value or returns validation/I/O error.
```
cjp256: these details should be consolidated into the function docs.  This should just be a high level doc for the trait.

e.g. key-value store that supports the semantics of Hyper-V's KVP while also being generic to support non-file-backed-pool-KVP in-memory implementation, for tests etc.

libazureinit-kvp/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+    /// The key and value byte-size limits for this store.
+    ///
+    /// Consumers (e.g. diagnostics, tracing layers) should call this
+    /// instead of hardcoding size constants, so the limits stay correct
+    /// regardless of the underlying implementation.
+    fn limits(&self) -> KvpLimits;
+
+    /// Write a key-value pair into the store.
+    ///
+    /// Returns an error if:
```
cjp256: these errors should be defined here so we can be specific about what is being returned under what conditions

libazureinit-kvp/src/hyperv.rs
```diff
@@ -0,0 +1,649 @@
...
+
+//! Hyper-V-backed [`KvpStore`] implementation.
+//!
+//! ## Record format
+//! - Fixed-size records of [`RECORD_SIZE`] bytes (512-byte key +
+//!   2,048-byte value), zero-padded on disk.
+//! - No record-count header or explicit record cap in this layer.
+//!
+//! ## Behavior summary
+//! - **`write`**: append-only; one record appended per call.
```
cjp256: Keep the module docs relevant to the Hyper-V KVP pool / protocol and don't include any rust implementations details. The implementation should flow from this description and be detailed in the impl docs.

libazureinit-kvp/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+    /// - The key is empty.
+    /// - The key exceeds the configured maximum key size.
+    /// - The value exceeds the configured maximum value size.
+    /// - An I/O error occurs during the write.
+    fn write(&self, key: &str, value: &str) -> io::Result<()>;
```
cjp256: we'll actually want there to be an implementation here with common validations and then a call-out to the backend-processing.

for example:

```rust
trait KvpStore {
    const MAX_KEY_SIZE: usize;
    const MAX_VALUE_SIZE: usize;

    fn backend_read(&self, key: &str) -> Result<String, KvpError>;
    fn backend_write(&self, key: &str, value: &str) -> Result<(), KvpError>;

    fn read(&self, key: &str) -> Result<String, KvpError> {
        Self::validate_key(key)?;
        self.backend_read(key)
    }

    fn write(&self, key: &str, value: &str) -> Result<(), KvpError> {
        Self::validate_key(key)?;
        Self::validate_value(value)?;
        self.backend_write(key, value)
    }

    fn validate_key(key: &str) -> Result<(), KvpError> {
        let actual = key.len();
        let max = Self::MAX_KEY_SIZE;

        if actual > max {
            return Err(KvpError::KeyTooLarge { max, actual });
        }

        Ok(())
    }

    fn validate_value(value: &str) -> Result<(), KvpError> {
        let actual = value.len();
        let max = Self::MAX_VALUE_SIZE;

        if actual > max {
            return Err(KvpError::ValueTooLarge { max, actual });
        }

        Ok(())
    }
}
```

libazureinit-kvp/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+    ///
+    /// Returns `true` if at least one record was removed, `false` if
+    /// the key was not found.
+    fn delete(&self, key: &str) -> io::Result<bool>;
+}
+
+/// Key/value byte limits for write validation.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct KvpLimits {
+    pub max_key_size: usize,
```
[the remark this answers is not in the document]
cjp256: if you want to keep KvpLimits, make it a trait and implement Azure & HyperV variants

libazureinit-kvp/src/kvp_pool.rs
```diff
@@ -0,0 +1,922 @@
...
+use std::os::unix::io::AsRawFd;
+use std::path::{Path, PathBuf};
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+
+use sysinfo::System;
+
+use crate::{KvpError, KvpStore};
+
+/// Default KVP pool path when none is provided.
+pub const DEFAULT_KVP_POOL_PATH: &str = "/var/lib/hyperv/.kvp_pool_1";
```
bpryan99: Is there any chance for multiple KVP pools where filename would need to be distinct? (ex. kvp_pool_2)

libazureinit-kvp/src/kvp_pool.rs
```diff
@@ -0,0 +1,922 @@
...
+    }
+
+    fn pool_is_stale(&self) -> Result<bool, KvpError> {
+        let metadata = match std::fs::metadata(&self.path) {
+            Ok(m) => m,
+            Err(ref e) if e.kind() == ErrorKind::NotFound => return Ok(false),
+            Err(e) => return Err(e.into()),
+        };
+        let boot = Self::boot_time()?;
+        Ok(metadata.mtime() < boot)
```
bpryan99: Is this function intended to return true when metadata timestamp is less than time since boot in order to consider the pool stale?

libazureinit-kvp/src/store.rs
```diff
@@ -0,0 +1,922 @@
...
+    }
+
+    #[cfg(test)]
+    fn pool_is_stale_at_boot(&self, boot_time: i64) -> Result<bool, KvpError> {
+        let metadata = match std::fs::metadata(&self.path) {
+            Ok(m) => m,
+            Err(ref e) if e.kind() == ErrorKind::NotFound => return Ok(false),
+            Err(e) => return Err(e.into()),
+        };
+        Ok(metadata.mtime() < boot_time)
```
bpryan99: What's the functional difference between this function and the above function if both return on the evaluation of metadata time vs boot time? Are both needed?

libazureinit-kvp/src/kvp_pool.rs
peytonr18: Technically other filenames exist (.kvp_pool_2 does sometimes exist), but the only one hyperV looks at is .kvp_pool_1. I gave the option to accept a different pool path in case the user wants to not use hyperV and just do their own thing but otherwise, the only filename we would want to target for HyperV ingestion is pool_1. 

Let me know if that answers your question!!

libazureinit-kvp/src/kvp_pool.rs
peytonr18: Yes! 
The boot timestamp is compared to the `,etadata.mtime` and if `mtime >= boot_time`, then the file is treated as current. If not, it's stale!

libazureinit-kvp/src/kvp_pool.rs
```diff
@@ -0,0 +1,922 @@
...
+                records.into_iter().collect();
+
+            if !map.contains_key(key) && map.len() >= MAX_UNIQUE_KEYS {
+                return Err(KvpError::MaxUniqueKeysExceeded {
+                    max: MAX_UNIQUE_KEYS,
+                });
+            }
+
+            map.insert(key.to_string(), value.to_string());
+            rewrite_file(&mut file, &map)?;
```
bpryan99: If the map contains all records in this case, the whole file would be rewritten. Is this necessary/cause any unneeded latency?

libazureinit-kvp/src/store.rs
peytonr18: `boot_is_stale(&self)` is the production path that computes the boot time internally via `Self::boot_time()` and then compares it to the metadata. 

`pool_is_stale_at_boot(&self, boot_time: i64` is a test only helper that takes in the boot time as an argument so that the tests can inject a fixed value and avoid the issue of not having a real system clock/uptime. 

So, they functionally do the same comparison, one is just for the testing env. But, if this is confusing, I could extract the shared logic into a helper function!!

libazureinit-kvp/src/kvp_pool.rs
peytonr18: Great question!!! Chris and I talked about this at length offline, and I ran some small tests to evaluate some of the other approaches (memory mapping, indexed in-place inserts or key updates, etc). but based on those tests, I thought this was the best approach when thinking about speed and safety. 

The good news is because there is a hard-cap of 1024 records, the rewrite cost is bounded. I think the idea was that we start with this but if we start seeing latency issues, we can go to a more complex approach with memory mapping or whatever alternative.

libazureinit-kvp/src/kvp_pool.rs
bpryan99: Makes complete sense, thanks!

libazureinit-kvp/src/store.rs
bpryan99: Makes sense to me -- I noticed the separate var for boot time in the latter, but just needed clarification on intent. Maybe a comment/docstring would be helpful but I wouldn't ask for functional changes here.

libazureinit-kvp/src/kvp_pool.rs
bpryan99: Great, as long as it was discussed/tested -- It's not a whole lot to ever write given the record cap plus k/v size limits, but since this is involved in provisioning, I wanted to be sure we had considered even minor latency factors.

bpryan99 (approved):
My comments were addressed, LGTM! Great work.

libazureinit-kvp/src/kvp_pool.rs
```diff
@@ -0,0 +1,1498 @@
...
+        buf[..len].copy_from_slice(&bytes[..len]);
+
+        self.file.write_all(&buf)?;
+        Ok(())
+    }
+
+    /// Append a new record at the end of the file, zero-padded to the
+    /// fixed-width wire format.
+    fn append(&mut self, key: &str, value: &str) -> io::Result<()> {
+        self.file.seek(io::SeekFrom::End(0))?;
```
cjp256: we could do an assertion  here to ensure we are properly aligned

libazureinit-kvp/src/kvp_pool.rs
```diff
@@ -0,0 +1,1498 @@
...
+        Ok(())
+    }
+
+    fn append(&self, key: &str, value: &str) -> Result<(), KvpError> {
+        self.validate_key(key)?;
+        self.validate_value(value)?;
+
+        let mut iter = self.iter_mut()?;
+        iter.append(key, value)?;
+        iter.flush()?;
```
cjp256: the flush story needs to be worked out.  if kvpd will read from this file while we hold the lock, then every write should probably be followed up by flush.   if not, then maybe we only flush before we drop the lock

doc/kvp_pr_description.md
```diff
@@ -0,0 +1,74 @@
+# PR: Add `libazureinit-kvp` with unified KVP pool store
```
cjp256: file seems out of date - is it going to be deleted?

libazureinit-kvp/src/kvp_pool.rs
```diff
@@ -0,0 +1,1498 @@
...
+use std::os::unix::io::AsRawFd;
+use std::path::{Path, PathBuf};
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+
+use sysinfo::System;
+
+use crate::{KvpError, KvpStore};
+
+/// Default KVP pool path when none is provided.
+pub const DEFAULT_KVP_POOL_PATH: &str = "/var/lib/hyperv/.kvp_pool_1";
```
cjp256: let's do an enum for these, see https://github.com/Azure/azure-init/pull/290/changes#diff-0e11ae6109310d637614a88f49107f5f77f3f8adb457d64e32d83cf37b6875e0R64

libazureinit-kvp/src/kvp_pool.rs
```diff
@@ -0,0 +1,1498 @@
...
+    if unsafe { libc::fcntl(file.as_raw_fd(), libc::F_OFD_SETLK, &fl) } == -1 {
+        return Err(io::Error::last_os_error());
+    }
+    Ok(())
+}
+
+/// Policy mode controlling key/value size limits for writes.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum PoolMode {
+    /// Conservative limits for Azure compatibility
```
cjp256: the limits are for linux kernel compatibility

libazureinit-kvp/src/kvp_pool.rs
```diff
@@ -0,0 +1,1498 @@
...
+    /// (key <= 254 bytes, value <= 1022 bytes).
+    Restricted,
```
cjp256: let's call this "Safe"

libazureinit-kvp/src/kvp_pool.rs
```diff
@@ -0,0 +1,1498 @@
...
+    /// Full Hyper-V wire-format limits
+    /// (key <= 512 bytes, value <= 2048 bytes).
+    Full,
```
cjp256: let's call this "Unsafe"

libazureinit-kvp/src/kvp_pool.rs
cjp256: we might be able to figure out how to detect the kernel's limits and set behavior for Safe accordingly

libazureinit-kvp/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+    fn max_value_size(&self) -> usize;
+
+    /// Insert a new key-value pair or update an existing key's value.
+    fn insert(&self, key: &str, value: &str) -> Result<(), KvpError>;
+
+    /// Append a key-value pair without checking for an existing key.
+    fn append(&self, key: &str, value: &str) -> Result<(), KvpError>;
+
+    /// Read the value for a key. Returns `Ok(None)` when absent.
+    fn read(&self, key: &str) -> Result<Option<String>, KvpError>;
```
cjp256: this should document behavior if multiple matches for "key" are found.  (last one wins)

libazureinit-kvp/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    /// Remove a key. Returns `true` if the key was present.
+    fn delete(&self, key: &str) -> Result<bool, KvpError>;
```
cjp256: similarly document behavior if multiple matches are found (all are delted)

libazureinit-kvp/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    /// Remove all entries from the store.
+    fn clear(&self) -> Result<(), KvpError>;
+
+    /// Return all key-value pairs.
+    fn entries(&self) -> Result<HashMap<String, String>, KvpError>;
```
cjp256: can we get an interface that provides a list of key-values, so we can validate when multiple keys are present?

it'd be a useful interface to test with.

e.g.

append("key", "val1")
append("key", "val2")
append("key", "val3")
assert entries() == {"key": "val3"}
assert dump() == [("key", "val1"), ("key", "val2"), ("key", "val3")]

libazureinit-kvp/src/kvp_pool.rs
```diff
@@ -0,0 +1,1498 @@
...
+    fn is_empty(&self) -> Result<bool, KvpError> {
+        match std::fs::metadata(&self.path) {
+            Ok(m) => Ok(m.len() == 0),
+            Err(ref e) if e.kind() == ErrorKind::NotFound => Ok(true),
+            Err(e) => Err(e.into()),
+        }
+    }
+
+    fn is_stale(&self) -> Result<bool, KvpError> {
+        self.pool_is_stale()
```
cjp256: i think pool_is_stale() can be moved into here. no need for the jump and "pool" is implied

libazureinit-kvp/src/kvp_pool.rs
```diff
@@ -0,0 +1,1498 @@
...
+    fn len(&self) -> Result<usize, KvpError> {
+        match std::fs::metadata(&self.path) {
+            Ok(m) => Ok(m.len() as usize / RECORD_SIZE),
+            Err(ref e) if e.kind() == ErrorKind::NotFound => Ok(0),
+            Err(e) => Err(e.into()),
+        }
+    }
+
+    fn is_empty(&self) -> Result<bool, KvpError> {
+        match std::fs::metadata(&self.path) {
```
cjp256: would it make sense to check len() == 0?

libazureinit-kvp/src/kvp_pool.rs
```diff
@@ -0,0 +1,1498 @@
...
+    fn pool_is_stale_at_boot(&self, boot_time: i64) -> Result<bool, KvpError> {
+        let metadata = match std::fs::metadata(&self.path) {
+            Ok(m) => m,
+            Err(ref e) if e.kind() == ErrorKind::NotFound => return Ok(false),
+            Err(e) => return Err(e.into()),
+        };
+        Ok(metadata.mtime() < boot_time)
+    }
+
+    fn truncate_pool(&self) -> Result<(), KvpError> {
```
cjp256: "_pool" is implied, maybe just remove it

cjp256: actually maybe just move all this into clear

doc/kvp_pr_description.md
peytonr18: Yes, totally did not mean to include that!

libazureinit-kvp/src/kvp_pool.rs
peytonr18: In the latest commit, I made it so we flush before dropping the lock instead of after every write. insert(), append(), and delete() now also flush once at the end of the operation while they still hold the lock.

libazureinit-kvp/src/kvp_pool.rs
peytonr18: I can look more into detecting kernel limits, maybe in a future PR?

libazureinit-kvp/src/store.rs
```diff
@@ -29,8 +29,42 @@ use sysinfo::System;
...
-pub const DEFAULT_KVP_POOL_PATH: &str = "/var/lib/hyperv/.kvp_pool_1";
+/// Default base directory for KVP pool files.
+const DEFAULT_KVP_DIR: &str = "/var/lib/hyperv";
+
+/// Hyper-V KVP pool indices.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[repr(u8)]
+pub enum KvpPool {
+    /// Host-to-guest data pushed by the host admin (`.kvp_pool_0`).
+    External = 0,
```
cjp256, line 53: these should be the file paths so we don't new two params to new()

libazureinit-kvp/src/kvp_pool.rs
```diff
@@ -607,18 +651,33 @@ impl KvpStore for KvpPoolStore {
...
-            Ok(m) => Ok(m.len() == 0),
-            Err(ref e) if e.kind() == ErrorKind::NotFound => Ok(true),
-            Err(e) => Err(e.into()),
-        }
+        Ok(self.len()? == 0)
     }
 
     fn is_stale(&self) -> Result<bool, KvpError> {
-        self.pool_is_stale()
+        self.is_stale_inner()
```
cjp256: is_stale_inner() should be moved into is_stale()

libazureinit-kvp/src/kvp_pool.rs
```diff
@@ -128,21 +162,24 @@ pub struct KvpPoolStore {
...
+    /// - `path`: override pool file path; defaults to
+    ///   [`KvpPool::default_path`].
     /// - `mode`: size-limit policy (see [`PoolMode`]).
     /// - `truncate_on_stale`: if `true`, clears stale data from a
     ///   previous boot.
     pub fn new(
+        pool: KvpPool,
         path: Option<PathBuf>,
         mode: PoolMode,
         truncate_on_stale: bool,
```
cjp256: perhaps clear_if_stale?

libazureinit-kvp/src/store.rs
peytonr18: There are kinda two ways to do this, depending if you want KvpPoolStore to be pool oriented or path oriented. 
 
The first is this one, which just does an impl and new() will just take in the path (or a custom path if the user wants instead)
 
```
libazureinit-kvp/src/kvp_pool.rs
/// Hyper-V KVP pool indices.
#[derive(Clone, Copy, Debug, PartialEq, Eq)]
#[repr(u8)]
pub enum KvpPool {
   /// Host-to-guest data pushed by the host admin (`.kvp_pool_0`).
   External = 0,
   /// Guest-to-host data; cloud-init / azure-init write here (`.kvp_pool_1`).
   Guest = 1,
   // ...
}
impl From<KvpPool> for PathBuf {
   fn from(pool: KvpPool) -> Self {
       pool.default_path()
   }
}
impl KvpPoolStore {
   /// Create a new KVP pool store.
   ///
   /// - `path`: pool file path. Pass a [`KvpPool`] variant for a
   ///   well-known pool (e.g. `KvpPool::Guest`), or any
   ///   [`PathBuf`]/[`&Path`] for a custom location.
   /// - `mode`: size-limit policy (see [`PoolMode`]).
   /// - `clear_if_stale`: if `true`, clears stale data from a
   ///   previous boot.
   pub fn new(
       path: impl Into<PathBuf>,
       mode: PoolMode,
       clear_if_stale: bool,
   ) -> Result<Self, KvpError> {
       let store = Self {
           path: path.into(),
           mode,
       };
       if clear_if_stale && store.is_stale()? {
           store.clear()?;
       }
       Ok(store)
   }
}
```

Or, we can semantically change KvpPoolStore to be path oriented as a whole instead. 
 
```
use std::path::{Path, PathBuf};
#[derive(Clone, Copy, Debug, PartialEq, Eq)]
pub enum KvpPool {
   External,
   Guest,
   Auto,
   AutoExternal,
   AutoInternal,
}
impl KvpPool {
   pub fn as_path(self) -> &'static Path {
       match self {
           Self::External => Path::new("/var/lib/hyperv/.kvp_pool_0"),
           Self::Guest => Path::new("/var/lib/hyperv/.kvp_pool_1"),
           Self::Auto => Path::new("/var/lib/hyperv/.kvp_pool_2"),
           Self::AutoExternal => Path::new("/var/lib/hyperv/.kvp_pool_3"),
           Self::AutoInternal => Path::new("/var/lib/hyperv/.kvp_pool_4"),
       }
   }
}
impl From<KvpPool> for PathBuf {
   fn from(pool: KvpPool) -> Self {
       pool.as_path().to_path_buf()
   }
}
```
 
I was leaning towards the first approach, but am open to either depending on what you envisioned here!

libazureinit-kvp/src/kvp_pool.rs
```diff
@@ -0,0 +1,1662 @@
...
+impl KvpPoolStore {
+    /// Create a new KVP pool store.
+    ///
+    /// - `pool`: which Hyper-V pool to open (see [`KvpPool`]).
+    /// - `path`: override pool file path; defaults to
+    ///   [`KvpPool::default_path`].
+    /// - `mode`: size-limit policy (see [`PoolMode`]).
+    /// - `truncate_on_stale`: if `true`, clears stale data from a
+    ///   previous boot.
+    pub fn new(
```
cjp256: let's leave path out of the default new....
```rust
impl KvpPool {
    pub fn filename(self) -> &'static str {
        match self {
            Self::External => ".kvp_pool_1",
            Self::Guest => ".kvp_pool_2",
        }
    }

    pub fn default_dir() -> &'static Path {
        Path::new("/var/lib/hyperv")
    }

    pub fn default_path(self) -> PathBuf {
        Self::default_dir().join(self.filename())
    }
}

#[derive(Debug)]
pub struct KvpStore {
    pool: KvpPool,
    path: PathBuf,
    mode: PoolMode,
}

impl KvpStore {
    /// Open using the default Hyper-V directory.
    pub fn new(pool: KvpPool, mode: PoolMode) -> Result<Self, KvpError> {
        Self::new_in(pool, KvpPool::default_dir(), mode)
    }

    /// Open using a custom directory (filename derived from pool).
    pub fn new_in(
        pool: KvpPool,
        dir: impl AsRef<Path>,
        mode: PoolMode,
    ) -> Result<Self, KvpError> {
        let path = dir.as_ref().join(pool.filename());
        Ok(Self { pool, path, mode })
    }

    pub fn path(&self) -> &Path {
        &self.path
    }
...
```

libazureinit-kvp/src/lib.rs
```diff
@@ -0,0 +1,119 @@
...
+impl From<io::Error> for KvpError {
+    fn from(err: io::Error) -> Self {
+        Self::Io(err)
+    }
+}
+
+pub use kvp_pool::KvpPoolStore;
+
+/// Key-value store with Hyper-V KVP semantics.
+pub trait KvpStore: Send + Sync {
```
cjp256: let's remove this in favor of the definition in kvp_pool.rs

and let's do a little restructuring, e.g.:

```
src/
├── lib.rs
├── error.rs
└── store.rs
```

and lib.rs is something like:
```rust

mod error;
mod store;

pub use error::KvpError;
pub use store::{KvpPool, KvpStore, PoolMode};
```

peytonr18: done!

libazureinit-kvp/src/kvp_pool.rs
peytonr18: done!

libazureinit-kvp/src/store.rs
```diff
@@ -0,0 +1,2149 @@
...
+            io::Error::last_os_error()
+        )));
+    }
+    Ok(())
+}
+
+/// Release a lock on the entire file.
+///
+/// Unlock errors are non-actionable (the fd is about to close, which
+/// releases the lock unconditionally), so the return value is discarded.
```
cjp256: it seems like this should raise the error, but then have the caller ignore it if it's not needed.

peytonr18: Done!

libazureinit-kvp/src/store.rs
```diff
@@ -0,0 +1,2138 @@
...
+
+        let mut records = Vec::with_capacity(iter.record_count());
+        for record in iter {
+            records.push(record?);
+        }
+        Ok(records)
+    }
+
+    /// Dump all entries (deduplicated) to a JSON file at the given path.
+    pub fn dump_json(&self, path: &Path) -> Result<(), KvpError> {
```
cjp256: if it's going to be named dump_ json() it probably should match the semantics of dump().

but if we don't have a strong use case for it, i'm in favor of removal instead

peytonr18: +1 for removal, will do that!

libazureinit-kvp/src/error.rs
```diff
@@ -0,0 +1,59 @@
...
+#[derive(Debug)]
+pub enum KvpError {
+    /// The key was empty.
+    EmptyKey,
+    /// The key exceeds the store's maximum key size.
+    KeyTooLarge { max: usize, actual: usize },
+    /// The value exceeds the store's maximum value size.
+    ValueTooLarge { max: usize, actual: usize },
+    /// The store already has the maximum allowed number of unique keys.
+    MaxUniqueKeysExceeded { max: usize },
```
cjp256, line 20: alpha sort

libazureinit-kvp/src/store.rs
```diff
@@ -0,0 +1,2138 @@
...
+}
+
+/// Policy mode controlling key/value size limits for writes.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum PoolMode {
+    /// Conservative limits for Linux kernel compatibility
+    /// (key <= 254 bytes, value <= 1022 bytes).
+    Safe,
+    /// Full Hyper-V wire-format limits
+    /// (key <= 512 bytes, value <= 2048 bytes).
```
cjp256: this is ambiguous.  let's be clear about the utf-X encodings expected

libazureinit-kvp/src/store.rs
```diff
@@ -0,0 +1,2138 @@
...
+    pub fn delete(&self, key: &str) -> Result<bool, KvpError> {
+        let file = match self.open_for_read_write() {
+            Ok(f) => f,
+            Err(ref e) if e.kind() == ErrorKind::NotFound => {
+                return Ok(false);
+            }
+            Err(e) => return Err(e.into()),
+        };
+
+        let mut iter = KvpPoolIter::new(file, true)?;
```
cjp256: should we use iter_mut() here?

libazureinit-kvp/src/store.rs
```diff
@@ -0,0 +1,2138 @@
...
+        let mut found = false;
+        let mut unique_keys = HashSet::new();
+
+        while let Some(record) = iter.next() {
+            let (k, _) = record?;
+            let is_target = k == key;
+            unique_keys.insert(k);
+            if is_target {
+                if !found {
+                    iter.overwrite_current_value(value)?;
```
cjp256: i htink we should bail from loop here?

cjp256: nevermind, i see - this is clearing any other conflicting keys later.

maybe make note of that.

libazureinit-kvp/src/store.rs
```diff
@@ -0,0 +1,2138 @@
...
+            Err(ref e) if e.kind() == ErrorKind::NotFound => return Ok(false),
+            Err(e) => return Err(e.into()),
+        };
+        let boot = boot_time()?;
+        Ok(metadata.mtime() < boot)
+    }
+
+    /// Test-only variant of [`is_stale`](Self::is_stale) that accepts
+    /// an explicit boot time for deterministic assertions.
+    #[cfg(test)]
```
cjp256, line 315: what's the need for this variant?  i don't think duplicating is_stale()'s code is adding any value.

if it's truly needed....

is_stale() calls is_stale_at(boot_time) would avoid code duplication

libazureinit-kvp/src/store.rs
cjp256: 256 utf-8 vs 512 utf-16 bytes.

on-disk pool is utf-8.

libazureinit-kvp/src/store.rs
```diff
@@ -0,0 +1,2138 @@
...
+    }
+
+    /// Flush any buffered writes to the underlying file.
+    fn flush(&mut self) -> io::Result<()> {
+        self.file.flush()
+    }
+
+    fn new(mut file: File, lock_exclusive: bool) -> Result<Self, KvpError> {
+        let lock_result = if lock_exclusive {
+            fcntl_lock_exclusive(&file)
```
cjp256: we want to take both fcntl and flock locks:

this lock works for hv_kvp_daemon, but cloud-init uses flock(LOCK_EX)

https://github.com/canonical/cloud-init/blob/c7ae6a4cc0dad5a411945e0644dab6926d41a997/cloudinit/reporting/handlers.py#L286

cjp256: maybe add a lock_for_reading() and lock_for_writing()  

and the semantic here, if needed, perhaps should be read_only: bool, because we want to guarantee the locks for any write operations

libazureinit-kvp/src/store.rs
```diff
@@ -0,0 +1,2138 @@
...
+                "no record has been read yet",
+            ));
+        }
+
+        let delete_index = self.current_index - 1;
+        let last_index = self.record_count - 1;
+
+        if delete_index != last_index {
+            self.file.seek(io::SeekFrom::Start(
+                last_index as u64 * RECORD_SIZE as u64,
```
cjp256, line 914: why last index?  doesn't this reverse ordering if copying the last record and moving it to current?

libazureinit-kvp/src/store.rs
```diff
@@ -0,0 +1,2138 @@
...
+        self.record_count -= 1;
+        Ok(())
+    }
+}
+
+impl Drop for KvpPoolIter {
+    fn drop(&mut self) {
+        // Drop cannot return a Result; the fd closes immediately
+        // after, which releases the lock unconditionally.
+        let _ = fcntl_unlock(&self.file);
```
cjp256: I think we should be explicit in the lock/unlock and avoid drop traits to do it for us so it's clear.

libazureinit-kvp/src/store.rs
```diff
@@ -0,0 +1,2138 @@
...
+        match self.file.read_exact(&mut buf) {
+            Ok(()) => {
+                self.current_index += 1;
+                Some(decode_record(&buf))
+            }
+            Err(e) => Some(Err(e)),
+        }
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
```
cjp256, line 960:
```suggestion
    fn remaining(&self) -> (usize, Option<usize>) {
``` ?

Comment (cjp256) [reply libazureinit-kvp/src/store.rs:960]: nvm, rust thing

File: libazureinit-kvp/src/store.rs (hunk 34 above)
Comment (peytonr18) [reply libazureinit-kvp/src/store.rs]: I think this is correct as is because `delete()` opens the file via `open_for_read_write()` where there isn't a create call and then the function exits with `Ok(false)` if the file is missing. using `iter_mut()` would open the file via `open_for_read_write_create()` and subsequently create an empty pool file as a side effect from trying to delete from it

File: libazureinit-kvp/src/error.rs (hunk 32 above)
Comment (peytonr18) [reply libazureinit-kvp/src/error.rs:20]: done!

File: libazureinit-kvp/src/store.rs (hunk 33 above)
Comment (peytonr18) [reply libazureinit-kvp/src/store.rs]: done!

File: libazureinit-kvp/src/store.rs (hunk 38 above)
Comment (peytonr18) [reply libazureinit-kvp/src/store.rs:914]: I added some comments to make the iteration logic clearer here, but let me know if you still have questions/concerns and I can rework the approach!

File: libazureinit-kvp/src/store.rs [hunk 41]
```diff
@@ -0,0 +1,2360 @@
...
+        if self.is_stale()? {
+            self.clear()?;
+        }
+        Ok(())
+    }
+
+    /// Remove all records matching the key. Returns `true` if at least
+    /// one record was present.
+    pub fn delete(&self, key: &str) -> Result<bool, KvpError> {
+        let file = match self.open_for_read_write() {
```
cjp256: why is this special cased and not handled by iter_mut()? we should be able to see the NotFound error [and ignore it] through that interface too?

libazureinit-kvp/src/store.rs
cjp256:
```rust


    /// Exposes a bug where `remove_current()` reorders duplicate
    /// records via its swap-with-last strategy, silently changing
    /// which duplicate `entries()` reports as last-write-wins.
    ///
    /// Initial on-disk order has `("dup", "C")` as the final record
    /// for key `"dup"`, so `entries()["dup"] == "C"`. After deleting
    /// the unrelated `"middle"` key, the swap moves `("dup", "C")`
    /// into the freed slot ahead of `("dup", "B")`, leaving `"B"` as
    /// the on-disk-last record for `"dup"`.
    #[test]
    fn test_delete_reorders_duplicates_breaking_last_write_wins() {
        let dir = TempDir::new().unwrap();
        let store = safe_store(dir.path());

        store
            .populate(pairs([
                ("dup", "A"),
                ("middle", "m"),
                ("dup", "B"),
                ("other", "x"),
                ("dup", "C"),
            ]))
            .unwrap();

        assert_eq!(
            store.entries().unwrap().get("dup"),
            Some(&"C".to_string()),
            "precondition: last-write-wins picks the final on-disk record",
        );

        assert!(store.delete("middle").unwrap());

        assert_eq!(
            store.entries().unwrap().get("dup"),
            Some(&"C".to_string()),
            "deleting an unrelated key must not change which duplicate \
             entries() reports as the winning value",
        );
    }

    /// Companion to
    /// [`test_delete_reorders_duplicates_breaking_last_write_wins`]
    /// for the other production caller of `remove_current()`: the
    /// duplicate-collapsing branch of [`KvpPoolStore::insert`].
    ///
    /// Initial on-disk order has `("dup", "C")` as the final record
    /// for key `"dup"`, so `entries()["dup"] == "C"`. Inserting
    /// `"ins"` overwrites the first `"ins"` record in place and then
    /// calls `remove_current()` on the second `"ins"`, which swaps
    /// the tail `("dup", "C")` forward of `("dup", "B")`, leaving
    /// `"B"` as the on-disk-last record for `"dup"`.
    #[test]
    fn test_insert_reorders_duplicates_breaking_last_write_wins() {
        let dir = TempDir::new().unwrap();
        let store = safe_store(dir.path());

        store
            .populate(pairs([
                ("ins", "i1"),
                ("ins", "i2"),
                ("dup", "B"),
                ("dup", "C"),
            ]))
            .unwrap();

        assert_eq!(
            store.entries().unwrap().get("dup"),
            Some(&"C".to_string()),
            "precondition: last-write-wins picks the final on-disk record",
        );

        store.insert("ins", "new").unwrap();

        assert_eq!(
            store.entries().unwrap().get("dup"),
            Some(&"C".to_string()),
            "insert collapsing its own duplicates must not change which \
             duplicate of an unrelated key entries() reports as winning",
        );
    }
```

```bash


failures:

---- store::tests::test_delete_reorders_duplicates_breaking_last_write_wins stdout ----

thread 'store::tests::test_delete_reorders_duplicates_breaking_last_write_wins' (1817184) panicked at libazureinit-kvp/src/store.rs:1341:9:
assertion `left == right` failed: deleting an unrelated key must not change which duplicate entries() reports as the winning value
  left: Some("B")
 right: Some("C")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- store::tests::test_insert_reorders_duplicates_breaking_last_write_wins stdout ----

thread 'store::tests::test_insert_reorders_duplicates_breaking_last_write_wins' (1817203) panicked at libazureinit-kvp/src/store.rs:1382:9:
assertion `left == right` failed: insert collapsing its own duplicates must not change which duplicate of an unrelated key entries() reports as winning
  left: Some("B")
 right: Some("C")


failures:
    store::tests::test_delete_reorders_duplicates_breaking_last_write_wins
    store::tests::test_insert_reorders_duplicates_breaking_last_write_wins

test result: FAILED. 100 passed; 2 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.72s

error: test failed, to rerun pass `-p libazureinit-kvp --lib`
```

cjp256: or for a dump-based view

```rust
    /// Exposes a bug where `remove_current()` reorders duplicate
    /// records via its swap-with-last strategy, silently changing
    /// which duplicate `entries()` reports as last-write-wins.
    ///
    /// Initial on-disk order has `("dup", "C")` as the final record
    /// for key `"dup"`, so `entries()["dup"] == "C"`. After deleting
    /// the unrelated `"middle"` key, the swap moves `("dup", "C")`
    /// into the freed slot ahead of `("dup", "B")`, leaving `"B"` as
    /// the on-disk-last record for `"dup"`.
    #[test]
    fn test_delete_reorders_duplicates_breaking_last_write_wins() {
        let dir = TempDir::new().unwrap();
        let store = safe_store(dir.path());

        store
            .populate(pairs([
                ("dup", "A"),
                ("middle", "m"),
                ("dup", "B"),
                ("other", "x"),
                ("dup", "C"),
            ]))
            .unwrap();

        assert_eq!(
            store.entries().unwrap().get("dup"),
            Some(&"C".to_string()),
            "precondition: last-write-wins picks the final on-disk record",
        );

        assert!(store.delete("middle").unwrap());

        assert_eq!(
            store.dump().unwrap(),
            pairs([
                ("dup", "A"),
                ("dup", "B"),
                ("other", "x"),
                ("dup", "C"),
            ]),
            "deleting an unrelated key must preserve the relative order \
             of every other record (so last-write-wins still picks C \
             for `dup`)",
        );
        assert_eq!(
            store.entries().unwrap().get("dup"),
            Some(&"C".to_string()),
            "deleting an unrelated key must not change which duplicate \
             entries() reports as the winning value",
        );
    }

    /// Companion to
    /// [`test_delete_reorders_duplicates_breaking_last_write_wins`]
    /// for the other production caller of `remove_current()`: the
    /// duplicate-collapsing branch of [`KvpPoolStore::insert`].
    ///
    /// Initial on-disk order has `("dup", "C")` as the final record
    /// for key `"dup"`, so `entries()["dup"] == "C"`. Inserting
    /// `"ins"` overwrites the first `"ins"` record in place and then
    /// calls `remove_current()` on the second `"ins"`, which swaps
    /// the tail `("dup", "C")` forward of `("dup", "B")`, leaving
    /// `"B"` as the on-disk-last record for `"dup"`.
    #[test]
    fn test_insert_reorders_duplicates_breaking_last_write_wins() {
        let dir = TempDir::new().unwrap();
        let store = safe_store(dir.path());

        store
            .populate(pairs([
                ("ins", "i1"),
                ("ins", "i2"),
                ("dup", "B"),
                ("dup", "C"),
            ]))
            .unwrap();

        assert_eq!(
            store.entries().unwrap().get("dup"),
            Some(&"C".to_string()),
            "precondition: last-write-wins picks the final on-disk record",
        );

        store.insert("ins", "new").unwrap();

        assert_eq!(
            store.dump().unwrap(),
            pairs([("ins", "new"), ("dup", "B"), ("dup", "C")]),
            "collapsing duplicate `ins` records must preserve the \
             relative order of unrelated `dup` records (so last-write-wins \
             still picks C for `dup`)",
        );
        assert_eq!(
            store.entries().unwrap().get("dup"),
            Some(&"C".to_string()),
            "insert collapsing its own duplicates must not change which \
             duplicate of an unrelated key entries() reports as winning",
        );
    }


```

```bash


failures:

---- store::tests::test_delete_reorders_duplicates_breaking_last_write_wins stdout ----

thread 'store::tests::test_delete_reorders_duplicates_breaking_last_write_wins' (1818095) panicked at libazureinit-kvp/src/store.rs:1341:9:
assertion `left == right` failed: deleting an unrelated key must preserve the relative order of every other record (so last-write-wins still picks C for `dup`)
  left: [("dup", "A"), ("dup", "C"), ("dup", "B"), ("other", "x")]
 right: [("dup", "A"), ("dup", "B"), ("other", "x"), ("dup", "C")]
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- store::tests::test_insert_reorders_duplicates_breaking_last_write_wins stdout ----

thread 'store::tests::test_insert_reorders_duplicates_breaking_last_write_wins' (1818139) panicked at libazureinit-kvp/src/store.rs:1394:9:
assertion `left == right` failed: collapsing duplicate `ins` records must preserve the relative order of unrelated `dup` records (so last-write-wins still picks C for `dup`)
  left: [("ins", "new"), ("dup", "C"), ("dup", "B")]
 right: [("ins", "new"), ("dup", "B"), ("dup", "C")]
```

cjp256: I think we just have to bite the bullet and redo these to shift everything up (and delete dupe keys in the process).  I think if we have to iterate through all entries anyways, the cost isn't really a problem.  Callers should avoid introducing duplicate keys anyways so we likely won't hit the worst-case.

libazureinit-kvp/src/store.rs
```diff
@@ -0,0 +1,2360 @@
...
+}
+
+/// Run `f` against `iter`, then unlock the iterator regardless of
+/// whether `f` succeeded. The first error (work or unlock) wins.
+fn run_with_iter<T, F>(mut iter: KvpPoolIter, f: F) -> Result<T, KvpError>
+where
+    F: FnOnce(&mut KvpPoolIter) -> Result<T, KvpError>,
+{
+    let work_result = f(&mut iter);
+    let unlock_result = iter.unlock().map_err(KvpError::from);
```
cjp256: OK, I think I've come around on the drop pattern here.

this way if scope of iter is too long, it can be simply managed with something like:

```rust
let result = {
    let mut iter = self.iter_mut()?;
    iter.append(key, value)?;
};
```

to ensure it's unlocked when the caller needs it to be.

I think this is more generally safe than trying to ensure all callers run_with_iter().  And file closure depends on the same mechanism, so this makes it consistent.

peytonr18: Adopted this in the latest commit!

libazureinit-kvp/src/store.rs
peytonr18: Good call out - this should be fixed in the most recent commit!

cjp256 (approved):

cadejacobson (approved):

cjp256:
Great job @peytonr18, thanks!

PR #294, opened 2026-04-20
feat(ci): per-file code coverage reporting

cadejacobson:
Adds a Python script to enable per-file code coverage reporting. `cargo llvm-cov` does not offer great ways to show the coverage by file, but the cobertura.xml file that is produced for the `irongut` coverage report can be easily parsed to view this information. 

This brings us to a similar spot as PRs in ADO. Per-file and per-line reporting is now right in the same location.

## How to use
Completed automatically via the E2E pipeline. Click the `Details` on the `Code Coverage Report` pipeline, then `Summary` in the top left.

### DIFF
--- a/.github/workflows/coverage-report.yml
+++ b/.github/workflows/coverage-report.yml
@@ -50,6 +50,46 @@ jobs:
         if: always()
         run: cat code-coverage-results.md >> "$GITHUB_STEP_SUMMARY"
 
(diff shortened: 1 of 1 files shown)


peytonr18 (approved):
Thank you for doing this! This is much more helpful than having to run the html check manually.

cjp256 (approved):
this is great, thanks!

PR #300, opened 2026-05-05
fix(ci): add a timeout to stop failed containers

cadejacobson:
Adds a timeout to stop containers that have failed to build correctly, but have not thrown the expected errors.

## Testing done
Forced a timeout in commit 5ac1145 to show the correct behavior taking place. Reverted this commit after proof.

### DIFF
--- a/testinit/start-all.sh
+++ b/testinit/start-all.sh
@@ -22,7 +22,7 @@ while true; do
   # Check journal logs for either a success or a failure to ensure the completion of the service
   if docker exec azureinit-provisioning-agent journalctl -u azure-init.service --no-pager | grep -q "Finished azure-init.service"; then
     echo "azure-init.service has finished"
-    # break
+    break
   fi
   if  docker exec azureinit-provisioning-agent journalctl -u azure-init.service --no-pager | grep -i 'azure-init' | grep -Eiq 'failed|failure'; then
     echo "azure-init.service has failed."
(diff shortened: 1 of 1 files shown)


cadejacobson:
Resolves #301.

cjp256 (approved):

peytonr18 (approved):

testinit/start-all.sh
```diff
@@ -25,8 +28,13 @@ while true; do
     echo "azure-init.service has failed."
     break
   fi
-  echo "Waiting for azure-init.service to finish..."
+  if [ "$ELAPSED" -ge "$TIMEOUT" ]; then
+    echo "ERROR: Timed out after 20 minutes waiting for azure-init.service to complete."
+    exit 1
+  fi
+  echo "Waiting for azure-init.service to finish... (${ELAPSED}s/${TIMEOUT}s)"
```
bpryan99, line 35: One way to make this a little more reliable would be to replace ELAPSED and instead calculate a datetime in the future that's the start time plus the timeout. Then you just compare timeout to current time via subshell. It's a bit more idiomatic versus checking a counter, though it won't make a functional difference here so feel free to disregard!

bpryan99 (approved):

testinit/start-all.sh
cadejacobson: This is a very cool solution! I will go this route for future time out implementations now that I know. Since we previously had the setup of using the `sleep` command, I will leave this solution for now. Thank you for this information!!

PR #308, opened 2026-06-09
feat(kvp): add append_multiple and delete_multiple to KvpPoolStore

peytonr18:
## Summary

Add bulk KVP store operations for appending and deleting multiple records under a single lock.

- Add `append_multiple` to write a batch of records contiguously while preserving input order
- Add `delete_multiple` to remove all records matching a set of keys in one file pass
- Reuse a shared locked write helper between `populate` and `append_multiple`
- Keep behavior aligned with existing single-record APIs:
  - `append_multiple` follows `append` and does not enforce `MAX_UNIQUE_KEYS`
  - `delete_multiple` follows `delete` and validates keys without applying mode-specific size caps
  - empty bulk inputs are no-ops
  - missing files return `Ok(0)` for bulk delete

## Details

The bulk append path validates the full input before opening the pool file, then holds one exclusive lock for the entire write. This prevents cooperating writers from interleaving records into the middle of a batch, which is useful for diagnostics or chunked records that should remain adjacent on disk.

The bulk delete path collects validated keys into a set, scans the pool once, and removes every matching on-disk record. Duplicate stored keys are counted individually in the returned removal count. Like existing delete behavior, removals use swap-with-tail truncation, so remaining record order is not preserved.

## Test coverage

28 new tests; all 201 unit tests pass; production-code region coverage on the diff is 100% (verified with `cargo llvm-cov -p libazureinit-kvp`).

### DIFF
--- a/libazureinit-kvp/src/store.rs
+++ b/libazureinit-kvp/src/store.rs
@@ -143,6 +143,38 @@ impl KvpPoolStore {
         Ok(())
     }
 
+    /// Append records under one exclusive lock, preserving order and
+    /// preventing cooperating writers from interleaving with the batch.
+    ///
+    /// Like [`append`](Self::append), this does not enforce
+    /// [`MAX_UNIQUE_KEYS`]. Validation happens before the file is
+    /// opened; empty input is a no-op. If an I/O error occurs after
+    /// writing starts, the file may contain a partial batch.
+    pub fn append_multiple<I, K, V>(&self, records: I) -> Result<(), KvpError>
+    where
+        I: IntoIterator<Item = (K, V)>,
+        K: Into<String>,
+        V: Into<String>,
+    {
+        let records: Vec<(String, String)> = records
+            .into_iter()
+            .map(|(k, v)| (k.into(), v.into()))
+            .collect();
+
+        if records.is_empty() {
+            return Ok(());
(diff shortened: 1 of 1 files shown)


cjp256 (approved):
LGTM! Should we remove populate() in favor of append_multiple()?  I'm not sure the value add of both.

peytonr18:
> LGTM! Should we remove populate() in favor of append_multiple()? I'm not sure the value add of both.

I'm very open to talking about this further, but at this point, I think we should keep both. 

They both accept multiple records, but they represent different write semantics: `append_multiple()` is an atomic batch append that preserves existing pool contents, while `populate()` is an atomic full replacement of the pool.

The duplicate-key behavior is actually the same for both: both preserve duplicate records as provided. The main differences are that `populate()` truncates/replaces existing contents, treats empty input as clearing the pool, and enforces the unique-key cap because it has the full replacement set. On the other hand, `append_multiple()` extends existing contents, treats empty input as a no-op, and intentionally follows `append()` by not enforcing the unique-key cap.

I very much agree the distinction was not obvious from the comments, so I updated the docs to spell out replace-vs-append, empty input behavior, duplicate preservation, and key-cap behavior.

cadejacobson (approved):

PR #309, opened 2026-06-10
feat(kvp): add libazureinit-kvp CLI

peytonr18:
## Summary

Adds a `libazureinit-kvp` command-line interface around the existing KVP store APIs, supporting development, debugging, and operational workflows against Azure/Hyper-V KVP pool files. The CLI can inspect pool metadata, read/write/delete records, bulk-load records, clear stale data, and check staleness.

Adds `clap` (derive) as a direct dependency on `libazureinit-kvp`.

## Usage

The CLI operates on one pool at a time. Defaults: pool `guest`, safe mode, default Hyper-V KVP directory.

```sh
libazureinit-kvp info
libazureinit-kvp --pool guest --dir /var/lib/hyperv info
libazureinit-kvp --pool 1 --unsafe dump
```

Global options (valid with any command):

- `--pool <POOL>` — `external`/`0`, `guest`/`1`, `auto`/`2`, `auto-external`/`3`, `auto-internal`/`4`
- `--dir <PATH>` — pool directory
- `--unsafe` — use the full wire-format key/value limits instead of the safe profile

## Commands

| Command | Description |
|---|---|
| `info` | Print store metadata: pool, path, mode, record count, empty/stale state, and key/value size limits. |
| `dump` | Print every record in insertion order as `KEY=VALUE`; duplicates preserved. |
| `entries` | Print one `KEY=VALUE` per key (last value wins), sorted by key. |
| `read <KEY>` | Print the last value for a key. Exit `1` if missing. |
| `write [--append] <KEY> <VALUE>` | Write a record. Replaces existing records for `KEY` unless `--append` is set. |
| `load [--file PATH]` | Replace pool contents from `KEY=VALUE` lines (file or stdin). |
| `delete <KEY>` | Delete every record for a key; prints `true`/`false`. |
| `clear [--yes] [--if-stale]` | Clear the pool. Requires `--yes` unless `--if-stale` is set. |
| `is-stale` | Print whether the pool is stale. Exit `0` if stale, `1` if not. |

Examples:

```sh
libazureinit-kvp write hostname vm-01
libazureinit-kvp write --append history booted
printf 'a=1\nb=2\n' | libazureinit-kvp load
libazureinit-kvp clear --if-stale
```

## Exit codes

- `0` — success
- `1` — not found / false status (missing key, `is-stale=false`)
- `2` — usage or validation error
- `3` — I/O error

KVP store validation errors surface as CLI errors on stderr. `read` of a missing key writes nothing to stderr.

## Test Coverage
```
Filename                      Regions    Missed Regions     Cover   Functions  Missed Functions  Executed       Lines      Missed Lines     Cover    Branches   Missed Branches     Cover
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
cli.rs                            917                37    95.97%          52                 0   100.00%         510                 0   100.00%           0                 0         -
error.rs                           39                 0   100.00%           3                 0   100.00%          21                 0   100.00%           0                 0         -
main.rs                             3                 0   100.00%           1                 0   100.00%           3                 0   100.00%           0                 0         -
store.rs                         4661                51    98.91%         264                 0   100.00%        2142                 0   100.00%           0                 0         -
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
TOTAL                            5620                88    98.43%         320                 0   100.00%        2676                 0   100.00%           0                 0         -
```

### DIFF
--- a/libazureinit-kvp/src/cli.rs
+++ b/libazureinit-kvp/src/cli.rs
@@ -7,6 +7,7 @@ use std::path::PathBuf;
 use std::process::ExitCode;
 
 use clap::{Parser, Subcommand, ValueEnum};
+use serde_json::json;
 
 use crate::{KvpError, KvpPool, KvpPoolStore, PoolMode};
 
@@ -51,8 +52,29 @@ struct Cli {
     #[arg(long = "unsafe", global = true)]
     unsafe_mode: bool,
 
+    /// Emit machine-readable JSON for commands that produce output.
+    #[arg(long, global = true)]
+    json: bool,
+
     #[command(subcommand)]
     command: Command,
+}
+
+/// Output format for commands that print data.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum OutputMode {
+    Text,
+    Json,
+}
+
+impl OutputMode {
+    fn from_flag(json: bool) -> Self {
+        if json {
+            Self::Json
+        } else {
+            Self::Text
+        }
+    }
 }
 
 #[derive(Subcommand, Debug)]
@@ -79,16 +101,24 @@ enum Command {
         #[arg(long)]
         file: Option<PathBuf>,
     },
+    /// Append KEY=VALUE lines read from --file or stdin.
+    AppendMultiple {
+        /// Read records from PATH instead of stdin.
+        #[arg(long)]
+        file: Option<PathBuf>,
+    },
     /// Delete every record with KEY (prints `true` if any were removed).
     Delete { key: String },
-    /// Clear the pool. Requires --yes unless --if-stale is set.
+    /// Delete every record matching any KEY (prints removed record count).
+    DeleteMultiple {
+        #[arg(required = true)]
+        keys: Vec<String>,
+    },
+    /// Clear the pool. Pass --if-stale to clear only when stale.
     Clear {
         /// Only clear if the store is currently stale.
         #[arg(long = "if-stale")]
         if_stale: bool,
-        /// Confirm unconditional clear.
-        #[arg(long)]
-        yes: bool,
     },
     /// Print whether the pool is stale (exit 0 if stale, 1 otherwise).
     IsStale,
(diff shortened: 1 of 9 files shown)


libazureinit-kvp/src/cli.rs
```diff
@@ -0,0 +1,766 @@
...
+        PoolMode::Unsafe
+    } else {
+        PoolMode::Safe
+    };
+
+    let store = match cli.dir {
+        Some(dir) => KvpPoolStore::new_in(pool, dir, mode),
+        None => KvpPoolStore::new(pool, mode),
+    }
+    .expect("KvpPoolStore construction is currently infallible");
```
cjp256, line 167: lol

cjp256 (changes requested):
load actually sounds better than populate.  maybe we should rename populate too.

i think --yes is overkill for clear.  None of this data is sacred and generally overwritten on reboot.

--json will be useful for outputs I think.  entries key=value gets awkward if there are newlines unless we encode it, etc.

libazureinit-kvp/src/cli.rs
peytonr18: I originally had just a simple `?` since it can't fail, but that breaks the `cargo-llvm coverage` since it introduces a new path that we can't actually hit. This was the alternative solution lol

peytonr18:
> load actually sounds better than populate. maybe we should rename populate too.
> 
> i think --yes is overkill for clear. None of this data is sacred and generally overwritten on reboot.
> 
> --json will be useful for outputs I think. entries key=value gets awkward if there are newlines unless we encode it, etc.

Addressed all of these in the latest commit! I was trying to avoid adding too many dependencies to keep the size minimal but I agree that --json would be super helpful to have. 

I also like `load()` instead of `populate()` so I went ahead and renamed that function.

cadejacobson (approved):

cjp256 (approved):
LGTM, thanks!

PR #310, opened 2026-06-17
feat(kvp): add ProvisioningReport abstraction over KvpPoolStore

peytonr18:
## Summary
Introduces a `ProvisioningReport` type in the `libazureinit-kvp` crate that models a provisioning health report as structured data and persists it to a KVP pool as ordered key/value entries — replacing ad-hoc string building at the call site.

## What's included
- `ProvisioningReport` — report with `success(agent, vm_id)` and `error(agent, vm_id, reason)` constructors plus `with_pps_type`, `with_documentation_url`, and `with_extra` builders. Fields are private; the success⇔no-reason / error⇔reason invariant is enforced through the constructors.
- `ToKvp` trait — converts a report into ordered `Vec<(String, String)>` entries. Acts as the shared seam so a future diagnostics report can reuse `write_report` without changes here.
- `write_report` — persists entries via `KvpPoolStore::insert` (upsert / last-write-wins).

## Testing
- New unit tests cover all of the new functionality.

### DIFF
--- a/libazureinit-kvp/src/cli.rs
+++ b/libazureinit-kvp/src/cli.rs
@@ -138,10 +138,11 @@ enum Command {
         /// Reporting agent identifier (e.g. libazureinit-kvp/0.1.0).
         #[arg(long, default_value = DEFAULT_AGENT)]
         agent: String,
-        /// Additional key=value supporting data, comma-separated or the
-        /// flag repeated (e.g. --supporting-data k1=v1,k2=v2).
-        #[arg(long, value_delimiter = ',', value_parser = parse_key_value_pair)]
-        supporting_data: Vec<(String, String)>,
+        /// Additional key=value supporting data, comma-separated
+        /// (e.g. --supporting-data k1=v1,k2=v2). The flag may also be
+        /// repeated.
+        #[arg(long, value_parser = parse_supporting_data)]
+        supporting_data: Vec<SupportingData>,
     },
     /// Write a failure provisioning health report, overriding any existing
     /// `PROVISIONING_REPORT` record.
@@ -158,10 +159,22 @@ enum Command {
         /// Optional documentation URL describing the failure.
         #[arg(long)]
         documentation_url: Option<String>,
-        /// Additional key=value supporting data, comma-separated or the
-        /// flag repeated (e.g. --supporting-data k1=v1,k2=v2).
-        #[arg(long, value_delimiter = ',', value_parser = parse_key_value_pair)]
-        supporting_data: Vec<(String, String)>,
+        /// Additional key=value supporting data, comma-separated
+        /// (e.g. --supporting-data k1=v1,k2=v2). The flag may also be
+        /// repeated.
+        ///
+        /// To keep a literal comma in a value, wrap that value in matching
+        /// single or double quotes. The shell strips quotes first, so wrap
+        /// the whole argument to let the inner quotes through:
+        ///
+        /// --supporting-data "k1='a,b',k2=v2"  ->  k1=a,b  k2=v2
+        ///
+        /// --supporting-data 'k1="a,b",k2=v2'  ->  k1=a,b  k2=v2
+        ///
+        /// Quotes are only special at the start of a value and must be
+        /// balanced.
+        #[arg(long, value_parser = parse_supporting_data)]
+        supporting_data: Vec<SupportingData>,
     },
 }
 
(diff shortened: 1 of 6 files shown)


libazureinit-kvp/src/report.rs
```diff
@@ -0,0 +1,370 @@
...
+    /// Provisioning outcome (`result` field).
+    result: ReportResult,
+    /// Reporting agent identifier (`agent` field).
+    agent: String,
+    /// Virtual machine identifier (`vm_id` field).
+    vm_id: String,
+    /// Report timestamp (`timestamp` field), set to the current time
+    /// (RFC 3339) when the report is constructed.
+    timestamp: String,
+    /// Pre-provisioning type (`pps_type` field). Defaults to `None`.
```
cjp256: let's skip the default and just require it. caller can default to "None".  but also, let's maintain the types here, e.g. ReportPpsType

cadejacobson (approved):
Looks awesome! Great work

libazureinit-kvp/src/cli.rs
```diff
@@ -122,6 +125,35 @@ enum Command {
...
+        /// Optional human-readable message attached as an extra field.
+        #[arg(long)]
+        message: Option<String>,
+    },
+    /// Write a failure provisioning health report, overriding any existing
+    /// `PROVISIONING_REPORT` record.
+    ReportFailure {
+        /// Virtual machine identifier.
+        #[arg(long)]
+        vm_id: String,
```
cjp256: for the cli,

1. let's make this optional and auto populate it to the current vm's id.
2. support some form of --supporting-data key=value,key2=value2...
3. default agent to the version of this cli e.g. libazureinit-kvp/0.1

libazureinit-kvp/src/cli.rs
```diff
@@ -122,6 +125,35 @@ enum Command {
...
+    ReportSuccess {
+        /// Virtual machine identifier.
+        #[arg(long)]
+        vm_id: String,
+        /// Reporting agent identifier (e.g. Azure-Init/1.2.3).
+        #[arg(long)]
+        agent: String,
+        /// Optional human-readable message attached as an extra field.
+        #[arg(long)]
+        message: Option<String>,
```
cjp256: i don't know if we have any use for message.

peytonr18: I'd support getting rid of it! It's just a part of the cloudinit standard format, but I've never personally seen it used, so I think we have the flexibility to do whatever we think is best here.

libazureinit-kvp/src/cli.rs
peytonr18: Done in latest commit!

libazureinit-kvp/src/report.rs
cjp256: thanks!

libazureinit-kvp/src/cli.rs
cjp256: thanks!

libazureinit-kvp/src/cli.rs
```diff
@@ -130,29 +134,34 @@ enum Command {
...
         agent: String,
         /// Failure reason.
         #[arg(long)]
         reason: String,
         /// Optional documentation URL describing the failure.
         #[arg(long)]
         documentation_url: Option<String>,
+        /// Additional key=value supporting data, comma-separated or the
+        /// flag repeated (e.g. --supporting-data k1=v1,k2=v2).
+        #[arg(long, value_delimiter = ',', value_parser = parse_key_value_pair)]
```
cjp256: if the data contains a comma, will this break?  can v1 be quoted? e.g.    --suportting-data k1='foo,bar',k2=foo2

cjp256: if so maybe just allow repeating --supporting-data k1=v1 --suporting-data k2=v2...

cjp256: but i do like the comma syntax if it works or can be made to

peytonr18: Great catch - it would've broken! This is now fixed. 

 `--supporting-data` no longer uses a blanket comma delimiter, so a value can contain literal commas by wrapping it in matching single or double quotes (e.g. `--supporting-data "k1='foo,bar',k2=foo2"` yields `k1=foo,bar` and `k2=foo2`). The outer quotes are needed so the shell passes the inner quotes through.
 
It also supports repeating flags, like `--supporting-data k1=v1 --suporting-data k2=v2...`
 
I've also added rstest cases covering the quoting and error edge cases.

cjp256: thanks!

cjp256 (approved):

PR #313, opened 2026-07-09
feat(kvp): add DIAG V1 diagnostics and CLI tooling

peytonr18:
## Summary

This PR adds a `DiagnosticsKvp` layer that centralizes azure-init's KVP telemetry behavior, including event key formatting, message chunking, and message reassembly. This provides a single tested implementation that can be reused by the follow-up wiring PR and simplifies the existing logic in `kvp.rs` and `logging.rs`.

It also extends the existing `dump` command with a `--parse-diagnostics` mode for decoding and viewing KVP telemetry as readable events, making provisioning diagnostics easier to inspect and troubleshoot.

Specifically, this PR adds:

- `DiagnosticsKvp`, a typed view over `KvpPoolStore` that implements telemetry-specific behavior while keeping the underlying store format-agnostic:
  - Event keys formatted and parsed as `<prefix>|<vm_id>|<level>|<name>|<event_id>`.
  - Message chunking at UTF-8 boundaries for values larger than a single KVP record, written via `append_multiple` under a single lock. Each chunk gets a unique `|<subevent_index>`-suffixed key so the Hyper-V host (which keeps one record per key) retains every fragment, and the chunks are reassembled when read.
  - Classification of records as diagnostic events, raw records such as `PROVISIONING_REPORT`, or malformed event keys.

- Diagnostics folded into the existing raw KVP commands via flags rather than a separate command tree:
  - `dump --parse-diagnostics` reassembles chunked events and decodes each record. Raw output remains the default when the flag is absent.
  - `--include-raw` also prints raw (non-event) records; `--level`, `--name`, and `-n/--tail` narrow the output to a filtered, events-only view.
  - The filter flags (`--level`, `--name`, `-n/--tail`) require `--parse-diagnostics`, and `--include-raw` is mutually exclusive with them — it applies only to the unfiltered view, where raw records can appear.
  - `clear --diagnostics` removes every diagnostic key (events and malformed event keys) while leaving raw records such as `PROVISIONING_REPORT` intact.

## Example

A long event is stored as multiple records, each under a unique `|<subevent_index>` key. The raw view shows the fragments; `--parse-diagnostics` decodes and reassembles them into a single event:

```console
$ libazureinit-kvp dump
azure-init-1.0.0|3f25…|INFO|user:create_user|8f3e…|0=Time: … | Event: creating
azure-init-1.0.0|3f25…|INFO|user:create_user|8f3e…|1= azureuser (uid 1000) …
PROVISIONING_REPORT=result=success|…

$ libazureinit-kvp dump --parse-diagnostics --name user:create_user
event level=INFO name=user:create_user event_id=8f3e… message=Time: … | Event: creating azureuser (uid 1000) …

$ libazureinit-kvp dump --parse-diagnostics --include-raw
event level=INFO name=user:create_user event_id=8f3e… chunks=2 message=Time: … | Event: creating azureuser (uid 1000) …
raw key=PROVISIONING_REPORT value=result=success|…

$ libazureinit-kvp dump --parse-diagnostics --include-raw --json
[{"chunks":2,"event_id":"8f3e…","kind":"event","level":"INFO","message":"…","name":"user:create_user"},{"key":"PROVISIONING_REPORT","kind":"raw","value":"result=success|…"}]
```


cadejacobson (approved):

libazureinit-kvp/src/cli.rs
```diff
@@ -176,6 +176,50 @@ enum Command {
...
+    Clear {
+        /// VM identifier whose events to remove.
+        #[arg(long)]
+        vm_id: String,
+        /// Event key prefix whose events to remove.
+        #[arg(long)]
+        event_prefix: String,
+        /// Confirm the removal (required).
+        #[arg(long)]
+        yes: bool,
```
cjp256: `yes` is overkill i think.  It's volatile data and likely past the window of collection if someone is using it. Would suggest removing it.

libazureinit-kvp/src/cli.rs
```diff
@@ -176,6 +176,50 @@ enum Command {
...
+    Tail {
+        /// Number of trailing events to print.
+        #[arg(short = 'n', long = "count", default_value_t = 20)]
+        count: usize,
+    },
+    /// Remove this layer's events, leaving raw records intact.
+    Clear {
+        /// VM identifier whose events to remove.
+        #[arg(long)]
+        vm_id: String,
```
cjp256: this seems more like "delete" than "clear" to me.

I would expect clear to nuke all diagnostics-based keys.

libazureinit-kvp/src/cli.rs
```diff
@@ -425,6 +470,188 @@ fn is_stale<W: Write>(
...
+}
+
+/// Render a [`DiagnosticEvent`] as a JSON object (without chunk count).
+fn diag_event_json(event: &DiagnosticEvent) -> serde_json::Value {
+    json!({
+        "kind": "event",
+        "level": event.level.to_string(),
+        "name": event.name,
+        "event_id": event.event_id,
+        "message": event.message,
```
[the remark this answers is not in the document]
peytonr18: This isn't true - the binary compiles completely fine.

libazureinit-kvp/src/cli.rs
peytonr18: Removed this

libazureinit-kvp/src/cli.rs
peytonr18: I've adjusted this so clear gets rid of all diagnostic keys!

libazureinit-kvp/src/diagnostics.rs
```diff
@@ -0,0 +1,631 @@
...
+        while dumped.peek().is_some_and(|(next, _)| *next == key) {
+            let (_, next_value) = dumped.next().expect("peeked value exists");
+            message.push_str(&next_value);
+            chunks += 1;
+        }
+        records.push(classify_record(key, message, chunks));
+    }
+
+    records
+}
```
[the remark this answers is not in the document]
peytonr18: After the bug was fixed, keys can't be identical and reassembly uses the subevent index to chunk split KVP entries back together.

libazureinit-kvp/src/diagnostics.rs
```diff
@@ -0,0 +1,1091 @@
...
+            if matches!(
+                classify_key(base),
+                KeyClass::Event { .. } | KeyClass::CloudInit { .. }
+            ) {
+                return (base, Some(index));
+            }
+        }
+    }
+    (key, None)
+}
```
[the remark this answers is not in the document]
peytonr18: Good catch - addressed this!

libazureinit-kvp/src/diagnostics.rs
```diff
@@ -0,0 +1,1091 @@
...
+                | DiagnosticRecord::Raw { .. }
+                | DiagnosticRecord::Malformed { .. } => None,
+            })
+            .collect())
+    }
+
+    /// Remove every diagnostic key under a single lock: any key that
+    /// parses as an event key (including every `|<subevent_index>` chunk
+    /// of a multi-record event) or a malformed event key. Raw records
+    /// such as `PROVISIONING_REPORT` are left intact.
```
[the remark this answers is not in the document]
peytonr18: Made this change int the latest commit!

libazureinit-kvp/src/diagnostics.rs
```diff
@@ -0,0 +1,1079 @@
...
+        .map(|(key, value)| {
+            let (base, index) = split_subevent_index(&key);
+            (base.to_string(), index, value)
+        })
+        .peekable();
+
+    let mut records = Vec::new();
+    while let Some((base, index, value)) = parsed.next() {
+        let mut indexed = vec![(index, value)];
+        while parsed.peek().is_some_and(|(next, _, _)| *next == base) {
```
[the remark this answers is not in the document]
peytonr18: This is not true - resolving.

libazureinit-kvp/src/cli.rs
```diff
@@ -425,6 +513,191 @@ fn is_stale<W: Write>(
...
+            if let Some(result) = &event.result {
+                map.insert("result".to_string(), json!(result));
+            }
+            if let Some(duration) = event.duration {
+                map.insert("duration".to_string(), json!(duration));
+            }
+            map.insert("chunks".to_string(), json!(chunks));
+            map.insert("message".to_string(), json!(event.message));
+            serde_json::Value::Object(map)
+        }
```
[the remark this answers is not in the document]
peytonr18: This is not true, as it has been every time this is suggested.

cjp256:
I tried testing on older cloud-init versions using `canonical:0001-com-ubuntu-server-jammy:22_04-lts:22.04.202205280` and got an error:
```
cloud-init-event type=diagnostic name=diagnostic message uuid=b9c3606d-3752-4f26-97ff-26f4f04acda4 ts=2026-08-06T20:20:13.477830Z incarnation=1786047606 chunks=1 message=Dumping last 56746 bytes of cloud-init.log file to KVP starting from index: 24961
malformed key=CLOUD_INIT|1786047606|compressed|cloud-init.log|b7a822ba-4eea-46c0-b559-e84396101132 reason=invalid cloud-init JSON value: EOF while parsing a string at line 1 column 1016 value={"name":"cloud-init.log","type":"compressed","ts":"2026-08-06T20:20:13.479078Z"
```

cjp256:
Is there a way to write diagnostics via CLI?

libazureinit-kvp/src/cli.rs
```diff
@@ -139,6 +139,26 @@ enum Command {
         key: String,
         value: String,
     },
+    /// Emit an azure-init diagnostic event: a structured KVP entry keyed
+    /// `<prefix>|<boot_epoch_time>|<event_level>|<name>|<vm_id>|<event_id>`.
```
cjp256: prefix => agent?

libazureinit-kvp/src/cli.rs
```diff
@@ -691,13 +726,32 @@ fn diagnostics_record_json(record: &DiagnosticRecord) -> serde_json::Value {
 fn diagnostics_event_json(event: &DiagnosticEvent) -> serde_json::Value {
     json!({
         "kind": "event",
-        "level": event.level.to_string(),
+        "boot_epoch_time": event.boot_epoch_time,
+        "event_level": event.event_level.to_string(),
```
cjp256: nit: kind=event and event_ is implied.  if event_ is nicer, I would extend it to name and message maybe?

libazureinit-kvp/src/diagnostics.rs
```diff
@@ -99,30 +97,41 @@ pub const MAX_CHUNK_BYTES: usize = 1022;
 const EVENT_KEY_DELIMITER: char = '|';
 
 /// Format a diagnostic event key as its `|`-delimited on-disk string:
-/// `<prefix>|<vm_id>|<level>|<name>|<event_id>`.
+/// `<prefix>|<boot_epoch_time>|<event_level>|<name>|<vm_id>|<event_id>`.
```
cjp256: probably cleaner to keep the vm_id up front for the new format

libazureinit-kvp/src/diagnostics.rs
```diff
@@ -692,6 +721,93 @@ fn classify_record(key: String, chunk_values: Vec<String>) -> DiagnosticRecord {
...
+        timestamp: meta.ts,
+        result: meta.result,
+        duration: meta.duration,
+        message,
+    })
+}
+
+/// Recover a chunk's raw (still-escaped) `msg` slice — the bytes between
+/// the `"msg":"` marker and the closing `"}` — without unescaping.
+fn escaped_msg_slice(chunk: &str) -> Result<&str, String> {
```
cjp256: nit: maybe leave the cloud_init in name so scope is clear?

libazureinit-kvp/src/cli.rs
peytonr18: This was reworked in the latest commit, let me know your thoughts!

libazureinit-kvp/src/diagnostics.rs
peytonr18: Done!

libazureinit-kvp/src/diagnostics.rs
peytonr18: Addressed in latest commit!

libazureinit-kvp/src/diagnostics.rs
```diff
@@ -5,31 +5,31 @@
...
+//!   every kind.
+//! - **cloud-init keys**
+//!   (`CLOUD_INIT|<incarnation>|<type>|<name>|[<vm_id>|]<uuid>`) store a
+//!   JSON value; the reader pulls `ts`/`result`/`duration`/`msg` from it
+//!   and takes everything else from the key, decoding into the same
+//!   [`DiagnosticEvent`]. This crate only *reads* cloud-init.
+//! - **Chunking**: values longer than [`MAX_CHUNK_BYTES`] are split at
+//!   UTF-8 codepoint boundaries into multiple records written atomically
+//!   under a single lock. Each chunk gets a unique key — the event key
+//!   with a `|<subevent_index>` suffix — so the Hyper-V host, which keeps
```
cjp256: Do non-spanning diagnostics have a subevent index?  It might help to keep it included in all cases so future changes result in easier field interpretation (there's no cost to including it).

libazureinit-kvp/src/diagnostics.rs
```diff
@@ -73,80 +67,124 @@
...
-/// cloud-init's `type`/`name`/`vm_id` layout. [`classify_key`] is the
-/// inverse. For example:
+/// in the same slot as cloud-init's incarnation. `kind` records the
+/// span/event shape. [`classify_key`] is the inverse. For example:
 ///
 /// ```text
-/// azure-init-0.1.0|1785187982|INFO|user:create_user|3f2504e0-...|8f3e9c4a-...
+/// azure-init-0.1.0|1785187982|3f2504e0-...|event|user:create_user|8f3e9c4a-...|2026-07-27T21:33:24.300Z
 /// ```
 fn format_event_key(
```
cjp256, line 150: it might help to clarify this as the _shared_ diagnostic key with the understanding that this is the portion of the key specifically does not have the subevent index as that part is handled by splitting/joining layer.

libazureinit-kvp/src/diagnostics.rs
peytonr18: Right now only multi-record events get the `<subevent_index>` suffix and single records keep the bare key since the `event_id` already makes them unique. 

Adding the index always makes the key structure uniform, which I like, so I will make that change!

This also simplifies the `write_chunked` method, which is a win.

libazureinit-kvp/src/diagnostics.rs
```diff
@@ -136,8 +135,8 @@ fn now_timestamp() -> String {
     Utc::now().format("%Y-%m-%dT%H:%M:%S%.3fZ").to_string()
 }
 
-/// Format an azure-init diagnostic event key as its `|`-delimited on-disk
-/// string:
+/// Format an azure-init diagnostic event's *shared* key as its
+/// `|`-delimited on-disk string:
```
cjp256: not clear what `delimited on-disk` implies as it is also matching the host / KVP channel representation (albeit different encoding?)

doc/kvp.md
```diff
@@ -239,3 +239,50 @@ no truncation.
...
+`Entry::Report`, or `Entry::Raw` in first-seen pool order. Unknown records
+remain raw; recognized but invalid records also carry a `DecodeError`.
+Cloud-init records are supported through a read-only compatibility bridge.
+Invalid physical UTF-8 fails the entire snapshot without partial output.
+Neither reading nor emitting diagnostics clears stale pool data implicitly.
+
+| Command | Output |
+|---------|--------|
+| `libazureinit-kvp dump` | JSON array of physical key/value records in pool order, including duplicates |
+| `libazureinit-kvp dump --parse` | JSON array of diagnostics and reports in timestamp order, then raw entries |
```
cjp256: i think pool ordering should continue to be maintained here.

doc/kvp.md
```diff
@@ -239,3 +239,50 @@ no truncation.
...
+| `libazureinit-kvp dump --text` | Physical records as `KEY=VALUE` lines |
+| `libazureinit-kvp dump --parse --text` | The same timestamp ordering, with binary payloads under `payload_b64` |
```
cjp256: i'm not sure what this means?  binary payloads will have `payload_b64=<base64 content>`?

doc/kvp.md
```diff
@@ -239,3 +239,50 @@ no truncation.
...
+| `libazureinit-kvp dump --parse --name ssh` | Filter diagnostic names by substring; retain reports and raw entries |
```
cjp256, line 267: --kind ?

libazureinit-kvp/diagnostics-proposal.md
```diff
@@ -0,0 +1,512 @@
+# KVP Diagnostics Specification
```
cjp256, line 1: delete before merge?  anything of value here that can be moved out? into /doc or docstrings?

libazureinit-kvp/Cargo.toml
```diff
@@ -9,17 +9,21 @@ license = "MIT"
...
+serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.96"
 tracing = "0.1.40"
-uuid = "1.3"
+uuid = { version = "1.3", features = ["v4"] }
 
 [dev-dependencies]
 rstest = { version = "0.26", default-features = false }
 tempfile = "3"
+uuid = "1.3"
```
cjp256: already a dep
```suggestion
```

libazureinit-kvp/diagnostics-proposal.md
peytonr18: I meant to ask this in our chat -- I could trim this down and put some of the info in kvp.md, but I was also thinking if we choose to ship libazureinit-kvp independently, it might be good to have a /docs section in there as well? 

Open to whatever approach but my initial reaction would be to trim it down, put any kvp specific things into kvp.md, and then create a /docs/diagnostics.md or something similar (and possibly move kvp.md into that folder too)

doc/kvp.md
peytonr18: Yes! Sorry, that may not have been obvious enough with that wording, which I've since adjusted a little. 

That line is intended to clarify how binary payloads are displayed in --text output.

Internally, decoded diagnostics can contain either text (Text) or raw bytes (Bytes). When the payload is raw bytes, we can't safely print it directly, so the CLI base64-encodes it for display and renders it as payload_b64=<base64>. Text payloads continue to be rendered as payload=<text>.

The important detail is that the different key name (payload_b64) signals that the displayed value is a base64 representation of the payload, not the payload itself. This is only a presentation detail for --text output; JSON output preserves the payload structure separately.

doc/kvp.md
peytonr18: Added this in latest commit!

doc/kvp.md
peytonr18: Added this in latest commit, reverted the timeline sorting and fixed all the tests and such accordingly.

doc/kvp.md
```diff
@@ -239,3 +239,55 @@ no truncation.
...
+The writer validates the agent identifier and VM UUID at construction.
+`emit_start` and `emit_finish` share a caller-supplied event UUID;
+`emit_event` generates its own. Durations are `std::time::Duration`
+values stored as integer microseconds. Timestamps are RFC 3339 UTC (`Z`)
+at millisecond precision by default; `DiagnosticWriter::with_timestamp_precision`
+selects second, microsecond, or nanosecond precision, and the reader
+accepts any of those canonical precisions.
+Payloads are plain UTF-8 text or gzip plus base64 (`Encoding::GzB64`).
+
+Each reader call takes one fresh snapshot and returns `Entry::Diagnostic`,
```
cjp256: I get why it is referred to as a snapshot, but it's awkward framing.  I would suggest scrubbing it and simplifying it some.

libazureinit-kvp/src/diagnostics/cloud_init.rs
```diff
@@ -0,0 +1,751 @@
...
+fn decode_compressed(data: &str) -> Result<DiagnosticPayload, DecodeError> {
+    let compact: Vec<_> = data
+        .bytes()
+        .filter(|byte| !byte.is_ascii_whitespace())
+        .collect();
+    let compressed = STANDARD
+        .decode(compact)
+        .map_err(|_| DecodeError::Undecodable)?;
+    let mut bytes = Vec::new();
+    // cloud-init labels zlib streams and line-wrapped base64 as `gz+b64`.
```
cjp256: isn't gz+b64 universal here?  does zlibdecoder path ever get used?

libazureinit-kvp/src/diagnostics/reader.rs
```diff
@@ -0,0 +1,1087 @@
...
+    store: KvpPoolStore,
+}
+
+impl DiagnosticReader {
+    /// Construction performs no I/O and needs no local producer identity.
+    pub fn new(store: KvpPoolStore) -> Self {
+        Self { store }
+    }
+
+    /// Invalid pool UTF-8 fails the snapshot; entries keep first-seen order.
```
cjp256: what could cause invalid pool utf-8?  is this any underlying pool error?

libazureinit-kvp/src/diagnostics/cloud_init.rs
peytonr18: Yes!! The zlibdecoder path does get used -- the reason for this is that the gz+b64 label is shared, but the compressed format isn’t always gzip. Cloud-init’s Azure compressed-event producer calls base64.encodebytes(zlib.compress(event_content)) and labels the result gz+b64. That produces a zlib stream, not a gzip stream, despite the label. Base64 turns the compressed bytes into text suitable for KVP storage, but it doesn’t make those compression formats interchangeable.

So, we have to handle zlib too because we read cloud-init records.  When cloud-init stores a compressed log, we need the correct decompressor to recover its contents because a gzip-only decoder would fail, leaving those records as Raw with an Undecodable error instead of returning the decoded payload.

libazureinit-kvp/src/diagnostics/reader.rs
peytonr18: Invalid UTF-8 is one specific cause of a pool-read failure, not a catch-all for any underlying pool error.
It could happen if another process writes binary data where the pool expects text, truncates a key or value in the middle of a UTF-8 character, or if the pool contents become corrupted. Those are just plausible causes, not issues we've observed in production. Also, we only validate the actual key/value contents up to the terminating NUL; any padding after that is ignored.
 
When this happens, our storage API cannot represent the key or value as a Rust string, so the read fails with KvpError::Io(InvalidData). Other underlying pool errors, such as permission failures or file-read errors, are surfaced separately and are not treated as UTF-8 errors.
 
I mostly introduced that as a safeguard but if you think we don't need to account for any of those scenarios, I can easily remove the invalid poot utf-8 error!

doc/diagnostics.md
```diff
@@ -0,0 +1,273 @@
...
+
+### Reader Output
+
+Each decoded group appears at its first record's position. Failed groups retain
+their physical records and positions. Reads do not modify the pool or sort by
+timestamp. Decoding errors are attached to preserved records:
+
+| Condition | Error token |
+|---|---|
+| Unsupported `DIAG_V*` schema | `unsupported_version` |
```
cjp256: s/DIAG_V/DIAG

libazureinit-kvp/src/diagnostics/diagnostic.rs
```diff
@@ -150,61 +183,87 @@ impl fmt::Display for DecodeError {
...
     pub vm_id: Option<String>,
+    /// Operation or observation name, such as `provision:run` or `dmesg`.
     pub name: String,
-    /// Span endpoints share this ID; standalone events have their own.
+    /// UUID shared by an operation's start and finish; unique for a standalone event.
     pub event_id: String,
+    /// When this diagnostic was emitted, in UTC.
     #[serde(serialize_with = "serialize_timestamp")]
     pub timestamp: DateTime<Utc>,
+    /// Stored payload encoding; `None` means plain text.
```
cjp256: plaintext -> utf-8 encoded plain-text

libazureinit-kvp/src/diagnostics/diagnostic.rs
```diff
@@ -230,21 +291,29 @@ impl Diagnostic {
...
 }
 
+/// A stored record that was not decoded as a diagnostic or report.
 #[derive(Clone, Debug, PartialEq, Eq, Serialize)]
 pub struct RawKeyValue {
+    /// Original key.
     pub key: String,
+    /// Original value, without payload decoding.
     pub value: String,
+    /// Why a recognized record could not be decoded; `None` for unrelated keys.
```
cjp256: s/unrelated/unknown, or non-diagnmostic, key-pairs

libazureinit-kvp/src/diagnostics/diagnostic.rs
```diff
@@ -305,7 +372,7 @@ mod tests {
     use rstest::rstest;
     use serde_json::{json, Value};
 
-    const AGENT: &str = "azure-init-0.1.1";
+    const AGENT: &str = "azure-init/0.1.1";
```
cjp256, line 377: :D

libazureinit-kvp/src/diagnostics/diagnostic.rs
peytonr18: Done in latest!

doc/diagnostics.md
peytonr18: Done in latest!

libazureinit-kvp/src/diagnostics/diagnostic.rs
peytonr18: Done in latest!

doc/diagnostics.md
```diff
@@ -0,0 +1,272 @@
...
+| `timestamp` | Emission time as RFC 3339 UTC with a `Z` suffix |
+| `encoding` | Value representation: `none`, `zlib+b64`, or `gz+b64` |
+| `result` | `success` or `fail`, when applicable |
+| `duration` | Nonnegative elapsed seconds, when measured |
+| `chunk_index` | Decimal chunk index starting at zero, including for a single record |
+
+All fields except `result` and `duration` are required and nonempty. Key fields
+cannot contain `|` or NUL; there is no key-field escaping. Agent strings remain
+opaque to readers, including unversioned names and older naming conventions.
+UUID spellings are preserved rather than rewritten during reading.
```
cjp256:
```suggestion
cannot contain `|` or NUL; there is no key-field escaping.
```

doc/diagnostics.md
```diff
@@ -0,0 +1,272 @@
...
+
+| Kind | Result | Duration | Meaning |
+|---|---|---|---|
+| `start` | Empty | Empty | An operation began |
+| `finish` | Required | Required | An operation ended with a reported outcome and elapsed time |
+| `event` | Optional | Optional | A standalone observation; outcome and timing are independent |
+
+A start and finish share an event ID, agent, VM and operation name. A finish
+carries its own elapsed duration; do not derive it by subtracting wall-clock
+timestamps. Either endpoint remains valid if its counterpart is missing.
```
cjp256:
```suggestion
carries its own elapsed duration to allow caller to accurately measure
the operation of interest without relying on the timestamps of the emitted
diagnostics.
```

doc/diagnostics.md
```diff
@@ -0,0 +1,272 @@
...
+
+### Timestamps and Durations
+
+Timestamps use RFC 3339 UTC `Z` form, with zero, three, six or nine fractional
+digits. The default is milliseconds, for example `2026-08-31T12:34:56.789Z`.
+Numeric offsets and other fractional widths are invalid in native records.
```
cjp256: Do we care so long as it conforms to RFC 3339?  In this case I would say it MUST conform to RFC 3339, this implementation has options for three forms....

Just a thought.  It's fine to keep it constrained.

doc/diagnostics.md
```diff
@@ -0,0 +1,272 @@
...
+
+Durations are decimal **seconds**: `0.312000` is 312 milliseconds. Emission
+defaults to six fractional digits, independently of timestamp precision.
+Producers can select zero, three, six or nine digits; lower digits are discarded.
+
+Accept decimal digits with an optional decimal point and one to nine fractional
+digits. The whole-seconds component is at most `18446744073709551615`. Signs,
```
cjp256: That's a lot of seconds :D 

It might be cleaner to state that it MUST conform to IEEE 754 double and >= 0.

doc/diagnostics.md
```diff
@@ -0,0 +1,272 @@
...
+| `gz+b64` | Standard padded base64 of one gzip member (RFC 1952) | Bytes |
+
+For compressed encodings, decode base64 before decompressing. Native base64 has
+no whitespace. Zlib uses DEFLATE with a 32 KiB window (`wbits=15`) and no preset
+dictionary; gzip uses a basic header without optional fields. The tokens are
+not aliases. Compression does not imply that the decoded content is text.
+
+### Kusto Consumption
+
+Use `zlib+b64` for new compressed telemetry targeting Kusto's
```
cjp256: drop this.

doc/diagnostics.md
```diff
@@ -0,0 +1,272 @@
...
+
+| Item | Limit |
+|---|---|
+| Complete key, including delimiters and index | 254 bytes |
+| Encoded value per chunk | 1,022 bytes |
+| Chunks per payload | 1,023, indexed 0 through 1022 |
+| Agent / name | 32 / 48 bytes |
+| Each UUID | 36 bytes |
+
+These budgets do not impose equivalent read limits on other producers' records.
```
cjp256: There isn't really a full accounting here of potential max widths vs. expected/typical widths and how they fit in the available space.

doc/diagnostics.md
```diff
@@ -0,0 +1,272 @@
...
+Encoded size also does not bound decompressed size.
+
+Invalid metadata, unsupported encodings, duplicate or missing indices, invalid
```
cjp256: I don't know that this paragraph says much.  It's fair to assume that missing/corrupted data will induce errors.  Focus on the ones that are worth calling out like the next paragraph (missing trailing chunks) because it was a design consideration.

cjp256: and maybe split Limits section from Error Handling or something like that.

doc/diagnostics.md
```diff
@@ -0,0 +1,272 @@
...
+
+## Provisioning Reports
+
+The exact key `PROVISIONING_REPORT` identifies a separate, single-record health
+report. Its value is pipe-delimited CSV whose fields contain `key=value`.
+Fields containing pipes, quotes or newlines use CSV double-quote escaping;
+split each decoded field on its first `=`. Field order is not significant for
+reading.
+
+Required fields are `result`, `agent`, `vm_id`, `pps_type` and an RFC 3339
```
cjp256: maybe turn this into a table for easy reading

doc/diagnostics.md
```diff
@@ -0,0 +1,272 @@
...
+as unparsed data. Other source types become events without a normalized outcome
+or duration. Negative or overflowing finish durations are invalid.
+
+Compressed artifacts put a JSON object with `encoding` and `data` inside `msg`;
+extract `data` after reassembly and unescaping.
+Cloud-init's [Azure producer](https://github.com/canonical/cloud-init/blob/main/cloudinit/sources/helpers/azure.py)
+uses zlib compression and line-wrapped base64 while labeling the envelope
+`gz+b64`. Remove ASCII base64 whitespace and accept zlib or gzip under that
+source label, using the actual stream format to select the decompressor.
+This exception does not apply to native `gz+b64` records. Other envelope
```
cjp256: it's implied in here, but maybe call out the reason (cloud-init says gz+b64 but it actually is zlib+b64).

doc/diagnostics.md
```diff
@@ -0,0 +1,272 @@
...
+more records. Each key has exactly eleven pipe-delimited fields:
+
+```text
+DIAG|<agent>|<vm_id>|<kind>|<name>|<event_id>|<timestamp>|<encoding>|<result>|<duration>|<chunk_index>
+```
+
+`DIAG` selects the current format. Do not parse an unsupported diagnostic schema
+using this layout. Agent versions identify the producer, not the schema.
+
+| Field | Meaning |
```
cjp256: Field | Description | Format

doc/kvp.md
```diff
@@ -239,3 +239,34 @@ no truncation.
...
+Appending retains duplicates without a record-count cap. Inserting updates a
+key and collapses its duplicates; inserting new keys and replacing the pool
+enforce a limit of 1,024 distinct keys. That limit is a library policy, not an
+extra field or universal constraint of the KVP format. Map-style reads use the
+last stored value, while physical reads retain every record.
+
+Deletion may swap a record with the file's tail, changing order. Stale-data
+cleanup is explicit and compares modification time with system boot time under
+the write lock. Invalid framing or invalid UTF-8 content fails a read; padding
+after a NUL is ignored rather than interpreted as text.
```
cjp256: This is a lot packed into a few paragraphs.  maybe could use a little sectioning and/or lists.

peytonr18: Agreed, addressed in latest commit!

doc/diagnostics.md
peytonr18: Done!

doc/diagnostics.md
peytonr18: Done! I tried to make that section a little more straightforward too, let me know if you feel like anything is missing.

doc/diagnostics.md
peytonr18: Done in latest!

doc/diagnostics.md
peytonr18: Addressed in latest!

doc/diagnostics.md
peytonr18: Removed in latest!

doc/diagnostics.md
peytonr18: Done in latest!

doc/diagnostics.md
peytonr18: Done in latest!

doc/diagnostics.md
peytonr18: Agreed - we don't care as long as it confroms! The reader now accepts RFC 3339 forms, including offsets and other fractional widths. The writer’s UTC and precision choices are documented separately.

doc/diagnostics.md
peytonr18: I should have addressed this in the latest commit!

doc/diagnostics.md
peytonr18: Addressed in latest!

doc/diagnostics.md
```diff
@@ -0,0 +1,267 @@
...
+
+### Timestamps and Durations
+
+Timestamps must conform to RFC 3339. The writer emits UTC `Z` form with
+second, millisecond (default), microsecond or nanosecond precision.
+
+Durations are finite, nonnegative IEEE 754 double-precision seconds, in
+decimal or exponent notation. Empty means absent; zero is a measured duration.
+
+The writer emits fixed-point durations with 0, 3, 6 (default) or 9 fractional
```
cjp256: this dupes two lines above.

peytonr18: Addressed in latest!

cjp256 (approved):
LGTM, thanks!

